PR #3, opened 2022-06-10
Add govuk frontend

mec:
Get the GOVUK Frontend working nicely with Rails 7.

We'll likely switch to [DfE GOVUK-components](https://github.com/DFE-Digital/govuk-components) (view components) but it has a [bug](https://github.com/DFE-Digital/govuk-components/pull/331) right now, so here we use partials.

The result is:

image


jacksonj04 (approved):
LGTM 🚢

PR #6, opened 2022-06-13
Install dotenv

mec:
Dotenv helps us manage environment variables accross our environmenets.

https://github.com/bkeepers/dotenv


jacksonj04 (approved):
LGTM, my only comment would be it could do with a `.env.example` with a complete list of keys which can be set (also so tooling can use it as a template).

PR #7, opened 2022-06-13
Add Scripts To Rule Them All

jacksonj04:
As per [RFC 023](https://github.com/dxw/tech-team-rfcs/blob/main/rfc-023-use-scripts-to-rule-them-all.md), add the [Scripts To Rule Them All from our Rails template](https://github.com/dxw/rails-template/tree/develop/script).

This simplifies things like onboarding and maintaining dependencies, eg by allowing `script/setup` to install all necessary dependencies and perform tasks to bring a working environment up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,8 +24,11 @@
 # Ignore master key for decrypting credentials and more.
 /config/master.key
 
-# Ignore node_modules
-/node_modules
+# Dependencies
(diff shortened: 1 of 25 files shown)


mec (approved):
Happy to merge this, even without an actual Docker config as we are working towards mirroring in the convenions laid out in the dxw Rails template.

🚢

PR #8, opened 2022-06-13
Use yarn to manage node packages instead of npm

jacksonj04:
This brings us closer to the dxw Rails template, which uses yarn instead of npm.


mec (approved):
🚢

PR #9, opened 2022-06-13
Vendor the applicaiton bundle

mec:
The bundle contains all the gems the applications depends on.

Here we configure bundler to use `vendor/bundle` to store them, this
makes the application more protable and self contained.

Find out more about confgiguring Bundler:

https://bundler.io/v2.3/man/bundle-config.1.html


mec:
As @jacksonj04 points out, Bundler recommend not commiting the `.bundle` directory, so let us not!

PR #11, opened 2022-06-13
Add Docker setup from dxw Rails template

mec:
We want to follow the conventions laid out in the dxw Rails tempalte
[1].

Here we port the Docker setup, modifying it slightly as needed.

[1] https://github.com/dxw/rails-template


jacksonj04 (approved):
There are a few instances of `dfe-accademies` which I think is a typo for `dfe-academies`, but other than that this looks good to me.

mec:
One day I may learn to spell!

PR #25, opened 2022-06-21
(98752) Feature: authenticate users

mec:
We will be using Microsoft to authenticate users as we know all DfE users of our service will have an acoount.

To do so we use the Omniauth gem, a light touch way of intergrating with authentication services. Along with Omniauth, we'll use a Microsoft Graph plugin to orchistrate the authentication.

Initially we will authorise users by storing 'registered' users in our applicaiton and checking the user authenticatin has an email address we know of, this prevents any DfE user access our application.

Once we have an authenticated and authorised user, we store our user id in the secure Rails session cookie, the presence of this id tells us a user is authenticated.

The sign in flow is based around the idea of returning the user to the sign in page whenever something happens:

- successful sign out
- email address not recognised
- authentication failure

And showing a helpful message. The design and content of the messages has been through a brief design review and is our first attempt, we're not looking for perfection there!

Upon a successful authentication and authorisation, users are simply redirected to the existing placeholder home page. Again, we know this is set to change, but we prove the user is signed in by showing their email address.

At this point we don't feel the need to provide a sign out UI as it seems unnecessary as our users would need to sign out of all Microsoft authenticated services in order to do so. We have added a sign out route here and can always add UI should it be seen as helpful.

Most of the work here is done in the final commit, I wanted to break this up, but I am struggling to do so, apologies for the size.

We need to add an ADR for all this, but this PR is large enough so I will follow up with that.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,3 +9,9 @@
 # Reference: https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
 
 DATABASE_URL=postgres://postgres@localhost:5432/database-name
+
+# Azure Active Directory
+#
+# Admin url: https://portal.azure.com.mcas.ms/#view/Microsoft_AAD_RegisteredApps/ApplicationMenuBlade/~/Overview/appId/05b7db93-6384-4b44-9d27-23eb6bd97366/isMSAApp~/false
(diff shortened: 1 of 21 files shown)


jacksonj04 (approved):
LGTM!

PR #28, opened 2022-06-22
Add phase banner and breadcrumbs block to layout

jacksonj04:
Add the phase banner to the top of the site layout, and a block which can be used for breadcrumb navigation.

### Before
localhost_3000_ (1)

### After
localhost_3000_ (2)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -32,6 +32,19 @@
     <%= render partial: "shared/header" %>
 
     <div class="govuk-width-container ">
+
+      <div class="govuk-phase-banner">
+        <p class="govuk-phase-banner__content">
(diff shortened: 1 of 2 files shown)


mec:
Thanks for taking the time to think about these and add them!  I am a little hesitant to add anything to our code base until we know we need to - both these feel like examples of that? 

I am not really sure what phase we are in or if we are even in one yet, so it feels early to add the banner, I would hope we are somewhere in private/public beta territory, but you've gone for pre-alpha - so there is already debate!

Breadcrumbs are often more tricky than they should be (in my experience!), I would question adding anything to the code that implies we have them or users need them until we are confident and can delivery the feature of 'breadcrumbs' we know our users would find helpful.

I know these might feel excesive, but I do feel strongly about keeping the code as lean and simple as we possibly can. Happy to talk about this if helpful.

PR #29, opened 2022-06-22
Add ability to view prototypical task list

jacksonj04:
Add a new button to the home page which takes the user to a view of the current task list as defined in `flows/conversion.yml`. The schema for this workflow definition is in `flows/schemas/task_list.json`.

## Screenshots

localhost_3000_ (3)

localhost_3000_tasks


mec:
I've got some questions around this - do you want to talk sync or async here?

PR #30, opened 2022-06-22
(99040) Add simplecov to monitor test coverage

jacksonj04:
As per [dxw RFC 173](https://github.com/dxw/tech-team-rfcs/blob/main/rfc-173-testing-our-work.md), set up test coverage monitoring so that anything less than 100% coverage counts as a test failure.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,6 +44,7 @@ end
 group :test do
   gem "capybara"
   gem "climate_control"
+  gem "simplecov", "~> 0.21.2"
 end
 
 group :development do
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -56,3 +56,5 @@ group :development do
   # Speed up commands on slow machines / big apps [https://github.com/rails/spring]
   # gem "spring"
 end
+
+gem "simplecov", "~> 0.21.2"
```
mec, line 60: I think this should go into the `test` group as it is only needed there?

mec (changes requested):

mec (approved):
🚢

PR #33, opened 2022-06-24
[87] Add Projects data model and views

jacksonj04:
Add the underlying data model for a `Project`, as well as the ability to list existing projects and create new ones.

### List all projects
localhost_3000_ (1)

### New project
localhost_3000_project_new

### Project
localhost_3000_project_e93e768a-67a0-4264-8230-e4dec8ed0dc7


db/migrate/20220623165020_create_projects.rb
```diff
@@ -0,0 +1,10 @@
+class CreateProjects < ActiveRecord::Migration[7.0]
+  def change
+    enable_extension "pgcrypto" unless extension_enabled?("pgcrypto")
```
mec, line 3: This is in the schema now, so I don't think we need it https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/58e5f827bfe08570ca8312d32f93521164eb0ba2/db/schema.rb#L15

db/migrate/20220623165020_create_projects.rb
```diff
@@ -0,0 +1,10 @@
...
+    
+    create_table :projects, id: :uuid do |t|
+      t.integer :urn # Every URN really is an integer, not a string which only contains numerical characters.
```
mec, line 6: As `urn` is so important to the model I wonder if we should:
- require it in the db and the applicaiton (validate the model)
- put an index on it

spec/rails_helper.rb
```diff
@@ -62,6 +62,9 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 
+  # Include helpers for tests
+  config.include SignInHelpers
```
mec, line 66: ❤️ this, not seen it done this way! 👍

spec/support/feature_helpers.rb
```diff
@@ -0,0 +1,7 @@
+module FeatureHelpers
+  def sign_in_with_user(user_email_address)
+    mock_auth_with_user(user_email_address)
+    visit sign_in_path
+    click_button("Sign in with your DfE Microsoft account")
```
mec, line 5: It would be worth using the locale here, incase we change it, one less thing to do.

`click_button(I18n.t("sign_in.button"))`

spec/rails_helper.rb
```diff
@@ -64,6 +64,7 @@
 
   # Include helpers for tests
   config.include SignInHelpers
+  config.include FeatureHelpers, type: :feature
```
mec, line 67: Very cool!

app/controllers/project_controller.rb
```diff
@@ -0,0 +1,23 @@
+class ProjectController < ApplicationController
+  default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
```
mec, line 2: It might be worth popping this into `ApplicationController` so it is the default everywhere.

spec/features/users_can_view_projects_spec.rb
```diff
@@ -10,3 +10,16 @@
       expect(page).to have_content("19283746")
     end
 end
+
+RSpec.feature "Users can view a single project" do
+
+  scenario "by following a link from the home page" do
+    sign_in_with_user("user@education.gov.uk")
+    
+    Project.create!(urn: 19283746)
```
mec, line 19: It might be nice to assign the Project and then refer to the urn via that, for me it makes things that tiny bit more legible:
```ruby
single_project = Project.create!(urn: 19283746)
click_on single_project.urn
expect(page).to have_content(single_project.urn)
```

spec/features/users_can_create_projects_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+RSpec.feature "Users can create a new project" do
+
```
mec, line 4: I think this space will cause a linter fail - is the linter part of your standard testing?

app/views/layouts/application.html.erb
```diff
@@ -32,6 +32,7 @@
     <%= render partial: "shared/header" %>
 
     <div class="govuk-width-container ">
+      <%= yield :breadcrumbs %>
```
mec, line 35: Is this a breadcrumb? https://design-system.service.gov.uk/components/back-link/

mec (changes requested):
I have left a few comments.

I wonder if we drop the `home` view and just use `Project#index` as the home view now, we can aliase `root_path` so it shouldn't be a big change - what do you think?

app/models/project.rb
```diff
@@ -0,0 +1,3 @@
+class Project < ApplicationRecord
+  validates :urn, presence: true
```
mec, line 2: Given our level of confidence about this being an integer, what do you think about validating that as well?

app/controllers/project_controller.rb
```diff
@@ -4,4 +4,20 @@ class ProjectController < ApplicationController
   def index
     @projects = Project.all
   end
+
+  def new
+    @project = Project.new
+  end
+
+  def create
+    Project.create(project_params)
```
mec, line 17: As we are at least validating the prescence, it would be helpful for users to see an error when the validation fails. The GOV.UK Formbuilder has really good support for it: 

https://govuk-form-builder.netlify.app/introduction/error-handling/#input-erb-a-form-with-multiple-errors

Have a look at the RODA controllers to get an idea of the approach: 

https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/0b75cab3aaa73db9e0c600413eac548279b201b5/app/controllers/staff/organisations_controller.rb#L30

app/views/project/new.html.erb
```diff
@@ -0,0 +1,10 @@
+<% content_for :pre_content_nav do %>
+<%= link_to "Back", root_path, class: "govuk-back-link" %>
+<% end %>
+
+<h1 class="govuk-heading-l">Add a new project</h1>
+
+<%= form_with model: @project, url: project_index_path do |form| %>
+  <%= form.govuk_text_field :urn, placeholder: "134679", label: { text: "What is the existing school's URN?", tag: 'h3', size: 'm' } %>
```
mec, line 8: It might be a me thing, but I am not convinced that the placeholder is valuable here? I would much rather see a hint:

https://design-system.service.gov.uk/components/text-input/

mec (changes requested):
There a couple of things I think we should do to really wrap this up, looking really nice though! 👍

app/controllers/project_controller.rb
```diff
@@ -0,0 +1,7 @@
+class ProjectController < ApplicationController
```
mec, line 1: This is absolutely my bad as I should have spotted it earlier! By convention, Rails controllers are pluralised: 

https://guides.rubyonrails.org/action_controller_overview.html#controller-naming-convention

This singular project may bite us later so I think we should change it to`ProjectsController` the routes will need updating as well.

Apologies!

app/views/project/new.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :pre_content_nav do %>
+<%= link_to "Back", root_path, class: "govuk-back-link" %>
+<% end %>
+
+<%= form_with model: @project, url: project_index_path do |form| %>
```
mec, line 5: Once we hare conventional controller naming, this can become:

```ruby
form_for @project do |form|
```

app/views/project/new.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+  <%= form.govuk_error_summary %>
+
+  <h1 class="govuk-heading-l"><%= t("project.new.title") %></h1>
+
+  <%= form.govuk_number_field :urn,
```
mec, line 11: This is super interesting! I booted up the app and could never see the 'not a number' error!? After much head scratching I tracked it down to this.

Because we are using `govuk_number_field` modern browsers do not submit anything else, so providing 'three' actually submits '' (empty string).

The feature spec you have passes becuase the headless browser Capybara uses is not modern enough to handle this nuance so it submits 'three'.

Because we want to help users I think it is better to change this to `govuk_text_field` so we can display helpful errors, I also think the behaviour is more consistent - what do you think?

config/locales/en.yml
```diff
@@ -45,6 +45,14 @@ en:
     users:
       create:
         error: You must provide a users email address e.g. bin/rails users:create EMAIL_ADDRESS=name@example.com
-  project_list:
-    title:
-      all: All projects
+  project:
```
mec, line 48: We can leverage the GOV.UK Form Builder locales here by following convention here:

https://govuk-form-builder.netlify.app/building-blocks/localisation/#populating-label-caption-and-hint-text-from-the-localisation-data

mec (changes requested):
Last time I promise!

mec (approved):
🚢

PR #37, opened 2022-06-28
(106) Assign a delivery officer to a project

jacksonj04:
- [x] A project has a concept of `delivery_officer`, which maps to a `User`
- [x] The delivery officer for a project can be set or unset

## Screenshots

localhost_3000_projects_ec97479b-300c-4dd7-8bed-d4432ffa56be

localhost_3000_projects_ec97479b-300c-4dd7-8bed-d4432ffa56be_edit

localhost_3000_projects_ec97479b-300c-4dd7-8bed-d4432ffa56be (1)

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -25,9 +25,23 @@ class ProjectsController < ApplicationController
     end
   end
 
+  def edit
+    @project = Project.find(params[:id])
+    @users = User.all
+  end
+
+  def update
+    @project = Project.find(params[:id])
+    @project.assign_attributes(project_params)
+
+    @project.save
+    flash[:notice] = I18n.t("project.update.success")
+    redirect_to project_path(@project)
+  end
+
   private
 
   def project_params
-    params.require(:project).permit(:urn)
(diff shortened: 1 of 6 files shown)


mec (commented):
What is the thinking behind not setting the delivery officer on Project creation?

Right now, I see creating a poject as a manual process a 'team leader' will perform and it feels right that is one step? Later I can imagine automation that creates the project in our service and at that point the team leader goes in and assigns delivery officers. but I don't see that happening anytime soon?

db/migrate/20220628153154_add_team_leader_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddTeamLeaderToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :team_leader, :boolean, default: false
```
mec, line 3: Not sure we needed this yet, but it's early days and I know you are working on the authorisation, so happy it is here. 👍

mec (approved):
I have one non-blocking question about the user journey, but otherwise this is good.

PR #41, opened 2022-07-04
(142) Create sections and tasks from workflow

jkempster34:
https://trello.com/c/4U1Nvl1E

### Add models for section and task
Add models to represent the sections and tasks associated with the task list of a project.

Add `shoulda-matchers` Gem to use one-liner tests.

### Add YAML conversion workflow
We want to generate task list sections and tasks based on a YAML definition. This will mean that we can easily change and update the task list, as well as introduce new definitions for other workflows.

This also adds the `ajv` package so that we can validate that our workflows fit the schema defined in `./workflows/schemas/task-list.json`

### Add TaskListCreator service
Add a service that creates task list sections and tasks when passed a project. This will be used at on project creation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -85,6 +85,7 @@ COPY script ${APP_HOME}/script
 COPY public ${APP_HOME}/public
 COPY vendor ${APP_HOME}/vendor
 COPY bin ${APP_HOME}/bin
+COPY workflows ${APP_HOME}/workflows
 COPY config ${APP_HOME}/config
 COPY lib ${APP_HOME}/lib
 COPY db ${APP_HOME}/db
--- a/app/services/task_list_creator.rb
+++ b/app/services/task_list_creator.rb
@@ -1,0 +1,22 @@
+class TaskListCreator
+  WORKFLOW_NAME = "conversion".freeze
+
+  def call(project)
+    workflow = load_workflow
+    create_task_list_from_workflow(project, workflow)
+  end
+
+  private def load_workflow
+    YAML.load_file("workflows/#{WORKFLOW_NAME}.yml")
+  end
+
+  private def create_task_list_from_workflow(project, workflow)
+    workflow["sections"].each_with_index do |workflow_section, index|
+      section = Section.create(title: workflow_section["title"], order: index, project: project)
+
+      workflow_section["tasks"].each_with_index do |workflow_task, index|
+        Task.create(title: workflow_task["title"], order: index, section: section)
+      end
+    end
+  end
+end
--- a/script/all/test
+++ b/script/all/test
@@ -42,7 +42,7 @@ else
   fi
 
   echo "==> Validating workflows..."
-  yarn run ajv -s workflows/schemas/task-list.json -d workflows/conversion.yml --spec=draft2020
+  yarn run ajv -s workflows/schemas/task-list.json -d workflows/*.yml --spec=draft2020
 
   echo "==> Running the tests..."
   bundle exec rspec
--- a/script/all/test
+++ b/script/all/test
@@ -41,6 +41,9 @@ else
     yarn run lint:js
   fi
 
+  echo "==> Validating workflows..."
+  yarn run ajv -s workflows/schemas/task-list.json -d workflows/conversion.yml --spec=draft2020
+
   echo "==> Running the tests..."
   bundle exec rspec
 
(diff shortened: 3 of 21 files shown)


script/all/test
```diff
@@ -41,6 +41,9 @@ else
     yarn run lint:js
   fi
 
+  echo "==> Validating workflows..."
+  yarn run ajv -s workflows/schemas/task-list.json -d workflows/conversion.yml --spec=draft2020
```
jacksonj04, line 45: It might be worth documenting the existence of this in the commit message.

spec/services/task_list_creator_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "rails_helper"
+
+RSpec.describe TaskListCreator, type: :model do
+  let(:task_list_creator) { TaskListCreator.new }
+
+  let(:mock_workflow) { file_fixture("workflows/conversion.yml") }
+
+  before do
+    allow(YAML).to receive(:load_file).with("workflows/conversion.yml").and_return(
```
jacksonj04, line 9: You can use `. and_call_original` here instead of writing your own `and_return` to make the mock behave as the original would.

Dockerfile
```diff
@@ -85,6 +85,7 @@ COPY script ${APP_HOME}/script
 COPY public ${APP_HOME}/public
 COPY vendor ${APP_HOME}/vendor
 COPY bin ${APP_HOME}/bin
+COPY workflows ${APP_HOME}/workflows
```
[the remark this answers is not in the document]
jkempster34: 🙌

spec/services/task_list_creator_spec.rb
jkempster34: I don't think this will work. I want to load the mock workflow defined in `./spec/fixtures/files/workflows/conversions.yml`. If I do `and_call_original`,  it will load the proper workflow (the on in `./workflows`)

mec (changes requested):
Nice bit of work to kick things off!

This is my first 'proper' review for you so I hope you'll forgive some nitpicking style things that I would like to raise! Always happy to talk about anything that comes up in a review if preferable.

Could we keep commit messages line length to around 80, generally I would like us to follow this advice: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html it's not a deal breaker, but I will raise it here this once and I am a big believer in a good commit history!

Nothing to do with you but I feel like the ruby linter is not running in CI..?

I am not convinced we need all the yaml schema checks outside of the applicaiton, I would hope our tests would break elsewhere if the workflow file was incorrect in some way?  For me the `conversion.yml` is part of the our application and not user generated so we should treat it as such. Happy to discuss though and shoot for concensus.

app/services/task_list_creator.rb
```diff
@@ -0,0 +1,24 @@
+class TaskListCreator
+  WORKFLOW_NAME = "conversion".freeze
+
+  def call(project)
+    workflow = load_workflow
+    create_task_list_from_workflow(project, workflow)
+  end
+
+private
```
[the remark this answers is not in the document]
jkempster34: Nice, I actually prefer that!

jacksonj04:
@mec I'd be in favour of keeping the yaml validation because although we can write tests which will fail if the yaml isn't valid (ie we try parse it, we can't, fall over) if we want to check to see if it's _correct_ (does everything have required fields, are those fields the right type etc) then we have to write tests which basically replicate a schema's constraints anyway.

mec:
I think that is the point I am trying to make @jacksonj04 the yaml is part of our applicaiton so if reading the file results in 0 tasks or with sections missing then our application is broken, but we should have those tests inside the system, i.e. the test suite.

I would make the same case for the correctness, tests that check that loading the file results in the correct behaviour is exactly what I would want to see in our test suite and it worries me that here we are bringing a lot of tooling to do just that, but outside the system.

I disagree with the point about end to end test as well, I would expect to see tests that gives us confidence that the file has been loaded successfully and correctly.

I wonder if this is caused by a difference in who we see is adding these yaml files? To be clear, I can only see world in which we (developers) add them to the system, they could just as easily be a `ConversionProjectBlueprint` object in my view, I just think the yaml is a nicer way to view and update them with the expected churn on the values.

As ever, happy to talk this over here or elsewhere. As you and @jkempster34  have worked on this, I will accepted the consensus, if you both feel this is valuable.

jacksonj04 (approved):
LGTM! 🚢

mec (approved):
🚢 👍

jkempster34:
[Create task list view](https://trello.com/c/YpH2TlYQ/173-create-task-list-view)

PR #45, opened 2022-07-07
(173) Task list view

jkempster34:
https://trello.com/c/YpH2TlYQ

### Add task list view
- Add a task list the the project `show` page. The GOV.UK task list component is currently in a "being worked on" state, meaning that we have to pull in the styling from the prototype kit for now.

- Generate a task list on project creation, via the TaskListCreator.

- Add the `FactoryBot` Gem, along with minimal factories for project, section, and task, so that we can easily create models in various states for easier testing.

image

### DIFF
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -1,5 +1,5 @@
 <% content_for :pre_content_nav do %>
-<%= link_to "Back", root_path, class: "govuk-back-link" %>
+  <%= link_to "Back", root_path, class: "govuk-back-link" %>
 <% end %>
 
 <div class="govuk-grid-row">
@@ -15,3 +15,7 @@
     </p>
   </div>
 </div>
+
+<h1 class="govuk-heading-l"><%= t('project.show.title') %></h1>
+
+<%= render "shared/task_list" %>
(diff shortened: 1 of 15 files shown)


app/views/projects/show.html.erb
```diff
@@ -15,3 +15,30 @@
     </p>
   </div>
 </div>
+
+<h1 class="govuk-heading-l"><%= t('project.show.title') %></h1>
+
+<h2 class="govuk-heading-s govuk-!-margin-bottom-2">Application incomplete</h2>
```
mec, line 21: I would loose these two lines - I know they are on the exmple of a task list, but not sure they add value for our use case.

app/views/projects/show.html.erb
```diff
@@ -15,3 +15,30 @@
...
+<p class="govuk-body govuk-!-margin-bottom-7">You have completed 3 of 8 sections.</p>
+
+<ol class="app-task-list">
```
mec, line 24: I would probably have broken this out into partials, but it's not a blocker.

mec (approved):
I left two tiny, non-blocking comments. 🚢

jkempster34:
[Create task list view](https://trello.com/c/YpH2TlYQ/173-create-task-list-view)

PR #46, opened 2022-07-11
(181) Do not flag linters to auto fix in CI

mec:
We kept seeing PRs with all tests passing but we knew the PR contained
linting errors.

It turns out our scripts test for the presence of the
AUTOMATICALLY_FIX_LINTING variable, not it's value, you can set it to
true or false and see the same behaviour.

As we do not want to run in fix mode, we remove the variable from the ci
docker compose file and so, switch the statement in `script/all/test`.


jacksonj04 (approved):
Good ol' "false" being truthy!

PR #50, opened 2022-07-12
(163) Chore: update task list worflow yaml with latest data

mec:
Base on #49 do not merge.


app/workflows/conversion.yml
```diff
@@ -1,85 +1,203 @@
...
+              Check that KIM is up to date with information from the application
+              and the handover proforma
+          - title:
+              Do additional research about the school / project (if required)
+          - title:
+              Identify issues or questions to discuss with the regional delivery
+              officer in handover meeting
+          - title: Have handover meeting with the regional delivery officer
+
+      - title: Local authority questionaire
```
jacksonj04, line 25: I think this should be "questionnaire" throughout

jacksonj04 (approved):
One typo I spotted, but otherwise good to go with the caveat that although I can say this workflow is _valid_ I can't say it's _correct_.

mec:
@jacksonj04 All I can say is it was a lot of _copy_ and _pasting_! 😄

PR #56, opened 2022-07-15
ADR: Use the Academies Database

mec:


doc/decisions/0007-use-the-academies-db-as-the-datastore.md
```diff
@@ -0,0 +1,47 @@
...
+will use the Academies database as their back end in line with the convention
+followed by other RDD applications.
+
+## Consequences
+
+### Benefits
+
+- No need for an IT Health Check that will cost money and cause a delay
+- Follows the convention used elsewhere in RDD, making support easier
+- Keeps all RDD data in one place, making integrations and reporting easier
```
jacksonj04, line 42: Architecturally this makes me twitchy because integrations/reporting which rely on talking directly to our database are Bad Juju, so not convinced this is actually a benefit, but...

jacksonj04 (approved):

doc/decisions/0007-use-the-academies-db-as-the-datastore.md
mec: Yeah, I am not wholly sold on this generally, feels like it was a peace offering to stop all the noise. I am going to assume from the approval you will let it go.

PR #58, opened 2022-07-15
Validate establishment exists on Project creation

jkempster34:
We fetch establishment information from the API by URN. We want to enforce that the establishment exists at the point of Project creation.

image

I've tried to reduce the mocking by having a default establishment on the project factory `establishment { build(:academies_api_establishment) }`, but it didn't help much. I've also used `any_args` a couple of times where we are testing the other validations. You can't control the order of validation, and mocking a specific URN doesn't make sense in those tests (we could also potentially remove the feature specs as the model specs cover the validations anyway)

### DIFF
--- a/spec/features/users_can_create_projects_spec.rb
+++ b/spec/features/users_can_create_projects_spec.rb
@@ -62,6 +62,8 @@ RSpec.feature "Team leaders can create a new project" do
   end
 
   context "the URN is empty" do
+    before { mock_successful_api_establishment_response(urn: any_args) }
+
     scenario "the user is shown an error message" do
       click_button("Continue")
       expect(page).to have_content("can't be blank")
@@ -69,6 +71,8 @@ RSpec.feature "Team leaders can create a new project" do
   end
 
   context "the URN is invalid" do
+    before { mock_successful_api_establishment_response(urn: any_args) }
+
     scenario "the user is shown an error message" do
       fill_in "project-urn-field", with: "three"
       click_button("Continue")
--- a/spec/helpers/task_list_helper_spec.rb
+++ b/spec/helpers/task_list_helper_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 
 RSpec.describe TaskListHelper, type: :helper do
   describe "#task_status_id" do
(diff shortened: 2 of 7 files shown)


spec/features/users_can_create_projects_spec.rb
```diff
@@ -62,13 +62,17 @@
   end
 
   context "the URN is empty" do
+    before { mock_successful_api_establishment_response(urn: any_args) }
```
mec, line 65: For me, these feature specs are too much, I would take my  confidence from the model validations specs, we are even checking the error messages, which is great - I like to rely on Rails doing the right thing and showing those messages and not test that with features.

I don't think we need to take this out in this work - but maybe we look to refactor our feature specs soon to set a precedence we want to continue.

spec/models/project_spec.rb
```diff
@@ -6,9 +6,27 @@
...
+        end
+
+        before do
+          allow_any_instance_of(AcademiesApi::Client).to \
+            receive(:get_establishment) { establishment_result }
+        end
+
+        it "is invalid" do
+          expect(subject).to_not be_valid
+          subject.errors[:urn].include?("No establishment exists with that URN. Enter a valid URN.")
```
mec, line 27: Checking the error, nice! 👍

spec/helpers/task_list_helper_spec.rb
```diff
@@ -2,6 +2,8 @@
 
 RSpec.describe TaskListHelper, type: :helper do
   describe "#task_status_id" do
+    before { mock_successful_api_establishment_response(urn: 12345) }
```
mec, line 5: The scope of this is a little wide, I think you felt like you shouldn't need this mocking and I don't think you do.

This helper simply takes a Task and renders a string from the title, to exercise that behaviour all we need is:

```ruby
      task = Task.new(title: "Clear land questionnaire")
      expect(helper.task_status_id(task)).to eq "clear-land-questionnaire-status"
```
```

mec (changes requested):
I think you can tighten up the helper spec and we could loose the features at a later date, but otherwise this is super nice! 👍

PR #61, opened 2022-07-19
Validate a single matching conversion project exists on project creation

jacksonj04:
When creating a project, catch any exceptions around either missing or multiple conversion projects from the API and present an in-form error message to the user.

Simplifies places in the feature specs where the `mock_successful_api_establishment_response` helper is called with the new `mock_successful_api_responses`, which mocks both establishment and conversion project calls in one go.

## Screenshots

localhost_3000_projects

### DIFF
--- a/spec/features/users_can_view_projects_spec.rb
+++ b/spec/features/users_can_view_projects_spec.rb
@@ -6,9 +6,9 @@ RSpec.feature "Users can view a list of projects" do
   let(:user_2) { create(:user, email: "user2@education.gov.uk") }
 
   before do
-    mock_successful_api_establishment_response(urn: 1001)
-    mock_successful_api_establishment_response(urn: 1002)
-    mock_successful_api_establishment_response(urn: 1003)
+    mock_successful_api_responses(urn: 1001)
+    mock_successful_api_responses(urn: 1002)
+    mock_successful_api_responses(urn: 1003)
     @unassigned_project = Project.create!(urn: 1001)
     @user1_project = Project.create!(urn: 1002, delivery_officer: user_1)
     @user2_project = Project.create!(urn: 1003, delivery_officer: user_2)
@@ -47,7 +47,10 @@ RSpec.feature "Users can view a single project" do
   let(:urn) { 19283746 }
   let(:establishment) { build(:academies_api_establishment) }
 
-  before { mock_successful_api_establishment_response(urn: urn, establishment: establishment) }
+  before do
+    mock_successful_api_establishment_response(urn: urn, establishment: establishment)
+    mock_successful_api_responses(urn: urn)
+  end
 
   scenario "by following a link from the home page" do
     sign_in_with_user(create(:user, :team_leader))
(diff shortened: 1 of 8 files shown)


spec/features/users_can_view_projects_spec.rb
```diff
@@ -47,7 +47,10 @@
   let(:urn) { 19283746 }
   let(:establishment) { build(:academies_api_establishment) }
 
-  before { mock_successful_api_establishment_response(urn: urn, establishment: establishment) }
+  before {
```
jkempster34, line 50: I think now this is multi-line, it's more conventional to update it to a `do...end`. Seems like a contentious issue though https://stackoverflow.com/questions/5587264/do-end-vs-curly-braces-for-blocks-in-ruby

spec/models/project_spec.rb
```diff
@@ -18,13 +18,48 @@
...
         before do
+          mock_successful_api_conversion_project_response(urn: 12345)
           allow_any_instance_of(AcademiesApi::Client).to \
             receive(:get_establishment) { establishment_result }
         end
 
         it "is invalid" do
           expect(subject).to_not be_valid
-          subject.errors[:urn].include?("No establishment exists with that URN. Enter a valid URN.")
+          expect(subject.errors[:urn]).to include(I18n.t("activerecord.errors.models.project.no_establishment_found"))
```
jkempster34, line 28: Thanks for catching this 🙏

spec/models/project_spec.rb
```diff
@@ -18,13 +18,48 @@
...
+            receive(:get_conversion_project) { conversion_project_result }
+        end
+
+        it "is invalid" do
+          expect(subject).to_not be_valid
+          expect(subject.errors[:urn]).to include(I18n.t("activerecord.errors.models.project.no_conversion_project_found"))
+        end
+      end
+
+      context "when no conversion project with that URN exists in the API" do
```
jkempster34, line 49: Copy paste error here. Should be "when the API returns multiple conversion projects" (or similar)

spec/models/project_spec.rb
```diff
@@ -41,7 +76,10 @@
     subject { described_class.new(urn: urn) }
 
     context "when the API returns a successful response" do
-      before { mock_successful_api_establishment_response(urn: urn, establishment:) }
+      before {
```
jkempster34, line 79: Another potential `do...end`

spec/support/academies_api_helpers.rb
```diff
@@ -1,4 +1,9 @@
 module AcademiesApiHelpers
+  def mock_successful_api_responses(urn:)
```
jkempster34, line 2: 🙌

jkempster34 (changes requested):
Awesome 👍 Found a couple of small things:
- Spotted a copy paste error in one of the specs. 
- Also left something about multi-line blocks, but I don't feel strongly about it 🤷 

Otherwise 🚢

jkempster34 (approved):
💯

PR #62, opened 2022-07-20
(194) View and update actions

jkempster34:
https://trello.com/c/CsNfWkNp

### Show a checklist of Actions on the individual Task page
Initially, we want to show our YAML Actions as a completable checklist on each individual Task page.

### DRY up relationships in existing factories
We can simply use the name of the related factory when declaring relationships in FactoryBot.

### Add update action to Task controller
Users need to be able to complete Actions, and for this to be persisted.

### DIFF
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -10,7 +10,7 @@ class TasksController < ApplicationController
     @task.actions.update_all(completed: false)
     @task.actions.where(title: titles).update_all(completed: true)
 
-    redirect_to(project_path(params[:project_id]))
+    redirect_to(project_path(@task.project))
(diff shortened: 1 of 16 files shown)


jkempster34:
[View and update a Tasks Actions](https://trello.com/c/CsNfWkNp/194-view-and-update-a-tasks-actions)

mec:
I started to review this and realised that @jacksonj04 is already doing one, so I will stop now.

jacksonj04 (changes requested):
A couple of little tweaks, along with the stuff we chatted about in Slack.

I also think we could do with adding a 'back' link to the task page.

jacksonj04 (approved):

PR #63, opened 2022-07-21
(227) Use DfE's GOV.UK Components library

jacksonj04:
Installs the DfE GOV.UK Components library and swaps out parts of our templates for the corresponding components (where they exist).

The only visual change is that the application now has the service name in the header, otherwise all HTML is identical.


jkempster34 (approved):
💯 Looks great

PR #67, opened 2022-07-22
(256) Tags in the task list reflect the actual state of a task

jacksonj04:
The completion status of each item in the task list should reflect reality instead of all being marked as "Completed".

### Before
localhost_3000_projects_f7e0efd0-dad6-47a3-a9da-e0e8e318db3e (1)

### After
localhost_3000_projects_f7e0efd0-dad6-47a3-a9da-e0e8e318db3e

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -6,7 +6,21 @@ class Task < ApplicationRecord
 
   delegate :project, to: :section
 
+  def actions_count
+    @actions_count ||= actions.all.count
+  end
+
+  def completed_actions_count
+    @completed_actions_count ||= actions.where(completed: true).count
+  end
+
   def status
-    :unknown
+    if completed_actions_count == 0
+      :not_started
+    elsif completed_actions_count == actions_count
+      :completed
+    elsif completed_actions_count < actions_count
+      :in_progress
+    end
   end
 end
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -5,4 +5,8 @@ class Task < ApplicationRecord
   default_scope { order(order: "asc") }
 
   delegate :project, to: :section
+
+  def status
+    :unknown
+  end
 end
--- a/spec/helpers/task_list_helper_spec.rb
+++ b/spec/helpers/task_list_helper_spec.rb
@@ -12,7 +12,13 @@ RSpec.describe TaskListHelper, type: :helper do
   describe "#task_status_tag" do
     let(:task) { Task.new(title: "Clear land questionnaire") }
 
-    it "returns a tag representing the task status" do
+    it "returns a tag representing the task status when the status is known" do
+      allow(task).to receive(:status).and_return(:completed)
+      expect(helper.task_status_tag(task)).to eq '<strong class="govuk-tag govuk-tag--turquoise app-task-list__tag" id="clear-land-questionnaire-status">Complete</strong>'
+    end
+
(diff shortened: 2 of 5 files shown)


app/models/task.rb
```diff
@@ -5,4 +5,22 @@ class Task < ApplicationRecord
   default_scope { order(order: "asc") }
 
   delegate :project, to: :section
+
+  def actions_count
+    actions_count ||= actions.all.count
```
jkempster34: I don't think this memoization will work without assigning to an instance variable (I haven't ever seen it without)

```
def actions_count
    @actions_count ||= actions.all.count
end
```

app/models/task.rb
```diff
@@ -5,4 +5,22 @@ class Task < ApplicationRecord
...
+  end
+
+  def completed_actions_count
```
jkempster34, line 13: Same here

spec/helpers/task_list_helper_spec.rb
```diff
@@ -8,4 +8,13 @@
...
     end
   end
+
+  describe "#task_status_tag" do
+    let(:task) { Task.new(title: "Clear land questionnaire") }
+
+    it "returns a tag representing the task status" do
+      allow(task).to receive(:status).and_return(:completed)
+      expect(helper.task_status_tag(task)).to eq '<strong class=\"govuk-tag govuk-tag--turquoise app-task-list__tag\" id=\"clear-land-questionnaire-status\">Complete</strong>'
+    end
```
jkempster34, line 18: Is it worth adding a test for the unknown state?

spec/models/task_spec.rb
```diff
@@ -24,4 +24,67 @@
       end
     end
   end
+
+  describe "Methods" do
```
jkempster34, line 28: We didn't group the methods into a describe block in project_spec.rb. Might be worth doing a small commit to update those for consistency.

spec/models/task_spec.rb
```diff
@@ -24,4 +24,67 @@
...
+      let!(:task) { create(:task) }
+
+      before { FactoryBot.create_list(:action, 3, task:) }
+
+      it "returns the number of actions associated with the task" do
+        expect(task.actions_count).to eq 3
+      end
+    end
+
+    describe "completed_actions_count" do
```
jkempster34, line 41: These can be `#completed_action_count` to match the Ruby style of documentation

https://www.betterspecs.org/#describe

spec/models/task_spec.rb
```diff
@@ -24,4 +24,67 @@
...
+        FactoryBot.create_list(:action, 1, task:)
+        FactoryBot.create_list(:action, 2, task:, completed: true)
+      end
+
+      it "returns the number of completed actions associated with the task" do
+        expect(task.completed_actions_count).to eq 2
+      end
+    end
+
+    describe "status" do
```
jkempster34, line 54: Same here

jkempster34 (changes requested):
Looks good 👍 Left some small comments. Looks similar to how it was done in Schools Experience (though we had over 20 different tags)
https://github.com/DFE-Digital/schools-experience/blob/master/app/helpers/schools/placement_dates_helper.rb
https://github.com/DFE-Digital/schools-experience/blob/f0222fe258262aacff2e1b0c974b6eead2b4c11a/app/presenters/candidates/placement_presenter.rb

jkempster34 (approved):

PR #68, opened 2022-07-22
(225) Chore: local environment uses mssql server

mec:
This is a first attempt to switch our development environment from Postgres to Microsoft SQL Server. The reasons for doing so are in this [ADR](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/develop/doc/decisions/0007-use-the-academies-db-as-the-datastore.md).

There is a lot here, take your time and please ask anything, but I'll be the frist to say this is not something I am super experienced with!

As SQL Server cannot be run natively on Mac OS we have to use Docker, this renders the whole Docker/No Docker thing even more confusing, but we are essentially still running Ruby and Rails natively. I haven't updated the Docker parts yet as:

- no one on the team uses them
- we have no production environment using them

I've have tried to update the  no-docker scripts, but I don't use them (for reasons) so I cannot test them, I'd appreciate feedback on that part of the work!

The important part is probably the decision to let our local development environments differ from production slightly in order to make the day-to-day usage simpler - details in the ADR, but in summary: using a SQL Server schema locally means  day-to-day developer task become incredibly complex and the trade off just doesn't seem worth it.

I don't think we need perfection at this point as I expect we'll be back here soon with Devops help and can do more then - the main goal is to make the switch before work starts on our deployments.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -77,6 +77,6 @@ Rails.application.configure do
   #
   # In production all out tables exist in a 'schema' called `complete` so we
   # tell Rails to look for it's internal tables there.
-  config.active_record.schema_migrations_table_name = "complete.schema_migrations"
-  config.active_record.internal_metadata_table_name = "complete.ar_internal_metadata"
+  config.active_record.schema_migrations_table_name = "#{ENV["SQL_SERVER_SCHEMA_NAME"]}.schema_migrations"
+  config.active_record.internal_metadata_table_name = "#{ENV["SQL_SERVER_SCHEMA_NAME"]}.ar_internal_metadata"
 end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -72,4 +72,11 @@ Rails.application.configure do
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  # SQL server table config
+  #
+  # In production all out tables exist in a 'schema' called `complete` so we
+  # tell Rails to look for it's internal tables there.
+  config.active_record.schema_migrations_table_name = "complete.schema_migrations"
+  config.active_record.internal_metadata_table_name = "complete.ar_internal_metadata"
 end
--- a/doc/decisions/0008-variance-between-local-and-production-environments.md
+++ b/doc/decisions/0008-variance-between-local-and-production-environments.md
@@ -1,0 +1,37 @@
+# 8. Variance between local and production environments
+
+Date: 2022-07-27
+
+## Status
+
+Accepted
+
+## Context
+
+The application will connect to a Microsoft SQL server to store data. This
+presents a number of challenges in setting up a local development environment
+with complete parity to production.
+
+In production the application connects with the credentials of a database user
+who can only access the tables inside a 'schema', this guarantees the integrity
+of the wider database.
+
+Locally Rails assumes it can drop databases at will, this is difficult to
+replicate along with the login, database user and permissions to exactly mirror
+production.
+
+For example the users in production must never drop the database, as it contains
+schemas from other applications, locally the test database is routinely dropped
+to clear it.
+
+## Decision
+
+We will use MS SQL Server locally but forego the same schema and user
+restrictions as in production. We can use the per environment configuration to
+ensure production is setup as expected.
+
+## Consequences
+
+We usually seek parity between our local and production environments and whilst
+there is risk in this variance, it is far better to have a fast and efficient
+developer experience.
--- a/doc/microsoft-sql-server.md
+++ b/doc/microsoft-sql-server.md
@@ -1,0 +1,55 @@
+# Microsoft SQL Server
+
+As per this [ADR](./0007-use-the-academies-db-as-the-datastore.md), the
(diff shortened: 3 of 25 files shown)


script/no-docker/console
```diff
@@ -21,6 +21,9 @@ else
   echo "==> Updating..."
   script/no-docker/update
 
+  echo "==> Staring SQL Server container..."
```
jacksonj04, line 24:
```suggestion
  echo "==> Starting SQL Server container..."
```

script/no-docker/server
```diff
@@ -10,5 +10,8 @@ cd "$(dirname "$0")/../.."
 echo "==> Updating..."
 script/no-docker/update
 
+echo "==> Staring SQL Server container..."
```
jacksonj04, line 13:
```suggestion
echo "==> Starting SQL Server container..."
```

script/no-docker/setup
```diff
@@ -8,6 +8,9 @@ set -e
 # Enable the running of this script from any subdirectory without moving to root
 cd "$(dirname "$0")/../.."
 
+echo "==> Staring SQL Server container..."
```
jacksonj04, line 11:
```suggestion
echo "==> Starting SQL Server container..."
```

script/no-docker/update
```diff
@@ -7,6 +7,9 @@ set -e
 echo "==> Bootstrapping..."
 script/no-docker/bootstrap
 
+echo "==> Staring SQL Server container..."
```
jacksonj04, line 10:
```suggestion
echo "==> Starting SQL Server container..."
```

config/environments/production.rb
```diff
@@ -72,4 +72,11 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  # SQL server table config
+  #
+  # In production all out tables exist in a 'schema' called `complete` so we
+  # tell Rails to look for it's internal tables there.
+  config.active_record.schema_migrations_table_name = "complete.schema_migrations"
```
jacksonj04, line 80: Is this also the schema name on the service proper? Do we need to park this in an ENV variable?

doc/decisions/0008-variance-between-local-and-production-environments.md
```diff
@@ -0,0 +1,37 @@
...
+Accepted
+
+## Context
+
+The application will connect to a Microsoft SQL server to store data. This
+presents a number of challenges in setting up a local development environment
+with complete parity to production.
+
+In production the application connects with the credentials of a database user
+who can only access the tables inside a 'schema', this guarentees the integrity
```
jacksonj04, line 16:
```suggestion
who can only access the tables inside a 'schema', this guarantees the integrity
```

doc/decisions/0008-variance-between-local-and-production-environments.md
```diff
@@ -0,0 +1,37 @@
...
+production.
+
+For example the users in production must never drop the database, as it contains
+schemas from other applications, locally the test database is routinely dropped
+to clear it.
+
+## Decision
+
+We will use MS SQL Server locally but forego the same schema and user
+restrictions as in production. We can use the per environment configuraiton to
```
jacksonj04, line 30:
```suggestion
restrictions as in production. We can use the per environment configuration to
```

doc/decisions/0008-variance-between-local-and-production-environments.md
```diff
@@ -0,0 +1,37 @@
...
+ensure produciton is setup as expected.
```
jacksonj04, line 31:
```suggestion
ensure production is setup as expected.
```

doc/decisions/0008-variance-between-local-and-production-environments.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Consequences
+
+We usually seek parity between our local and production environments and whilst
+there is risk in this variance, it is far better to have a fast and efficient
+developer experience.
```
jacksonj04, line 37: 👍🏼

doc/microsoft-sql-server.md
```diff
@@ -0,0 +1,52 @@
...
+
+As this setup is relatively unconventional, this documentation may be of use.
+
+## Setup
+See the [Setup Microsoft SQL
+Server](../README.md#setup-microsoft-sql-server)
+in the readme for more information.
+
+## Running the server
+As SQL Server does not run natively on Mac OS we use the [official Docker
```
jacksonj04, line 15:
```suggestion
As SQL Server does not run natively on macOS we use the [official Docker
```

jacksonj04 (changes requested):
Found a few typos that want ironing out.

jkempster34 (commented):
I've just pulled this down and got it running.

README instructions are complete and easy to follow 💯 💯  (though I haven't run the scripts because they don't work for me... I should have a look at that)

config/environments/production.rb
mec: Yes, it implies the schema name in production and staging will be `complete`, take the point - an ENV variable would document this a little clearer 👍

jacksonj04 (approved):
🚢 🚀

PR #69, opened 2022-07-22
(195) Project tabs

jkempster34:
1. Add project subnavigation tabs
We want to use the GOV.UK tab component to navigate between the project task list page and the project information page. Move the "key information" into a partial, add partials to build the tab component, and add a basic controller and view for the new project information page.

2. Include authentication on the TasksController

[screen-capture (1).webm](https://user-images.githubusercontent.com/47089130/180465758-43689642-a63b-4637-83fc-f90b56b8ff94.webm)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -18,5 +18,5 @@ Rails.application.routes.draw do
     resources :tasks
   end
 
-  resources :project_information, path: "/projects/information"
+  get "/projects/:id/information", to: "project_information#show", as: :project_information
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -17,4 +17,6 @@ Rails.application.routes.draw do
   resources :projects do
     resources :tasks
   end
+
+  resources :project_information, path: "/projects/information"
 end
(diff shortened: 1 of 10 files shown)


jkempster34:
[Add Project Information to Project view](https://trello.com/c/VS0BOMTz/195-add-project-information-to-project-view)

config/routes.rb
```diff
@@ -17,4 +17,6 @@
   resources :projects do
     resources :tasks
   end
+
+  resources :project_information
```
jacksonj04, line 21: I think this would be better as `/project/information` rather than `/project_information`.

jacksonj04 (changes requested):
A little routing tweak I think to make the paths more intuitive, but otherwise looking good!

config/routes.rb
```diff
@@ -17,4 +17,6 @@
   resources :projects do
     resources :tasks
   end
+
+  resources :project_information, path: "/projects/information"
```
jacksonj04, line 21: Should this be a child of the projects resource, so we get `projects/{id}/information"`?

jacksonj04 (approved):

PR #72, opened 2022-07-25
(269) Implement first pass of at-a-glance project summary.

jacksonj04:
Provide a single 'at a glance' project summary for DO orientation, rendered above the task list and project information.

## Screenshots
localhost_3000_projects_3240408b-6cea-464f-8aa8-ae953bc90ab4


app/views/shared/_project_summary.html.erb
```diff
@@ -0,0 +1,29 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+
+    <span class="govuk-caption-l">URN <%= @project.urn %></span>
+    <h1 class="govuk-heading-l"><%= @project.establishment.name %></h1>
+
+    <%= govuk_summary_list(actions: false) do |summary_list|
+        summary_list.row do |row|
+          row.key { t('project.summary.proposed_opening_date.title') }
+          row.value { Date.parse(@project.conversion_project.proposed_opening_date).strftime("%-d %B %Y") }
```
jkempster34, line 10: Just having a quick glance through this.

Remembered from SE that we can define our date formats in a nice way like this ⌚  https://github.com/DFE-Digital/schools-experience/blob/a648a489751f9c5eff9519dbb7eaa4e51b0bf766/config/initializers/date_format.rb

jacksonj04: I _knew_ I'd seen this done before somewhere!

app/controllers/projects_controller.rb
```diff
@@ -48,7 +48,7 @@ def update
 
     @project.save
     flash[:notice] = I18n.t("project.update.success")
-    redirect_to project_path(@project)
+    redirect_to project_information_path(@project)
```
jkempster34, line 51: How come we are redirecting to the project info path now?

app/models/academies_api/conversion_project.rb
```diff
@@ -1,9 +1,10 @@
 class AcademiesApi::ConversionProject < AcademiesApi::BaseApiModel
-  attr_accessor :name_of_trust
+  attr_accessor :name_of_trust, :proposed_opening_date
 
   def self.attribute_map
     {
-      name_of_trust: "nameOfTrust"
+      name_of_trust: "nameOfTrust",
+      proposed_opening_date: "proposedAcademyOpeningDate"
```
jkempster34, line 7: 🤔 Wondering whether these models should be responsible for making things Ruby-fied so that we can work with them more easily. I suppose we would have to define a getter and setter like this:

```ruby
class AcademiesApi::ConversionProject < AcademiesApi::BaseApiModel
  attr_accessor :name_of_trust
  attr_writer :proposed_opening_date
  
  def proposed_opening_date
    Date.parse(@proposed_opening_date)
  end

  def self.attribute_map
    {
      name_of_trust: "nameOfTrust",
      proposed_opening_date: "proposedAcademyOpeningDate"
    }
  end
end

```

jkempster34 (approved):
Added a couple of thoughts, looks good to go though!

app/controllers/projects_controller.rb
jacksonj04: The logic is "nothing I can edit is displayed on the task list, it's all on the project information page".

It's very tenuous though. Mostly it meant I could leave tests as they were (since the relevant info is now on project information but not in the summary), and we can pass it over to research for "how do people expect editing to behave".

PR #77, opened 2022-07-27
(279) Move authentication to be included in ApplicationController

jacksonj04:
This change means that all controller actions are now automatically authenticated unless explicitly instructed otherwise using `skip_before_action` (as in `SessionsController`).

In the case of `SessionsController` we use `only` to explicitly list actions which should be excluded from authentication. Although in this case we list every action, it means that if we add a new action to the controller it will be protected unless we make the conscious decision to also exclude it.


jkempster34 (approved):
This looks good to me. Think its a good improvement

PR #78, opened 2022-07-27
Feature/271 add team leader

jkempster34:
### Add team lead to project on creation
The current assumption is that the team lead for a project is the user who creates the project. Assign them on project creation.

### Refactor tests now that team leader ID is a non-nullable field
We have added a non-nullable association to `team_leader` on the `project` record.

Update the project factory with a team_leader association. This uses a random email address so that when we create multiple `project`s in a spec, we don't see a record not unique error (due to the unique constraint on the email column in `user`).

Some specs were written before we started to use FactoryBot. Update these to use the `project` factory for the above reason.

### DIFF
--- a/spec/models/project_spec.rb
+++ b/spec/models/project_spec.rb
@@ -3,6 +3,16 @@ require "rails_helper"
 RSpec.describe Project, type: :model do
   describe "Columns" do
     it { is_expected.to have_db_column(:urn).of_type :integer }
+    it { is_expected.to have_db_column(:delivery_officer_id).of_type :uuid }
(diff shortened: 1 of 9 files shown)


spec/models/project_spec.rb
```diff
@@ -1,13 +1,21 @@
 require "rails_helper"
 
 RSpec.describe Project, type: :model do
+  before { mock_successful_api_responses(urn: any_args) }
```
jacksonj04, line 4: Does this need to happen across the whole spec?

jacksonj04 (approved):
One quick non-blocking question - looks good to me!

PR #85, opened 2022-07-29
Update changelog for v1

jacksonj04:


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+## [v1]
```
mec, line 9: Can we go `release-1`? I know we haven't, but if we are okay with it, we should change it now?

mec (approved):

PR #86, opened 2022-07-29
317 - More school details

jkempster34:
### Handle nested responses in the BaseApiModel
Some of the Academies API responses contain nested data. We have decided to surface these on our models, for example, mapping `establishmentType.name` to a `type` attribute on `AcademiesApi::Establishment`.

Update the BaseApiModel to handle this. We can provide the full access path in dot notation as the value in the `attribute_map`.


jkempster34:
[Add additional school details to Project Information](https://trello.com/c/1dgfTo6p/317-add-additional-school-details-to-project-information)

mec:
It would be great to have a screenshot on this!

Does this need rebaasing onto develop - I can't see the changelog, which we should update for this.

mec (approved):
Nice bit of work! Small, well ordered commits! :heart:

PR #92, opened 2022-08-03
(322) Swap to trust endpoint

jkempster34:
### 1. [Add Trust model and method to get trust from the Academies API](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/92/commits/d16338dbba94e73af5ff810a40200897fbc1ae18)
Moving forward, we want to get trust data from the trusts endpoint rather than relying on the conversion project.

Add Trust model and get_trust method to the client.

### 2. [Add trust UKPRN to Project model](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/92/commits/7cd4743ed4ee753fd913517e4b8dcd8abf2016f2)
Once we start using the trusts endpoint to get trust information, we will need to ask the user for the trust UKPRN on Project creation.

Add trust UKPRN to the project model along with validation to make sure that it isn't empty.

### 3. [Swap out the conversion project endpoint for the trust endpoint](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/92/commits/e1a4378e95806b22f314e0fec5a0f08b859d4b36)
We want to decouple the application from prepare for conversions. Currently, if there is no conversion project available, our app won't function - users won't be able to create a project.

We want to use the Academies API trust endpoint to collect trust information - which we currently get from the conversion project.

This removes the ConversionProject model, and adds a Trust model. In order to get a trust we need to introduce a new field of trust UKPRN to the new project form. We will no longer be able to get proposed opening date, so we will also need to introduce a field for that - which will come in the next commit.

This includes a small refactor of our feature specs, removing project validation failed tests - we have agreed that we can maintain a more balanced test suite by relying on our model specs for this. The `rails-controller-testing` Gem is needed in order to use the `render_template` method.

### 4. [Add trust details to the project information tab](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/92/commits/87d5af1320e7a0d5ea92faa0737f2a6721fc0f7f)
Now that we are using the trusts endpoint, we can add a trust details section to the project information tab.

### 5. [Add target completion date to Project](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/92/commits/a0f978787044d3fe4e7e6f52c0ffd0acb822f75b)
Previously, we pulled this in from the Conversion Project endpoint. Now that we've removed this, we need to ask the user for the target completion date when they create the Project.

Because the target completion date is always on the first of the month, remove the day field from the date input, and add some validation around this.

## Screenshots
image
image

## Checklist


jkempster34:
[Decouple the app from Prepare: ConversionProject](https://trello.com/c/A8vsELXL/322-decouple-the-app-from-prepare-conversionproject)

jacksonj04 (dismissed):
~I'm temporarily requesting changes (but not really) on this so it isn't merged whilst we're migrating to trunk based development.~

Migration done!

app/models/project.rb
```diff
@@ -42,4 +44,13 @@ def trust
 
     @trust = result.object
   end
+
+  private def first_day_of_month
```
mec, line 48: Nice, really good to cover this. 👍

app/views/projects/new.html.erb
```diff
@@ -10,6 +10,7 @@
 
   <%= form.govuk_text_field :urn, label: { tag: 'h3', size: 'm' } %>
   <%= form.govuk_text_field :trust_ukprn, label: { tag: 'h3', size: 'm' } %>
+  <%= form.govuk_date_field :target_completion_date, omit_day: true %>
```
mec, line 13: That was easy! 👍

mec: It seems to render the day as the first by default? Is that documentend on the form builder?

mec (approved):
Really great work! Ther are a couple of end of file editor things and a quesiton, nothing to stop this going out!

app/views/projects/new.html.erb
jkempster34: It is yep 

```
When mapped to a Date attribute, Ruby will automatically set the Day to be 1 when none is provided. This should be taken into account when displaying the data or using it for analysis.
```

PR #102, opened 2022-08-09
(370) Move the new note form onto its own page

jkempster34:
Currently, the new note form is underneath the project summary and subnavigation. This means that the error summary sits halfway down the page, whereas the GOV.UK guidance says it should sit at the top.

After a chat with our interaction designer, we have decided to move the new note form onto its own page. This solves the error message issue, and has an additional benefit that editing messages will be easier if we choose to implement that in the future.

This includes a few style updates:
- Add a cancel link to the new note form
- Rename the textarea label from "Notes" to "Enter note"
- Change the notes form submission button from "Add note" to "Save note"
- Remove the top separator of the list of notes

## Screenshots
image
image
image

## Checklist

### DIFF
--- a/app/views/notes/_notes.html.erb
+++ b/app/views/notes/_notes.html.erb
@@ -1,14 +1,10 @@
-<% if @notes.present? && @notes.any? %>
-  <% @notes.each_with_index do |note, index| %>
-    <% unless index == 0 %>
-      <hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
-    <% end %>
-
-    <span class="govuk-caption-m"><%= note.created_at.to_date.to_formatted_s(:govuk) %></span>
-    <h3 class="govuk-heading-m"><%= note.user.email %></h3>
-
-    <p class="govuk-body"><%= note.body %></p>
+<% @notes.each_with_index do |note, index| %>
+  <% unless index == 0 %>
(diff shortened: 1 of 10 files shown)


jkempster34:
[Improve the error summary for the new note form](https://trello.com/c/mTiHWyTS/370-improve-the-error-summary-for-the-new-note-form)

app/views/notes/_notes.html.erb
```diff
@@ -1,15 +1,14 @@
...
     <span class="govuk-caption-m"><%= note.created_at.to_date.to_formatted_s(:govuk) %></span>
-
     <h3 class="govuk-heading-m"><%= note.user.email %></h3>
 
     <p class="govuk-body"><%= note.body %></p>
   <% end %>
 <% else %>
-  <hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible">
-
   <h3 class="govuk-heading-m"><%= t("note.index.no_notes_yet") %></h3>
```
jacksonj04, line 13: This doesn't feel quite right as a header now that there isn't a form on the page. Maybe an [inset text](https://design-system.service.gov.uk/components/inset-text/) above the "add form" button would be cleaner?

jacksonj04 (changes requested):
Couple of tiny picky bits; substance of the change looks sound to me :shipit:

jacksonj04 (approved):
:shipit:

PR #105, opened 2022-08-11
(376) Mini content review

jkempster34:
### Small style/content updates based on designer feedback
Our interaction designer suggested that we could improve our hint text and use of text inputs on the new project page.

- Add hint text to the UKPRN and URN specifying what kind of input is expected.
- Use more appropriate widths for the input boxes. When we know the length of
the content, the guidance says we should use matching fixed-width inputs. 

### Update error messages as per mini review

We did a mini review of our error messages because so far we have been making them up on the fly, or using the Rails default messages.

Update the messages based on input from our content and interaction designers.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Improvements to the text inputs and hints on the new project form to make
   clearer what is expected from users.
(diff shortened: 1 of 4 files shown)


jacksonj04 (approved):
This is a great example of iterative improvement we should probably highlight as "we didn't get it perfect first time, so we went back and made it better".

giphy

PR #107, opened 2022-08-12
(365) YAML guidance and hint text

jkempster34:
### Fix spelling of received/receive
IDE highlighted this so I thought I might as well fix it while I was there

### Add hint and guidance to actions
We want to offer users some additional hints and guidance. 

Update the models to include the new columns, and update the conversions workflow to match.

### Render hints and guidance
For now, render the hint as part of the checkbox, and conditionally render the guidance.

The guidance (rendered as a details component) currently sits within the checkbox fieldset (because we have modelled actions checkboxes within a fieldset). I can imagine this will change as move move toward each action being a distinct thing.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,7 @@
 ## Getting started
 
 If this is your first time running the application, see the
-[`getting started documentation`](/doc/getting-started.md)
-for instructions.
+[`getting started documentation`](/doc/getting-started.md) for instructions.
 
 ## Running a server
 
@@ -26,8 +25,26 @@ See .env.example
 Use the [release process template in Trello](https://trello.com/c/8enGdMyy) to
 start a new release.
 
+## Including Markdown in the workflow
+
+We define task list workflows in YAML. To include Markdown while preserving
+newlines and whitespace we can use the _literal style_. This is indicated by a
+pipe (`|`) character followed by indented content. For example:
+
+```yaml
+guidance_text: |
+  # A h1 title header
+
+  ## Lists
+
+  This is a list:
+
+  * of multiple
+  * things
+  * and such
+```
+
 ## ADRs
 
 You can find the [ADRs](https://adr.github.io/) for this project in the
-[`doc/decisions`](/doc/decisions)
-folder.
+[`doc/decisions`](/doc/decisions) folder.
--- a/app/workflows/conversion.yml
+++ b/app/workflows/conversion.yml
@@ -6,7 +6,7 @@ sections:
       - title: Handover with Regional Delivery Officer
         actions:
           - title:
-              Recieve email with handover proforma and check it's saved in
+              Receive email with handover proforma and check it's saved in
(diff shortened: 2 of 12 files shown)


jkempster34:
[Add Hint text and Guidance text to Actions](https://trello.com/c/rDmkVR1w/365-add-hint-text-and-guidance-text-to-actions)

mec:
Booted this up, it's not on the card but I think we need to support white space and probably bullets as well - these are not currently working?

Screenshot 2022-08-16 at 10-57-32 Complete conversions transfers and changes

jkempster34:
> Booted this up, it's not on the card but I think we need to support white space and probably bullets as well - these are not currently working?
> 
> Screenshot 2022-08-16 at 10-57-32 Complete conversions transfers and changes

Just added some guidance to the README. How does that look?

app/models/action.rb
```diff
@@ -2,4 +2,7 @@ class Action < ApplicationRecord
   belongs_to :task
 
   default_scope { order(order: "asc") }
+
+  validates :guidance_summary, presence: true, if: -> { guidance_text.present? }
```
mec, line 6: I am not sure we need to validate this - I guess it is okay, but feels a bit too much - less code is always better in my head. I generally see validation as a user input thing, which this is not.

app/workflows/conversion.yml
```diff
@@ -66,6 +66,9 @@ sections:
       - title: Land questionnaire & land registry title plans
         actions:
           - title: Land questionnaire received
+            hint: Hello, I'm a hint
```
mec, line 69: Total nitpick, but I would like to see something closer to real copy here and evidence of all the formatting we support.

spec/services/task_list_creator_spec.rb
```diff
@@ -35,11 +35,20 @@
...
 
     it "creates actions from the workflow" do
-      section = Section.find_by(title: "Clear legal documents")
-      task = section.tasks.find_by(title: "Clear land questionnaire")
+      section = Section.find_by(title: "Starting the project")
+      task = section.tasks.find_by(title: "Understand history and complete handover from Pre-AB")
 
       expect(Action.count).to eql 6
-      expect(Action.where(title: "Action one", order: 0, task: task)).to exist
+      expect(
```
mec, line 42: Do you think this warrants a feature spec around helping users? Something like `Users can see hint text on Actions` and `Users can get additional guidnace on actions` - which could include tests for basic formating?

README.md
```diff
@@ -26,8 +25,25 @@ See .env.example
 Use the [release process template in Trello](https://trello.com/c/8enGdMyy) to
 start a new release.
 
+## Including Markdown in the workflow
```
mec, line 28: :heart: this! I would like to split this out into it's own file and link to it, but not now!

PR #109, opened 2022-08-16
(336) Regional delivery officer

jkempster34:
### Add regional deliver officer to project
We want to add a new role of `regional delivery officer` to the User model.

On project creation, a regional delivery officer must be selected from a drop down to be associated with the project.

### Show regional delivery officer on the project information tab
Now that a project has a regional delivery officer, we want to display this on the project information tab.

### Only show assigned projects to regional delivery officers
We only want regional delivery officer to see projects that they are associated with.

We don't policy scope the `show` view currently, so this only applies to the `index` view.

## Checklist

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -7,11 +7,13 @@ class Project < ApplicationRecord
   validates :trust_ukprn, presence: true, numericality: {only_integer: true}
   validates :target_completion_date, presence: true
   validates :team_leader, presence: true
+  validates :regional_delivery_officer_id, presence: true, allow_blank: false
   validate :establishment_exists, :trust_exists, on: :create
   validate :target_completion_date, :first_day_of_month
 
   belongs_to :delivery_officer, class_name: "User", optional: true
   belongs_to :team_leader, class_name: "User", optional: false
+  belongs_to :regional_delivery_officer, class_name: "User", optional: true
 
   def establishment
     @establishment || retrieve_establishment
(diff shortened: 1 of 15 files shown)


app/models/project.rb
```diff
@@ -7,11 +7,13 @@ class Project < ApplicationRecord
   validates :trust_ukprn, presence: true, numericality: {only_integer: true}
   validates :target_completion_date, presence: true
   validates :team_leader, presence: true
+  validates :regional_delivery_officer_id, presence: true, allow_blank: false
   validate :establishment_exists, :trust_exists, on: :create
   validate :target_completion_date, :first_day_of_month
 
   belongs_to :delivery_officer, class_name: "User", optional: true
   belongs_to :team_leader, class_name: "User", optional: false
+  belongs_to :regional_delivery_officer, class_name: "User", optional: true
```
jkempster34, line 16: This is really annoying, if its `optional: false`, it adds a validation error on the `regional_delivery_officer` attribute.

Having looked how people generally deal with this, they tend to just set `optional: true` and set the validation on the foreign key instead like I've done here

jkempster34:
[Model the Regional delivery officer on User](https://trello.com/c/5y9S0ZEZ/336-model-the-regional-delivery-officer-on-user)

app/models/project.rb
mec: I think there are two options for us here.

1. treat `regional_delivery_officer` like `delivery_officer` i.e allow it to not be set `optional: true` and no db constraint
2. assume `regional_delievery_officer` must always be set like `team_lead`

Our thinking has moved on slightly since we wrote this card, so I think it would be good to align this work now - shall we have a chat?

mec: We've discussed this and as we know more about what this will look like moving forward, we can put this in and come back and refactor.

mec (approved):
Really nice work @jkempster34! I would love to see the altering user work and then using that in Project as seperate commits, for me anything that can be small and self contained should be, small atomic commits that change as little as possible - but that is a me thing and should not stop this! 🚢

PR #111, opened 2022-08-19
(413) Destroy a projects notes along with the project

mec:
Leaving any of a Projects associated entities will create orphans.

Here we destroy a Projects notes along with the other associations
(Contacts and Sections (which in turn destroy Tasks and Actions)).

We also add some coverage around this behaviour to document it and make
it explicit.

https://trello.com/c/qqY8bE6H

## Checklist

### DIFF
--- a/spec/models/project_spec.rb
+++ b/spec/models/project_spec.rb
@@ -13,9 +13,33 @@ RSpec.describe Project, type: :model do
     before { mock_successful_api_responses(urn: any_args, ukprn: any_args) }
 
     it { is_expected.to have_many(:sections).dependent(:destroy) }
-    it { is_expected.to have_many(:notes) }
+    it { is_expected.to have_many(:notes).dependent(:destroy) }
     it { is_expected.to belong_to(:delivery_officer).required(false) }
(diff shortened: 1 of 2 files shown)


spec/models/project_spec.rb
```diff
@@ -42,6 +42,30 @@
       end
     end
 
+    describe "delete associated entities" do
```
jkempster34, line 45: This is a really handy test! (arguably much handier than the `Shoulda matchers`)

Noticed it's sitting in the `Validations` block, where it should probably be sitting in the `Relationships` block. The shoulda matchers could be updated as well while you are there

`bundle exec rspec spec/models/project_spec.rb --format documentation`

image

jkempster34 (changes requested):
Looks great 💯, just noticed that the test is sitting in the wrong place

spec/models/project_spec.rb
mec: I was aiming for it's own block, but I can do Relationships no problem! Thank you!

jkempster34 (approved):
💯 LGTM. 

Looks like you have an extra newline thats upsetting the linter!

PR #113, opened 2022-08-19
(404) Task level hints and guidance

jkempster34:
### Add hints and guidance to Task model
Tasks will have hints and guidance in the same form that Actions do.

Add the model attributes.

### Update the workflow and TaskListCreator to handle hints and guidance
The TaskListCreator and workflow need updating to handle task-level hints and guidance. Update them, following the sam structure as used for the action-level hints and guidance.

Because some of the keys of the workflow task are now optional, this also replaces any calls to `fetch` with `[]` in the TaskListCreator. `fetch` raises an error when the key isn't found, whereas `[]` does not.

### Render task hints and guidance
We want to render the task-level hints and guidance as GOV.UK Frontend content.

Follow the same pattern as the action-level guidance, and call the GovukMarkdown gem with the workflow markdown.

NOTE: I haven't updated the real workflow at all yet. Here it is working though 👇 
image

## Checklist

### DIFF
--- a/app/services/task_list_creator.rb
+++ b/app/services/task_list_creator.rb
@@ -15,7 +15,14 @@ class TaskListCreator
       section = Section.create(title: workflow_section.fetch("title"), order: index, project: project)
 
       workflow_section.fetch("tasks").each_with_index do |workflow_task, index|
-        task = Task.create(title: workflow_task.fetch("title"), order: index, section: section)
+        task = Task.create(
+          title: workflow_task.fetch("title"),
+          hint: workflow_task.fetch("hint", nil),
+          guidance_summary: workflow_task.fetch("guidance_summary", nil),
+          guidance_text: workflow_task.fetch("guidance_text", nil),
+          order: index,
+          section: section
+        )
 
         create_actions(workflow_task, task)
       end
--- a/app/workflows/schemas/task-list.json
+++ b/app/workflows/schemas/task-list.json
@@ -16,6 +16,9 @@
               "type": "object",
               "properties": {
                 "title": { "type": "string" },
+                "hint": { "type": "string" },
+                "guidance_summary": { "type": "string" },
+                "guidance_text": { "type": "string" },
                 "actions": {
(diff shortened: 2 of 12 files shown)


jkempster34:
[Tasks can have hint text and guidance](https://trello.com/c/S3HRHf9l/404-tasks-can-have-hint-text-and-guidance)

app/services/task_list_creator.rb
```diff
@@ -11,19 +11,25 @@ def call(project)
   end
 
   private def create_task_list_from_workflow(project, workflow)
-    workflow.fetch("sections").each_with_index do |workflow_section, index|
```
mec, line 14: You seem to have changed more than the optional values here? Raising a `KeyNotFound` error might be more helpful than returning `nil` and having unexpected behaviour elsewhere?

If you want to use fetch with a default you can:

`fetch("sections", nil)`

app/workflows/schemas/task-list.json
```diff
@@ -16,6 +16,9 @@
               "type": "object",
               "properties": {
                 "title": { "type": "string" },
+                "hint": { "type": "string" },
```
mec, line 19: I think I did the same with the `optional` key, do we need to declare the property as optional as well?

mec (changes requested):
I think for non-optinal keys we should raise an error, I think we discussed this before in some work I did?

app/workflows/schemas/task-list.json
jkempster34: The schema only has `required` and not `optional`.

app/services/task_list_creator.rb
jkempster34: Updated them all back to `fetch`, providing a default value of `nil` when the field isn't declared.

I've also added a new mock workflow without any of the optional values so we can test it explicitly. I could have used the existing mock workflow, but this seems more obvious to me 🤔

mec (approved):
🚢

PR #114, opened 2022-08-19
(419) Feature: tasks can be optional i.e. not applicable

mec:
## Changes

Not all Tasks apply to every Project, a simple example would be the Church fundung agreement is not applicable to non-fatih establishments.

For the moment Users want to be able to mark this kind of task as 'Not applicable' in the applicaiton. To do this we need two things:

1. the ability to control which Tasks can be made not applicable i.e. the Task is optional
2. the ability to store the state i.e is the Task applicable to the Project

We've settled on `optional` for controlling which Tasks can be 'not applicable' and `not_applicable` to store the state. We settled on `not_applicable` as this reflects the UI and the way users think about this, being true when checked.

Allow users to do this and showing not applicable tasks in the task list means users get a clear view of what Task are out standing.

## Screen shots
Screenshot 2022-08-22 at 11-00-21 Complete conversions transfers and changes

Screenshot 2022-08-22 at 11-00-32 Complete conversions transfers and changes

## Checklist

https://trello.com/c/l5nCjxDI

### DIFF
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -6,18 +6,29 @@ class TasksController < ApplicationController
 
   def update
     @task.actions.each do |action|
-      action_completed = action.id.in?(action_params[:completed_action_ids] || [])
+      action_completed = action.id.in?(task_params[:completed_action_ids] || [])
       Action.update(action.id, completed: action_completed)
     end
+
+    @task.update(not_applicable: not_applicable)
+    unset_all_actions if @task.not_applicable?
 
     redirect_to(project_path(@task.project))
   end
 
-  def action_params
-    params.require(:task).permit(completed_action_ids: [])
+  def task_params
+    params.require(:task).permit(:not_applicable, completed_action_ids: [])
   end
 
   private def find_task
     @task = Task.find(params[:id])
+  end
+
+  private def not_applicable
+    params.dig("task", "not_applicable")
+  end
+
+  private def unset_all_actions
+    @task.actions.update_all(completed: false)
   end
 end
--- a/app/views/tasks/show.html.erb
+++ b/app/views/tasks/show.html.erb
@@ -29,9 +29,15 @@
           <%= sanitize GovukMarkdown.render(action.guidance_text) %>
         <% end %>
       <% end %>
+    <% end %>
 
+    <% if @task.optional? %>
+      <%= form.govuk_check_box_divider %>
+      <%= form.govuk_check_box :not_applicable,
+                                true,
+                                label: { text: t("tasks.not_applicable") },
+                                checked: @task.not_applicable? %>
     <% end %>
   <% end %>
-
   <%= form.govuk_submit %>
 <% end %>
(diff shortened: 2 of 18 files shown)


app/controllers/tasks_controller.rb
```diff
@@ -10,14 +10,25 @@ def update
       Action.update(action.id, completed: action_completed)
     end
 
+    @task.update(not_applicable: not_applicable)
+    unset_all_actions if @task.not_applicable?
+
     redirect_to(project_path(@task.project))
   end
 
   def action_params
```
jkempster34, line 19: Wonder if this should be `task_params`? I named it `action_params` because it was related to actions, but maybe task would be a better name now, because it's the task that is N/A 🤔

app/views/tasks/show.html.erb
```diff
@@ -19,9 +19,15 @@
           <%= sanitize GovukMarkdown.render(action.guidance_text) %>
         <% end %>
       <% end %>
+    <% end %>
 
+    <% if @task.optional? %>
+      <%= form.govuk_check_box_divider %>
+      <%= form.govuk_check_box :not_applicable,
```
jkempster34, line 26: Shame the JS isn't working here. I suppose they would need matching keys

jkempster34 (commented):
Looks great.

Wonder if there is a way to validate that both "N/A" and an action can't be selected at the same time. It's a bit odd that when you submit a form that looks like this 👇 it would a) allow that to happen, and b) clear "Received" in the background.

image

app/views/tasks/show.html.erb
mec: Yep, I had a very quick think about if we could use the JS, but decided it could wait for now.

app/controllers/tasks_controller.rb
mec: I am not even sure we need the strong params as we are not doing mass assignment, but I take the point - `task_params` feels better, it then reads:

`task_params[:completed_aciton_ids]`

Which feels good.

mec:
@jkempster34 validation could work, I wonder if we are overthinking it a this stage? 

For me, marking a task not applicable implies we would not recieved the document in the first place so marking any of the checkboxes is redunant - it feels a little onerous to validate this and tell the user 'you can't mark this task as not applicable because you have checked other options'.

That said simply setting them to false is not ideal either - but it is the smaller piece of work right now.

I wonder if we leave it as is and see what users make of this first?

I feel like ideally we need to refactor this all into a 'Document clearing task' that lets us model this better than it is now - but I would want to test with users first to ensure the effort is justified?

What are your thoughts?

jkempster34:
> @jkempster34 validation could work, I wonder if we are overthinking it a this stage?
> 
> For me, marking a task not applicable implies we would not recieved the document in the first place so marking any of the checkboxes is redunant - it feels a little onerous to validate this and tell the user 'you can't mark this task as not applicable because you have checked other options'.
> 
> That said simply setting them to false is not ideal either - but it is the smaller piece of work right now.
> 
> I wonder if we leave it as is and see what users make of this first?
> 
> I feel like ideally we need to refactor this all into a 'Document clearing task' that lets us model this better than it is now - but I would want to test with users first to ensure the effort is justified?
> 
> What are your thoughts?

Hmm, I'm tempted to say that the validation would replace the unsetting of the Actions, and would be the roughly the same amount of code and complexity. However, I'm happy if you want to go ahead with what's here!

jkempster34 (approved):

mec:
@jkempster34 okay, let's get this in and aim to get user feedback as soon as we can. 👍

PR #115, opened 2022-08-19
Change Project task list to task list

mec:
It was noted in a design review that as the context of this page is 'a
project' there is not need to repeat that, so Task list works.

We are not as confident about 'Project informaiton' just yet though!

## Checklist


jacksonj04 (approved):
🚢

PR #116, opened 2022-08-22
Rename "delivery officer" to "caseworker"

jkempster34:
## Changes

The directorate now uses the name "caseworker" rather than "delivery officer".

We have modelled our standard user role as "delivery officer". This renames all
relevant code references, database columns, spec names, and views to
"caseworker".

## Checklist


jkempster34:
[Change Delivery officer to Regional case worker](https://trello.com/c/U3sESwoh/421-change-delivery-officer-to-regional-case-worker)

jacksonj04 (approved):
This looks good to me.

It doesn't have a data migration for the old `delivery_officer` column. I don't think this is a problem since we're not actually deployed anywhere, but maybe worth recording in the changelog that projects will lose any caseworker/DO data they may have had?

PR #118, opened 2022-08-22
(389) - Improve validation

jkempster34:
### Validate correct length of URN
Establishment URNs are always 6 digits long. We can provide some help to the users by showing a useful message if the URN is not the right length.

### Validate correct format of trust UKPRN
UKPRNs are always 8 digits long and start with a 1. We can provide some help to the users by showing a useful message if the UKPRN is not the right format.

### Validate that completion date is in the future
We think it's very unlikely that users will need to set a completion date that isn't in the future.

Add some validation to cover this.

_NOTE: The Trello card says to validate that caseworker cannot be blank on the update page. However, there is a spec which tests the unsetting of a caseworker, so this sounds like an intentional thing, and I haven't added the validation_

## Screenshots
image

## Checklist

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -9,7 +9,7 @@ class Project < ApplicationRecord
   validates :team_leader, presence: true
   validates :regional_delivery_officer_id, presence: true, allow_blank: false
   validate :establishment_exists, :trust_exists, on: :create
-  validate :target_completion_date, :first_day_of_month
+  validate :first_day_of_month, :trust_ukprn_is_correct_format
 
   belongs_to :caseworker, class_name: "User", optional: true
   belongs_to :team_leader, class_name: "User", optional: false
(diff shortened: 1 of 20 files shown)


jkempster34:
[Add new validation](https://trello.com/c/BSINLcI0/389-add-new-validation)

app/models/project.rb
```diff
@@ -68,4 +70,12 @@ def trust
       errors.add(:trust_ukprn, :must_be_correct_format)
     end
   end
+
+  private def target_completion_date_is_in_the_future
+    return if target_completion_date.nil?
+
+    unless target_completion_date.future?
+      errors.add(:target_completion_date, :must_be_in_the_future)
```
mec, line 78: This is cool! You just add the symbols for the active record error locales. 👍

mec (approved):
Thanks for taking the time to do this. I am not sure I have confidence that all URN and UKPRNs follow this yet, but I trust that you do!

I've added the create context to the completion date just to stop edits being invalid if we need to update any projects.

PR #120, opened 2022-08-23
(424) Feature: change the project creation journey

mec:
## Changes

We decided that Regional delivery officers will create (aka Handover) projects from the prepare side of the service to us. For our MVP this will be a manual process with automating it coming later.

Originally we had the Team Lead creating projects and assigning them to Caseworkers, here we make the changes to move the responsibility for creating the project to Regional delivery officers.

In the spirit of 'leave the code better than you found it' I've made some changes to some specs to try and follow the general principle of features testing the happy path and tighter unit test for everything else.

I think we have some work to do to make the assignment jounrney more intuitive for Team Leads, but that will come later.

https://trello.com/c/90qSKSQt

## Checklist


jacksonj04 (approved):
It's a ✅ from me following the Huddle about team leader assignment!

🚢

PR #123, opened 2022-08-25
(447) Change migration adding RDOs to work with existing data

jacksonj04:
## Changes

The previous form of this migration tried to add a non-nullable `regional_delivery_officer_id` column, which fails in databases where there is existing data.

Making this relationship nullable allows migrations to be performed.

## Checklist


mec (approved):
These database constraints on these always bothered me! Thanks for sorting this. 🚢

PR #125, opened 2022-08-26
(405) Chore: allow the app to run background jobs

mec:
## Changes

Take the basic steps to add Sidekiq and Redis to the application ahead of setting up our infastructure.

As we have no background jobs right now, this is minimal and serves no purpose other than laying the ground work for us to build upon.

https://trello.com/c/kOw29Jxw

## Checklist


doc/decisions/0011-use-sidekiq-and-redis-to-process-background-jobs.md
```diff
@@ -0,0 +1,37 @@
...
+Accepted
+
+## Context
+
+Most web applications have a need to send email notifications and we can already
+see the points in out journey where we would like to do so.
+
+Best practice when sending email is to do so on a background job, this prevents
+the application having to wait for the email to be sent in order to continue
+exexution.
```
jacksonj04, line 16:
```suggestion
execution.
```

doc/decisions/0011-use-sidekiq-and-redis-to-process-background-jobs.md
```diff
@@ -0,0 +1,37 @@
...
+
+The most common way to approach this is to use Rails own
+[ActiveJob](https://guides.rubyonrails.org/active_job_basics.html) along with
+[Sidekiq](https://sidekiq.org/) which depends on Redis.
+
+## Decision
+
+We will adopt the stand approach to this and setup ActiveJob, Sidekiq and Redis.
```
jacksonj04, line 24:
```suggestion
We will adopt the standard approach to this and setup ActiveJob, Sidekiq and Redis.
```

doc/decisions/0011-use-sidekiq-and-redis-to-process-background-jobs.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Consequences
+
+- Standard approach
+- Removes the need to return to our production environments later to add
+  background jobs
+- Increased complexity
+- Redis will be a new dependency
+- There is risk we will not find the space in delivery to utilise the background
+  jobs in the delivery
```
jacksonj04, line 37:
```suggestion
  jobs
```

.env.example
```diff
@@ -31,3 +31,6 @@ ACADEMIES_API_TIMEOUT=0.6
 #
 SENTRY_ENV=SENTRY_ENV # the environment production, pre-production/staging or development
 SENTRY_DSN=SENTRY_DSN # the DSN for the project in Sentry see: https://docs.sentry.io/platforms/ruby/configuration/options/#environment-variables
+
+# Redis
+REDIS_URL=REDIS_URL
```
jacksonj04, line 36: Could this be the default URL for running locally under Docker?

docker-compose.local.yml
```diff
@@ -0,0 +1,20 @@
...
+  redis:
+    image: redis:6
+    ports:
+      - 6379:6379
+    volumes:
+      - redis-data:/data
+
+volumes:
+  sql-server-data:
+  redis-data:
```
jacksonj04, line 20: Is it intentional that we have a new `docker-compose.local` which does this? It sort of makes sense for local dev to have this but not prod (because we don't yet know how that will have Redis provisioned), but this also means we can't do things like run tests against it unless the test runner provisions its own.

jacksonj04 (commented):
Couple of tiny tweaks and one implementation question!

.env.example
mec: I like the way the example has no values, the Redis URL could be any number of values so I am not sure I want to propose one at this level?

docker-compose.local.yml
mec: It is intentional - my understanding is we have devs who use the script, those that run everything in Docker and devs that run hybrid.

The scripts install Redis natively so we only want to start the MSSQL server in Docker as it cannot be run natively on MacOS so we have `docker-compose.database.local.yml`

For all in Docker folks we have `docker-compose.yml` which starts everything

And for folks that want to run hybrid (i.e. me! :)) have `docker-compose.local.yml` to start the backing services but not the application.

In CI we use the `docker-compose.ci.yml` which will provision the Redis instance.

Does that anwswer the question?

jacksonj04 (approved):
📦 🚀

PR #126, opened 2022-08-26
(412) Add organisations to contacts

jacksonj04:
## Changes

Contacts can now be assigned to an organisation, and will be grouped accordingly.

## Checklist


jkempster34 (approved):
As discussed separately, if the edit and new forms will have a few differences, then it makes sense for them to be separate

💯 💯 💯

PR #132, opened 2022-09-01
Chore/416 additional user attributes

jkempster34:
### Add first name and last name to the user model
Currently, we refer to users throughout the app by email. In order to be more user friendly, we want to use the user's names instead. 

This adds first name and last name to the user model. The migration doesn't specify that these field are required, so no existing user record will be impacted.

This also adds "column" specs to the User model specs in order to match the other models.

### Show user's first and last names on the project information tab
Now that the User model has first and last name, we want to display this alongside email on the project information tab.

This adds email as a `mailto:` link after the full name.

It also adds a helper to DRY up the logic around displaying a name. So that this is backwards compatible for pre-existing users, display the email address for users who don't have a name.

_See screenshot below for all variations (user not assigned, user with an email only, user with name and email) 📸_

image

### Add a UserImporter service
Currently, we add users one at a time using the `users:create` task. To make creating users easier, we now want the ability to bulk add users from a CSV file.

This adds a UserImporter service that takes a CSV file path, and upserts any specified users.

### Add a Rake task to call the UserImporter service
We want a way for developers to easily call the UserImporter service.

This adds a `user_importer:import` task, along with some README guidance.

### Collect Azure Active Directory user ID on sign in
Currently, we link our users to Azure Active Directory through email address. 

We think it's sensible to start collecting the Azure Active Directory user ID as well, because email address has the potential to change.

We aren't actually doing anything except collecting the ID for now.

## Checklist


jkempster34:
[Add additional attributes to User](https://trello.com/c/PRlyoViR/416-add-additional-attributes-to-user)

app/models/user.rb
```diff
@@ -3,4 +3,8 @@ class User < ApplicationRecord
   has_many :notes
 
   scope :caseworkers, -> { where(team_leader: false).and(where(regional_delivery_officer: false)) }
+
+  def full_name
+    "#{first_name} #{last_name}" if first_name.present? && last_name.present?
```
jkempster34, line 8: Does this sound right? Don't show name if either are missing?

app/services/user_importer.rb
```diff
@@ -0,0 +1,16 @@
+require "csv"
+
+class UserImporter
+  def call(csv_path)
+    @csv_path = csv_path
+    upsert_users
+  end
+
+  private def upsert_users
```
mec, line 9: I wonder if we should wrap this in a transaction so that if there is a failure, we roll back?

https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html

app/services/user_importer.rb
```diff
@@ -0,0 +1,16 @@
...
+    CSV.foreach(@csv_path, headers: true) do |row|
+      User.find_or_create_by(email: row[:email]) do |user|
+        user.assign_attributes(row)
```
mec, line 12: This is my fault for hinting at `find_or_create_by`:

This should be able to be run against an existing user and update their attributes, because `email` has a unique constraint on it this `find_or_create_by` and `assign_attributes` does not appear to work here, we might have to do this manually as we cannot use `upsert` with the SQL adaptor?

As pseudo code:

```
for each row in the csv:
   find the User by email
   if one is found
     update the attributes but not email
   else
     create and save a new User
  end
end
```

mec (changes requested):
We should make the import update, we don't want to get into potentially juggling lists of users, we want to run the import and not worry about being supplied duplicates.

This is my fault for leading you to `create_or_find_by`!

app/services/user_importer.rb
jkempster34: I actually had a spec covering this, but it turns out the spec wasn't very good 🤦 Thanks for running this and checking!

We can still use `find_or_create_by` but not with a block, we need to do it like this 🤷 

```ruby
user = User.find_or_create_by(email: row[:email])
user.assign_attributes(row)
user.save
```

Fixed this so it should all work as expected now!

mec (approved):
Really nice work! I love the docs and how this commit history reads! ⭐ 🚢

PR #136, opened 2022-09-02
(486) Create Azure Container Registry

Stretch96:
## Changes

* We need somewhere to store the image that will be created by Github
  actions
* This uses terraform to launch an Azure Container Registry within a
  given resource group
* `skip_provider_registration` has been set, as we will be using
    accounts that do not have the permissions to register new Azure
    resource providers. https://github.com/hashicorp/terraform-provider-azurerm/issues/7206

## Checklist

### DIFF
--- a/terraform/providers.tf
+++ b/terraform/providers.tf
@@ -1,3 +1,4 @@
 provider "azurerm" {
   features {}
+  skip_provider_registration = true
 }
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -1,0 +1,14 @@
+variable "environment" {
+  description = "Environment name. Will be used along with `project_name` as a prefix for all resources."
+  type        = string
+}
+
+variable "project_name" {
+  description = "Project name. Will be used along with `environment` as a prefix for all resources."
+  type        = string
+}
+
+variable "resource_group" {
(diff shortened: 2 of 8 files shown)


terraform/providers.tf
```diff
@@ -1,3 +1,4 @@
 provider "azurerm" {
   features {}
+  skip_provider_registration = true
```
mec, line 3: Is it worth calling this out in the commit message as something that threw us initially and what this means?

terraform/variables.tf
```diff
@@ -0,0 +1,13 @@
+variable "environment" {
+  description = "Environment name. Will be used along with `project_name` as a prefix for all resources."
+  type        = string
+}
+
+variable "project_name" {
```
mec, line 6: Why is this not declared as a string?

mec (changes requested):
Two quesitons, otherwise I followed along.

terraform/variables.tf
Stretch96: Ah yup - changed now 👍

terraform/providers.tf
Stretch96: Yup okie ...

mec (approved):

PR #138, opened 2022-09-05
(481) Fix: Guard against an Action hint being nil

mec:
## Changes

The GovukMarkdown.render method expects a string and cannot handle nil.

Action hints are optional so we have lots of instances where the
`Action.hint` is nil.

Adding this guard clause means the view can still be rendered where
there are actions without a hint.

https://trello.com/c/FJaXxrXz

## Checklist


jkempster34 (approved):
👍

PR #140, opened 2022-09-05
(451) Add optional handover comment field to the new project form

jkempster34:
## Changes

The handover proforma and KIM allow users to add a handover note. We want to replicate this by adding an optional "handover comment" field to the new project form. This will create a new `Note` record for the `Project`.

## Checklist

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -19,9 +19,10 @@ class ProjectsController < ApplicationController
   end
 
   def create
-    @project = Project.new(project_params)
+    @note = Note.new(**note_params, user_id: user_id)
+    @project = Project.new(**project_params, regional_delivery_officer_id: user_id, notes_attributes: [@note.attributes])
+
     authorize @project
-    assign_regional_delivery_officer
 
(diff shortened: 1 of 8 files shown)


jkempster34:
[Collect optional Project note on create](https://trello.com/c/vMdsJt0x/451-collect-optional-project-note-on-create)

app/controllers/projects_controller.rb
```diff
@@ -19,9 +19,10 @@ def new
   end
 
   def create
-    @project = Project.new(project_params)
+    @note = Note.new(**note_params, user_id: user_id)
+    @project = Project.new(**project_params, regional_delivery_officer_id: user_id, notes_attributes: [@note.attributes])
```
jacksonj04, line 23: I wonder if it's worth slicing this tidying up of RDO assignment into its own commit? That's a very weak wonder though.

jacksonj04 (approved):
Looks shiny to me. One tiny ponder about commit granularity, but I don't feel strongly enough to be particularly worried about it given everything is somewhat related.

PR #141, opened 2022-09-06
(417) Add accessible-autocomplete to caseworker dropdown

jacksonj04:
## Changes

When selecting a caseworker, if the user has Javascript enabled the default select dropdown is replaced with an accessible autocomplete input.

## Checklist


jacksonj04:
I did mull over that one, I reckon at this point it's cleaner to do it inline and save the resource call.

PR #145, opened 2022-09-06
(453) Notes can be formatted, edited and deleted

jacksonj04:
## Changes

Notes can:

- [x] Have rich formatting including line breaks
- [x] Be edited by the user who created them
- [x] Be deleted by the user who created them

## Checklist

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -44,6 +44,20 @@ class NotesController < ApplicationController
     end
   end
 
+  def destroy
+    @note = Note.find(params[:id])
+    authorize @note
+
+    @note.destroy
+
+    redirect_to project_notes_path(@project), notice: I18n.t("note.destroy.success")
+  end
+
+  def confirm_destroy
+    @note = Note.find(params[:note_id])
+    authorize @note
+  end
+
   private def find_project
     @project = Project.find(params[:project_id])
   end
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -25,6 +25,25 @@ class NotesController < ApplicationController
     end
   end
 
+  def edit
+    @note = Note.find(params[:id])
+    authorize @note
+  end
+
+  def update
+    @note = Note.find(params[:id])
+    authorize @note
+
+    @note.assign_attributes(note_params)
+
+    if @note.valid?
+      @note.save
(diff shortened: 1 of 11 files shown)


app/controllers/notes_controller.rb
```diff
@@ -1,15 +1,19 @@
 class NotesController < ApplicationController
   before_action :find_project
   before_action :find_notes, only: :index
+  after_action :verify_authorized
```
jkempster34, line 4: Does this work if we put it in ApplicationController 🤔 I suppose not because we have controller with no auth...

spec/requests/notes_controller_spec.rb
```diff
@@ -87,4 +87,161 @@
       end
     end
   end
+
```
jkempster34, line 90: These specs are really detailed!

app/views/notes/_notes.html.erb
```diff
@@ -6,5 +6,9 @@
   <span class="govuk-caption-m"><%= note.created_at.to_date.to_formatted_s(:govuk) %></span>
   <h3 class="govuk-heading-m"><%= note.user.full_name || note.user.email %></h3>
 
-  <p class="govuk-body"><%= note.body %></p>
+  <%= sanitize GovukMarkdown.render(note.body) %>
```
jkempster34, line 9: I think you will need to rebase this work. Could we use the new method I added `render_markdown` here?

jkempster34 (approved):
Looks and works really great to me 👍 

Left a couple small comments for your consideration!

app/controllers/notes_controller.rb
mec: Sorry to jump in! Whilst this should go into ApplicationController to stop us forgetting, I have just seen how many issues this will raise - we need to swing back over our policies and clear some deby there!

Let's get this work in and write up more detailed work to cover our policies in more depth.

mec: Sorry to jump in! Whilst this should go into ApplicationController to stop us forgetting, I have just seen how many issues this will raise - we need to swing back over our policies and clear some deby there!

Let's get this work in and write up more detailed work to cover our policies in more depth.

jacksonj04: Agree it _should_ go in `ApplicationController`, but also agree that we need to solve policies for all other object types and actions first.

PR #146, opened 2022-09-06
(478) Contacts can be deleted

jkempster34:
### Add destroy action to the ContactsController
We want users to be able to delete contacts. Add the destroy action to the ContactsController

### Add `confirm_destroy` action to the ContactsController
We want to show an "Are you sure?" page when a user selects to delete a contact.

Add a `confirm_destroy` action to the ContactsController that will be used to show the "Are you sure?" page.

### Enable user to delete contacts
This adds a delete contact confirmation page, which is accessed through a link on the edit contacts page.

## Screenshots
image

image

image

## Checklist

### DIFF
--- a/app/views/contacts/edit.html.erb
+++ b/app/views/contacts/edit.html.erb
@@ -20,6 +20,7 @@
       <%= form.govuk_phone_field :phone, width: 10 %>
 
       <%= form.govuk_submit t("contact.edit.save_contact_button") do %>
+        <%= govuk_button_link_to "Delete", project_contact_delete_path(@project, @contact), warning: true %>
         <%= govuk_link_to "Cancel", project_contacts_path(@project) %>
       <% end %>
     <% end %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -140,6 +140,10 @@ en:
       success: Contact has been updated successfully
     destroy:
       success: Contact has been deleted successfully
+    confirm_destroy:
+      title:
+        Are you sure you want to delete %{contact_name}?
+      guidance: This will remove the contact for %{contact_title} called %{contact_name} from the contacts list.
     details:
       name: Name
       title: Role
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -138,6 +138,8 @@ en:
       save_contact_button: Save contact
     update:
       success: Contact has been updated successfully
+    destroy:
+      success: Contact has been deleted successfully
     details:
       name: Name
       title: Role
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   Javascript enabled
 - The date and time that a caseworker is first assigned to a project is
   recorded. Any subsequent assignments will not be recorded.
+- Contacts can be deleted via a button on the edit contact page.
 
 ### Changed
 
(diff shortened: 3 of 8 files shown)


jkempster34:
[Make contacts deletable](https://trello.com/c/krKvgOwU/478-make-contacts-deletable)

app/views/contacts/edit.html.erb
```diff
@@ -20,6 +20,7 @@
       <%= form.govuk_phone_field :phone, width: 10 %>
 
       <%= form.govuk_submit t("contact.edit.save_contact_button") do %>
+        <%= govuk_button_to "Delete", project_contact_path(@project, @contact), method: :delete, warning: true %>
```
jacksonj04, line 23: This doesn't seem to have any warning or confirmation, I think it may rely on some Javascript we don't have.

Would it be safer to have it just open a new view offering a delete button and back button?

jacksonj04 (changes requested):

app/views/contacts/edit.html.erb
mec: Yep, `warning: true` is part of UJS which we do not have. 

There might be a wider conversation here about the pattern we use to delete an entity, we either use something quick but JS based or build a more response/request type thing that we can hopefully reuse - it would be worth chatting to @titlescreen about this - I think he has already thought about this and picked the later.

There is a balance to strike here, effort vs value, I would like us to see if we can build a reusable request/response delete but it shouldn't become a time sink - happy to talk this through.

jkempster34: I think `warning: true` here is just the red colour of the button.

I didn't actually notice, but the prototype has a delete confirmation screen, I'm just in the middle of doing that now.

titlescreen: In the prototype I had put in a delete contact flow where the user would pick change and then on the edit contact page  there would be a delete option at the bottom. This would point them to an are you sure screen. 

I've added it to the contacts [Lucid chart](https://lucid.app/lucidspark/cbfc31ef-8d60-4a81-bdd0-da018d78cbc4/edit?view_items=S-~K8HK3EkYY&invitationId=inv_69b4f3cc-535d-4d7e-9176-16ddc010a0ec#)

image
image

jkempster34: @titlescreen Yep, that's what I've followed! Screenshots added to the description.

@jacksonj04 @mec How does that look? not sure about making it reusable

jacksonj04: That looks like what I would expect! I don't know how reusable we can make it tbh, this might be one of those things like the edit contact screen where it feels like it _should_ be very DRY but actually there's enough nuance in the template etc it's easier to just have each class's own flow.

mec: > I think `warning: true` here is just the red colour of the button.

Yep, sorry Joe, I am getting confused between delete conversation and jumping to conclusions!

config/locales/en.yml
```diff
@@ -138,6 +138,12 @@ en:
       save_contact_button: Save contact
     update:
       success: Contact has been updated successfully
+    destroy:
+      success: Contact has been deleted successfully
+    confirm_destroy:
+      title:
+        Are you sure you want to delete %{contact_name}?
+      guidance: This will remove the contact for %{contact_title} called %{contact_name} from the contacts list.
```
jacksonj04, line 146: This language doesn't feel good in my brain. Happy for it to ship now and we revisit it later with a content designer!

jacksonj04 (approved):
Does what it says on the tin. Into the wild with it!

config/locales/en.yml
jkempster34: Agreed!

PR #149, opened 2022-09-07
Fix: do not use a css compressor in production

mec:
## Changes
This known issue with sassc and the GOV.UK kit [1] can be avoided by explicitally setting `config.assets.css_compressor` to nil and then enabling compression in sassc with:

`config.sass.style`

[1] https://github.com/alphagov/govuk-frontend/issues/1350#issuecomment-493129270

## Checklist


Stretch96 (approved):
lgtm 🚀

PR #150, opened 2022-09-07
Fix: Remove unnecessary dependabot config to try address intermittent test failures

mec:
## Changes

We were seeing some internittent issues that linked to this, but are not
100% confident this is the fix.

We do know that github-actions for dependabot does not use
versioning-strategy[1] so we remove it.

Also removed the quotes for consistency with the rest of the file.

[1] https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy
## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,12 +18,11 @@ updates:
     open-pull-requests-limit: 10
     versioning-strategy: auto
     target-branch: main
-  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


jacksonj04 (approved):
I'm unclear on if this will solve the problem, but I'm also happy with reducing the amount of unnecessary stuff in config files so 👍🏼

PR #160, opened 2022-09-09
(500) Update task list content and layout

jkempster34:
### Use two thirds layout for task list
We want to use the GOV.UK two-thirds layout for the task list. This is recommended to stop lines of text getting so long that they are difficult to read.

### Hide "Actions" legend and update spacing
This removes the "Actions" legend for the checklist, and increases the whitespace beneath the task title and guidance to maintain the separation.

### Align the Action checkbox guidance with the checkbox hint
Include the guidance details component in the Proc that is passed into the Action hint in order to align them.

### Bold the labels of the action checkboxes
The labels of the action checkboxes are becoming lost amongst the content that has been added to the task page as hint text and guidance.

Bold the labels so that they stand out.

### Add spacing below each checkbox
So that each checkbox is distinct and easy to visually separate, add some additional spacing between each. 

This makes the legal documents section in its current form (with the divider) look a bit odd, but this will be tackled in the future.

### Update task title styling
Update the styling of the task title to match the prototype. This includes a caption with the establishment name.

### Update the content for the task list Project kick-off section
The content for this section has been reviewed.

### Make sure "Opens in new tab" links open in a new tab
Currently, any links we pass through the renderer do not open in a new tab (despite saying that they do). 

We use a custom Markdown renderer with GOV.UK styling (GovukMarkdown) that is derived from Redcarpet. Redcarpet allow us to configure what attributes their renderer adds to links.

This monkey patches the GovukMarkdown library to adds `target: "_blank"` option for link attributes.

This also swaps DRYs up any calls to `sanitize` into a `render_markdown` helper. `sanitize` strips out `target` attributes, so we need to explicitly allow these. The previous fix for enabling hint styling no longer works, so move this into the helper as well.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   Instead, users must be explicitly given the caseworker role.
 - Improved the layout of the new project page.
 - Improve the layout of the task page.
+- Update the content for the "Project kick-off" section of the task list.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Tasks that have actions without hint text can now be rendered.
+- Links on the task list now open in a new tab.
 
 ## [Release 2][release-2]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The caseworker role is no longer implied from the lack of any other roles.
   Instead, users must be explicitly given the caseworker role.
 - Improved the layout of the new project page.
+- Improve the layout of the task page.
 
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


jkempster34:
[Update task view to reflect latest layout and content (finishing off the Project Kick off section)](https://trello.com/c/CHQ2xrMX/500-update-task-view-to-reflect-latest-layout-and-content-finishing-off-the-project-kick-off-section)

jacksonj04 (approved):
All the visual changes look fine to me code-wise, and I _think_ they match the design.

We could possibly tighten up the markdown testing to be a bit more end-to-end, eg by passing a string into the function which we would in reality (like `<a href="foo" target="_blank">opens in a new tab</a>`) and verifying we still get the attribute back even after it's gone through markdown rendering and sanitising.

That said, it works just fine if you put HTML into the workflow files!

jkempster34:
> All the visual changes look fine to me code-wise, and I _think_ they match the design.
> 
> We could possibly tighten up the markdown testing to be a bit more end-to-end, eg by passing a string into the function which we would in reality (like `<a href="foo" target="_blank">opens in a new tab</a>`) and verifying we still get the attribute back even after it's gone through markdown rendering and sanitising.
> 
> That said, it works just fine if you put HTML into the workflow files!

👍 I've updated the helper_specs to use the actual GovukMarkdown library, rather than mocking it.

I think there is a bit of advice I've seen: "Don't mock what you don't own" that applies here

PR #165, opened 2022-09-12
(487) Feature: add region to project information

mec:
## Changes

Whilst not directly helpful right now, we are in Regional Services Division and have regional delivery officers and regional caseworkers - it seems likely that the concept of 'region' will be useful.

Here we are exposing the establishment (school) region which we get from get informatiion about schools via the academies API.

https://trello.com/c/x1coxpiz

## Checklist


jacksonj04 (approved):
One tiny typo in the name of `spec/features/users_can_view_a_projet_spec.rb` (should be `users_can_view_a_project_spec.rb`), but otherwise let's ship it!

PR #166, opened 2022-09-13
(545) Fix: show assigned user names

mec:
## Changes

This behaviour was rebased away during #165 this work restores it and adds coverage to prevent it happening again.

I've refactored the feature spec that covers for two reasons:

1. I had broken it
2. I was finding it very hard to parse

https://trello.com/c/J5Xk2cD6

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Tasks that have actions without hint text can now be rendered.
 - Links on the task list now open in a new tab.
+- The name and email address of assigned users is shown on the project
+  information tab.
 
 ## [Release 2][release-2]
 
--- a/app/views/project_information/show/_information_list.erb
+++ b/app/views/project_information/show/_information_list.erb
@@ -3,15 +3,15 @@
   <%= govuk_summary_list(actions: false) do |summary_list|
     summary_list.row do |row|
       row.key { t('project_information.show.project_details.rows.caseworker') }
-      row.value { @project.caseworker&.email or t('project_information.show.project_details.rows.unassigned') }
+      row.value { display_name(@project.caseworker) }
     end
     summary_list.row do |row|
       row.key { t('project_information.show.project_details.rows.team_lead') }
-      row.value { @project.team_leader&.email or t('project_information.show.project_details.rows.unassigned') }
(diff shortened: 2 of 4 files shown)


jkempster34 (commented):
Looks like the spec was rebased away too!

https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/737df5d7777394fab90f0da86e7195ff98818d25/spec/features/users_can_view_project_information_spec.rb#L37

mec:
> Looks like the spec was rebased away too!
> 
> https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/737df5d7777394fab90f0da86e7195ff98818d25/spec/features/users_can_view_project_information_spec.rb#L37

Is that not this helper method that I updated:

https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/2f68a1a8426d03823870b6d5c9789062dac971e5/spec/features/users_can_view_project_information_spec.rb#L66

jkempster34:
> > Looks like the spec was rebased away too!
> > https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/737df5d7777394fab90f0da86e7195ff98818d25/spec/features/users_can_view_project_information_spec.rb#L37
> 
> Is that not this helper method that I updated:
> 
> https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/2f68a1a8426d03823870b6d5c9789062dac971e5/spec/features/users_can_view_project_information_spec.rb#L66

Kind of - it looks like you have updated an older version of the helper method, and not the most recent method that captured the email addresses and full names

mec:
@jkempster34 I went back to this and finally saw what you were trying to show me! 

For me the fact I wasn't seeing it speaks to me of the issue, it may be personal, but I found the feature spec really hard to read. I think it was the helper was overburdened and so the language was breaking down.

I've refactored this to include the link checks and hopefully make it a little more readable, I've removed the label check as for me that feels like testing Rails behaviour and I am not sure there is much value there.

Let me know what you think and sorry for the confusion, I blame working on the train! :)

jkempster34 (approved):
Think I preferred it more explicit 🤔 but happy to defer to your decision.

It definitely reads a bit easier 💯

mec:
@jkempster34 okay, lets at least get the bug fixed and keep chipping away at improving the test suite for everyone! 👍

PR #171, opened 2022-09-15
(557) Chore: rename trust to incoming trust

mec:
## Changes

 When we modelled `trust` we kept the naming simple, we've since learnt
 that `trust` is a keyword in Ruby in certain conditions.
    
A conversion represents a school becoming an academy by joining a Trust,
we use the term 'incomming trust' to describe to which trust the school is
joining.
    
Later when adding transfers we will be storing both the
incomming and outgoing trust i.e the new trust and the trust the school
is leaving - changing this column name now solves the Ruby name clash and
sets us up nicely for a time when we introduce the concept of
`outgoing_trust_ukprn` as well.

This work also adds a new UKPN validator that decouples the validation from the speciific attribute.

https://trello.com/c/ma1FCLSP

## Checklist

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -61,7 +61,7 @@ class ProjectsController < ApplicationController
   private def project_params
     params.require(:project).permit(
       :urn,
-      :trust_ukprn,
+      :incoming_trust_ukprn,
       :target_completion_date
     )
   end
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -6,7 +6,7 @@ class Project < ApplicationRecord
   accepts_nested_attributes_for :notes, reject_if: proc { |attributes| attributes[:body].blank? }
 
   validates :urn, presence: true, numericality: {only_integer: true}, length: {is: 6}
-  validates :trust_ukprn, presence: true, numericality: {only_integer: true}
+  validates :incoming_trust_ukprn, presence: true, numericality: {only_integer: true}
   validates :target_completion_date, presence: true
 
   validate :first_day_of_month, :trust_ukprn_is_correct_format
@@ -34,7 +34,7 @@ class Project < ApplicationRecord
   private def trust_exists
     retrieve_trust
   rescue AcademiesApi::Client::NotFoundError
-    errors.add(:trust_ukprn, :no_trust_found)
+    errors.add(:incoming_trust_ukprn, :no_trust_found)
   end
 
   private def retrieve_establishment
@@ -45,7 +45,7 @@ class Project < ApplicationRecord
   end
 
   private def retrieve_trust
-    result = AcademiesApi::Client.new.get_trust(trust_ukprn)
+    result = AcademiesApi::Client.new.get_trust(incoming_trust_ukprn)
     raise result.error if result.error.present?
 
     @trust = result.object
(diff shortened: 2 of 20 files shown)


app/controllers/projects_controller.rb
```diff
@@ -59,7 +59,7 @@ def update
   private def project_params
     params.require(:project).permit(
       :urn,
-      :trust_ukprn,
+      :incomming_trust_ukprn,
```
jkempster34, line 62: Had to double check, but incoming is with one "m" sorry 😬 . Hopefully just a find a replace?

app/models/project.rb
```diff
@@ -18,37 +19,29 @@ class Project < ApplicationRecord
   belongs_to :regional_delivery_officer, class_name: "User", optional: true
 
   def establishment
-    @establishment || retrieve_establishment
```
jkempster34, line 21: These memoizations will result in making the API call even if the instance variable has a value

I think something like this might work (not tested)

```ruby
def establishment
  @establishment ||= begin
    result = AcademiesApi::Client.new.get_establishment(urn)
    raise result.error if result.error.present?
    establishment = result.object
  end
end

File: app/validators/ukprn_validator.rb [hunk 3]
```diff
@@ -0,0 +1,7 @@
+class UkprnValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    unless value.to_s.match?("^1[0-9]{7}$")
```
jkempster34, line 3: 👍

spec/validators/ukprn_validator_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  subject do
+    Class.new {
+      include ActiveModel::Validations
+      attr_accessor :ukprn_value
+      validates :ukprn_value, ukprn: true
+    }.new
+  end
+
+  context "with a valid UKPRN" do
+    it "is invalid" do
```
jkempster34, line 13: Copy paste error, should be "is valid"

spec/validators/ukprn_validator_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      subject.ukprn_value = 12345678
+      expect(subject.valid?).to be true
+    end
+  end
+
+  context "with a UKPRN that is longer than eight digits" do
+    it "is invalid" do
+      subject.ukprn_value = 123456789101112
```
jkempster34, line 21: Should this test the smallest change that can be invalid `123456789`. (i.e., one number more than is allowed)

I can't find a link to back me up here, but I'm sure this is a thing because we can't test every possibility

spec/validators/ukprn_validator_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect(subject.valid?).to be false
+    end
+  end
+
+  context "with a UKPRN that is shorter than eight digits" do
+    it "is invalid" do
+      subject.ukprn_value = 1234
```
jkempster34, line 28: Same here `1234567`

spec/validators/ukprn_validator_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
+
+RSpec.describe UkprnValidator do
```
jkempster34, line 3: These tests are duplicated in `project_spec.rb` which can now be removed

jkempster34 (changes requested):
LGTM with a few changes. Hope all of that makes sense

Like the separate validator, never done that before 👍

app/controllers/projects_controller.rb
mec: No probs!

app/models/project.rb
mec: Yep, I think you are right! Thank you! 👍

spec/validators/ukprn_validator_spec.rb
mec: I am not sure, I don't mind as long as it is visibly not valid! Happy to change it. 👍

jkempster34 (approved):
This all looks great to me! 🚅

mec:
@jkempster34 I've taken this a little further, refining the caching of the establishment and trust and covering that behaviour with tests - could you have one last look for me.

jkempster34:
> @jkempster34 I've taken this a little further, refining the caching of the establishment and trust and covering that behaviour with tests - could you have one last look for me.

Looks great! 👍

PR #172, opened 2022-09-16
(496) User role assignment

jkempster34:
### Remove the edit and update actions for Project
Currently, we use the edit action of project to assign team leader and caseworker. We want to move these into their own actions.

This removes the current implementation, policy, and specs.

### Add scopes for retrieving all team leaders and RDOs
We want users to be able to select a team leader and regional delivery officer from a list of all users with that role.

This adds the additional scopes to the User model.

### Use GOV.UK fonts for Accessible Autocomplete suggestions
By default, the accessible autocomplete component does not use GOV.UK GDS Transport font for suggestions.

This uses a wildcard to set the GDS Transport font for all children of the `autocomplete_wrapper`

### Add actions to assign team leader
This adds the GET and POST actions for assigning the project's team leader.

### Add actions to assign regional delivery officer
This adds the GET and POST actions for assigning the project's regional delivery officer.

### Add actions to assign caseworker
This adds the GET and POST actions for assigning the project's caseworker.

### Add authorisation policy for assignment actions
Only users with the team leader role should be able to assign a project's roles.

This adds a Pundit policy for the assignments controller, and conditionally renders the "change" links on the project information view depending on that policy.

### Add a default scope ordering users by first name
When we retrieve a list of users, we want them to be in order of first_name.

Add a default scope.

## Checklist


jkempster34:
[Change how users are assigned to Projects](https://trello.com/c/SYVazvKM/496-change-how-users-are-assigned-to-projects)

spec/requests/project_information_controller_spec.rb
```diff
@@ -25,7 +25,31 @@
     end
 
     it "returns a successful response" do
-      expect(subject).to have_http_status :success
+      expect(perform_request).to have_http_status :success
+    end
+
+    describe "links to assign users to project roles" do
+      let(:document) { Nokogiri.parse(body) }
```
mec, line 32: I am not sure what I think about forcing this request spec to be a feature spec by loading Nokogiri? Would we get the same confidence by expecting the unique paths to the change assignment to be in the body or not, something like:

```ruby
def has_change_links?
  expect(body).to include project_assign_team_lead_path(project) 
  expect(body).to include project_assign_regional_delivery_officer_path(project)
  expect(body).to include project_assign_caseworker_path(project)
end
```

mec (changes requested):
Very nice bit of work! I think there are a couple of things you should think about changing.

app/models/user.rb
```diff
@@ -2,6 +2,8 @@ class User < ApplicationRecord
   has_many :projects, foreign_key: "caseworker"
   has_many :notes
 
+  scope :team_leaders, -> { where(team_leader: true) }
```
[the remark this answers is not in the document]
jkempster34: Added a default scope ordering by name.

app/models/user.rb
```diff
@@ -2,6 +2,8 @@ class User < ApplicationRecord
   has_many :projects, foreign_key: "caseworker"
   has_many :notes
 
+  default_scope { order(first_name: :asc) }
```
mec, line 5: Okay! `default_scope` is pretty contentious! 

I let the last one slide as I just couldn't come up with a scernario in which Tasks are ordered by anything other that the `order` attribute, but here I can imagine Users being sorted different ways throughout the application and so we should avoid `default_scope`, here is some reading:

https://andycroll.com/ruby/dont-use-default-scope/

mec (changes requested):
All good except the controversial `default_scope`! I left a comment and link - let me know your thoughts.

app/models/user.rb
jkempster34: Interesting 🤔 

Added a separate scope, and called it on the others

mec (approved):
🚢

PR #176, opened 2022-09-16
Refactor TaskListCreator to accept workflow root as parameter

jacksonj04:
## Changes

This change makes it easier to separate workflows into multiple files by clearing up some test behaviour, and allowing us to reduce the amount of mocking needed.

This also includes some alterations to the exact behaviour which is being tested in the create projects feature spec; previously it mocked the YAML which was being loaded at project creation time, it now uses the real definition. This does mean that if the definition changes we may have to modify the test, but it also means we're doing less modification of the system whilst under test.

## Checklist


app/controllers/projects_controller.rb
```diff
@@ -3,6 +3,8 @@ class ProjectsController < ApplicationController
   after_action :verify_authorized
   after_action :verify_policy_scoped, only: :index
 
+  DEFAULT_WORKFLOW_ROOT = Rails.root.join("app", "workflows", "lists", "conversion").freeze
```
jkempster34, line 6: Only question I have is whether this should live here or in the `TaskListCreator`.

I suppose you could do this, which would allow you to pass arguments in specs when needed, but also keeps the path in the `TaskListCreator`

```ruby
class TaskListCreator
  DEFAULT_WORKFLOW_ROOT = Rails.root.join("app", "workflows", "lists", "conversion").freeze

  def call(project, workflow_root: DEFAULT_WORKFLOW_ROOT)
     workflow = load_workflow_definition(workflow_root)
     create_task_list_from_workflow(project, workflow)
   end
```

jkempster34 (approved):
Looks great to me. Left one comment about where the `DEFAULT_WORKFLOW_ROOT` constant should live. 

Thought I'd leave it up to you!

PR #178, opened 2022-09-20
(536) Split workflow definition files into sections

jacksonj04:
## Changes

The existing workflow definition file is becoming quite large and unwieldy, which increases the likelihood of mistakes being made during editing, as well as making multiple concurrent edits harder due to merge conflicts.

This PR slices the workflow definition file into a single `lists/{name}/definition.yml`, and then multiple files with one representing each section (in the `lists/{name}/sections/{section}.yml` format).

## Checklist

## Future work

The schema files are updated accordingly to check the integrity of the definition files, but there is no new check that each complete workflow has integrity. This is not needed at this stage (since our real workflow is exercised by function tests), but it may be worth adding a test in future which explicitly checks the integrity of the workflows, making sure all referenced sections exist and so-on.

- [Add a test to make sure workflows have integrity](https://trello.com/c/142umgiH/568-workflow-integrity-test-script)


app/workflows/lists/conversion/definition.yml
```diff
@@ -1,477 +1,7 @@
...
-          - title: Send completed handback proforma to regional delivery officer
-
-      - title: Complete processing of conversion grant
-        actions:
-          - title: Receive completed grant certificate from trust
-          - title: Save completed grant certificate to SharePoint
-          - title:
-              Update the "support grant certificate assurance report
-              spreadsheet"
+  - kickoff
```
mec, line 2: Could this file be called `tasklist` to match up with the language everywhere else?

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Show the region on the project summary and school details.
 - Notes can be edited by the user who created them
 - Notes can be deleted by the user who created them
+- Added richer documentation for the workflow schemas
```
mec, line 24: Is there a way to move the docs into the `doc` directory - I feel like that is the consistent place for all documentaiton which this breaks?

mec (changes requested):
I am strugging a bit with the naming on some of this - list, definition. workflow, it might just be me.

I do feel like all documentation should be in the `doc` directory - can that be done?

mec (approved):
🚢

PR #179, opened 2022-09-20
(523) Titleise the trust name for display

jacksonj04:
## Changes

Previously trust names would display in ALL CAPS, since this is how they arrive from upstream APIs. They are now run through `titleize` before display, which converts them to title case.

## Checklist


mec (approved):
I think I expected the change to happen in the view layer, but once we switch, we should do it everywhere, so this is probably better. 🚢

PR #183, opened 2022-09-22
(573) SharePoint link

jkempster34:
### Add SharePoint link to Project
We know that users collate documents in a SharePoint folder as part of the prepare journey, and that users of this service will need to update/add to them.

Add a SharePoint link property to the Project model.

### Restrict project routes
We don't use all seven default routes, only create the ones we do use.

### Collect SharePoint link on Project creation
We want to collect the SharePoint link when the user creates a project.

### Add URL validator
We want to check the hostname and scheme of the SharePoint link.

Add a generic validator.

### Validate that the SharePoint link follows the expected format
We believe all project SharePoint links will start with `https://educationgovuk-my.sharepoint.com/` or `educationgovuk.sharepoint.com`.

Validate this on project creation.

This also updates the UrlValidator to handle multiple URLs

### Add a safe_link_to helper
Currently, it's possible for a user to provide a malicious SharePoint link such as `"javascript:alert('malicious')"`, which is not escaped by the Rails `link_to` helper.

This was being flagged by Brakeman as a potential XSS issue.

Add a `safe_link_to` method that sanitizes links.

### Display a body for the SharePoint link
The SharePoint links can be long and not-very-nice to read.

Instead, we want to show a body for the link.

This also moves the SharePoint link to the bottom of the Project Information list so that the advisory board rows are together.

## Screenshots
image

## Checklist

### DIFF
--- a/app/views/project_information/show/_information_list.erb
+++ b/app/views/project_information/show/_information_list.erb
@@ -36,6 +36,10 @@
       row.value { @project.advisory_board_date }
     end
     summary_list.row do |row|
+      row.key { t('project_information.show.project_details.rows.establishment_sharepoint_link') }
+      row.value { link_to(@project.establishment_sharepoint_link, @project.establishment_sharepoint_link, target: :_blank) }
+    end
+    summary_list.row do |row|
       row.key { t('project_information.show.project_details.rows.advisory_board_conditions') }
       row.value { simple_format(@project.advisory_board_conditions, class: "govuk-body") }
     end
--- a/app/views/project_information/show/_information_list.erb
+++ b/app/views/project_information/show/_information_list.erb
@@ -37,7 +37,7 @@
     end
     summary_list.row do |row|
       row.key { t('project_information.show.project_details.rows.establishment_sharepoint_link') }
-      row.value { link_to(@project.establishment_sharepoint_link, @project.establishment_sharepoint_link, target: :_blank) }
+      row.value { safe_link_to(@project.establishment_sharepoint_link, @project.establishment_sharepoint_link) }
     end
     summary_list.row do |row|
       row.key { t('project_information.show.project_details.rows.advisory_board_conditions') }
--- a/app/views/project_information/show/_information_list.erb
+++ b/app/views/project_information/show/_information_list.erb
@@ -36,12 +36,12 @@
       row.value { @project.advisory_board_date }
     end
     summary_list.row do |row|
-      row.key { t('project_information.show.project_details.rows.establishment_sharepoint_link') }
-      row.value { safe_link_to(@project.establishment_sharepoint_link, @project.establishment_sharepoint_link) }
-    end
-    summary_list.row do |row|
       row.key { t('project_information.show.project_details.rows.advisory_board_conditions') }
       row.value { simple_format(@project.advisory_board_conditions, class: "govuk-body") }
+    end
+    summary_list.row do |row|
+      row.key { t('project_information.show.project_details.rows.establishment_sharepoint_link') }
+      row.value { safe_link_to(t("project.summary.establishment_sharepoint_link.value"), @project.establishment_sharepoint_link) }
     end
   end %>
 </div>
--- a/app/views/shared/_project_summary.html.erb
+++ b/app/views/shared/_project_summary.html.erb
@@ -22,8 +22,8 @@
             row.value { @project.establishment.region_name }
           end
           summary_list.row do |row|
-            row.key { t("project.summary.sharepoint_link.title") }
-            row.value { link_to(@project.establishment_sharepoint_link, @project.establishment_sharepoint_link, target: :_blank) }
+            row.key { t("project.summary.establishment_sharepoint_link.title") }
+            row.value { safe_link_to(@project.establishment_sharepoint_link, @project.establishment_sharepoint_link) }
           end
         end %>
 
--- a/app/views/shared/_project_summary.html.erb
+++ b/app/views/shared/_project_summary.html.erb
@@ -21,6 +21,10 @@
             row.key { t("project.summary.region.title") }
             row.value { @project.establishment.region_name }
           end
+          summary_list.row do |row|
+            row.key { t("project.summary.sharepoint_link.title") }
+            row.value { link_to(@project.establishment_sharepoint_link, @project.establishment_sharepoint_link, target: :_blank) }
+          end
         end %>
 
   </div>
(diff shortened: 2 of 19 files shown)


jkempster34:
[Collect a link to the project SharePoint folder at handover](https://trello.com/c/7UBJY8TQ/573-collect-a-link-to-the-project-sharepoint-folder-at-handover)

app/views/project_information/show/_information_list.erb
```diff
@@ -35,6 +35,10 @@
       row.key { t('project_information.show.project_details.rows.advisory_board_date') }
       row.value { @project.advisory_board_date }
     end
+    summary_list.row do |row|
+      row.key { t('project_information.show.project_details.rows.sharepoint_link') }
+      row.value { link_to(@project.sharepoint_link, @project.sharepoint_link, target: :_blank) }
```
mec, line 40: How do you feel about adding `sanatize` to this? I think it would start to document how we are doing all we can to prevent an injection attack? I am no sure it actually covers everything, perhaps we can even limit the tags down to `href` and `target` and wrap that up in a method?

mec: Also, given how long, ugly and unhelpful a SharePoint url is, I think it might be better to show so words instead of the url, something like:

> Open the project folder on SharePoint in a new tab

app/views/shared/_project_summary.html.erb
```diff
@@ -21,6 +21,10 @@
             row.key { t("project.summary.region.title") }
             row.value { @project.establishment.region_name }
           end
+          summary_list.row do |row|
+            row.key { t("project.summary.sharepoint_link.title") }
+            row.value { link_to(@project.sharepoint_link, @project.sharepoint_link, target: :_blank) }
```
mec, line 26: Same comments any time we use this, maybe the presenter could make this simpler?

mec (changes requested):
Thanks for taking the time to make the code base better as you go! :heart:

I think we could improve the user experience of this a little and possibly how we approach the validation.

Let me know what you think?

jkempster34:
@mec This is ready for re-review when you are ready. Moving the call to `sanitize` into a helper silenced the Brakeman warning which is nice (don't know why it wasn't silenced when called directly in the view 🤷 )

db/migrate/20220921141636_add_sharepoint_link_to_project.rb
```diff
@@ -0,0 +1,5 @@
+class AddSharepointLinkToProject < ActiveRecord::Migration[7.0]
+  def change
+    add_column :projects, :sharepoint_link, :string
```
mec, line 3: I know this is a total pain, but I have seen lots of evidence that there is an 'establishment' sharepoint folder and a 'trust' sharepoint folder, how do you feel about updating this to `establishment_sharepoint_link` now as we are here?

config/locales/en.yml
```diff
@@ -85,6 +85,7 @@ en:
         title: Region
       sharepoint_link:
         title: SharePoint link
+        value: Open the project folder on SharePoint in a new tab
```
mec, line 88: This is somewhere else where I think we should be explicit, my understanding is that this is the establishment SharePoint folder.

config/locales/en.yml
```diff
@@ -102,6 +102,7 @@ en:
           regional_delivery_officer: Regional delivery officer
           unassigned: Not yet assigned
           advisory_board_date: Date of advisory board
+          sharepoint_link: SharePoint link
```
mec, line 108: And here, I think it would be great to be super clear this is the establishment SharePoint folder, could we check with @SteveMilnes I guess we say 'school' not 'establishment'?

mec (changes requested):
Thanks @jkempster34 - Sorry to be a pain, could we be explicit that this is the 'school SharePoint folder'.

mec:
@jkempster34 I've also just noticed that a proper share link from SharePoint has a different host to one that is copy and pasted out of the URL bar - I expect this will cause issues for users, but I think we find out?

url: https://educationgovuk-my.sharepoint.com/personal/meyric_rawlings_education_gov_uk/_layouts/15/onedrive.aspx?id=%2Fsites%2FServiceDeliveryDirectorate%2FShared%20Documents%2FA2C%20and%20Interventions%2FComplete%20conversions%2C%20transfers%20and%20changes&listurl=https%3A%2F%2Feducationgovuk%2Esharepoint%2Ecom%2Fsites%2FServiceDeliveryDirectorate%2FShared%20Documents&remoteItem=%7B%22mp%22%3A%7B%22webAbsoluteUrl%22%3A%22https%3A%2F%2Feducationgovuk%2Dmy%2Esharepoint%2Ecom%2Fpersonal%2Fmeyric%5Frawlings%5Feducation%5Fgov%5Fuk%22%2C%22listFullUrl%22%3A%22https%3A%2F%2Feducationgovuk%2Dmy%2Esharepoint%2Ecom%2Fpersonal%2Fmeyric%5Frawlings%5Feducation%5Fgov%5Fuk%2FDocuments%22%2C%22rootFolder%22%3A%22%2Fpersonal%2Fmeyric%5Frawlings%5Feducation%5Fgov%5Fuk%2FDocuments%2FComplete%20conversions%2C%20transfers%20and%20changes%22%7D%2C%22rsi%22%3A%7B%22listFullUrl%22%3A%22https%3A%2F%2Feducationgovuk%2Esharepoint%2Ecom%2Fsites%2FServiceDeliveryDirectorate%2FShared%20Documents%22%2C%22rootFolder%22%3A%22%2Fsites%2FServiceDeliveryDirectorate%2FShared%20Documents%2FA2C%20and%20Interventions%2FComplete%20conversions%2C%20transfers%20and%20changes%22%2C%22webAbsoluteUrl%22%3A%22https%3A%2F%2Feducationgovuk%2Esharepoint%2Ecom%2Fsites%2FServiceDeliveryDirectorate%22%7D%7

proper share link: https://educationgovuk.sharepoint.com/:f:/s/ServiceDeliveryDirectorate/EltrV1yBDlxDh1MMDH35AuYBfDAygCzvTuvDy1B3a_vq9Q?e=NOGcIG

app/views/project_information/show/_information_list.erb
```diff
@@ -35,6 +35,10 @@
       row.key { t('project_information.show.project_details.rows.advisory_board_date') }
       row.value { @project.advisory_board_date }
     end
+    summary_list.row do |row|
```
mec, line 38: The position of this really bugs me! :) could we move it to the bottom so the two advisory board items are together?

mec (approved):
Thanks for sticking with this one! 🚢

PR #187, opened 2022-09-23
Add new action type: Subtasks

jacksonj04:
## Changes

To support legal document check tasks, add a new type of action which supports multiple checkboxes

## Checklist


app/workflows/schemas/section.schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "uk.gov.education/rsd/cctc/section",
+  "$id": "/schemas/section",
```
mec, line 3: I am not sure how this simplification makes this better organised?

package.json
```diff
@@ -6,7 +6,7 @@
     "lint:js": "eslint .",
     "lint:js:fix": "yarn run lint:js --fix",
     "workflows:validate:tasklists": "ajv -s app/workflows/schemas/tasklist.schema.json -d app/workflows/lists/**/tasklist.yml --spec=draft2020",
-    "workflows:validate:sections": "ajv -s app/workflows/schemas/section.schema.json -d 'app/workflows/lists/**/sections/*.yml' --spec=draft2020",
+    "workflows:validate:sections": "ajv -s app/workflows/schemas/section.schema.json -d 'app/workflows/lists/**/sections/*.yml' --spec=draft2020 -r app/workflows/schemas/action.schema.json",
```
mec: Does `workflows:validate:sections` also validate actions now then? That seems odd?

mec (commented):
It might just be my Monday morning brain, but I am struggling with all this! 

I've never really felt the need for this schema stuff and it seems to just get more and more complex as we go on.

I appreciate the idea that this protects us from invalid workflows and documents how the workflows are expected to work, but I am still unconvinced we need this level of protection right now.

I still feel like the `conversion` workflow should be part of the application code and tested as such, if the workflow is broken in some way, so is our application - it is not like these are user submitted files that we gracefully handle when incorrect?

Could we have stop for a moment and just ask ourselves if we think the is the right thing, genreally I think this work seems right, I just want to be sure the commitment is right for this stage of the project - happy to talk this over as needed.

app/workflows/schemas/section.schema.json
jacksonj04: Added explanation into the commit message as to why this change is happening (tl;dr; these are relative IDs, not explicitly rooted, which makes referencing them from other files hard)

package.json
jacksonj04: `workflows:validate:sections` has always validated actions, because we list the actions in our workflow in the `section.yml` files. What this does is tell the schema validator "validate the files at `app/workflows/lists/**/sections/*.yml` against `app/workflows/schemas/section.schema.json`, and the referenced schema you need is at `app/workflows/schemas/action.schema.json`".

We could just as easily nest all our definitions in one file, but it's likely to get unwieldy and hard to grok what it's specifying.

PR #188, opened 2022-09-23
(542) Show full name and email on assignment pages

jkempster34:
### Show full name and email on assignment pages
We want to show the full name and email of users in the autocomplete on the assignment pages so users can be confident that they are selecting the right person.

image

## Checklist

### DIFF
--- a/app/helpers/assignments_helper.rb
+++ b/app/helpers/assignments_helper.rb
@@ -1,0 +1,5 @@
+module AssignmentsHelper
+  def full_name_and_email(user)
+    "#{user.full_name} (#{user.email})"
+  end
+end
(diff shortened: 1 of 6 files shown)


jkempster34:
[Update the user assignment autocomplete](https://trello.com/c/8z4PJ0vu/542-update-the-user-assignment-autocomplete)

app/helpers/assignments_helper.rb
```diff
@@ -0,0 +1,5 @@
+module AssignmentsHelper
+  def full_name_and_email(user)
+    "#{user.full_name} (#{user.email})"
```
jacksonj04, line 3: How would we feel about plopping this in the model? It feels more like a representation of the data in the object to me than a manipulation of it, but weak opinion weakly held there.

jacksonj04 (approved):
One smol non-blocking comment, otherwise :shipit:

app/helpers/assignments_helper.rb
jkempster34: Hmmm, my thought process was that this was presentation logic relating only to how the user is represented in the view 🤔 but I see your point!

I'm just going to leave it for now, so we can get it merged

PR #190, opened 2022-09-26
Fixes and improvements for schema testing

jacksonj04:
## Changes

- An incorrect path in the file glob meant that sections weren't being correctly validated against the schema. They now are.
- Schema documentation wasn't being generated into a clean folder, so files which were no longer relevant could still be kicking around. It now nukes the folder first.


jkempster34 (approved):
Nice 👍

PR #192, opened 2022-09-27
(582) Required first name and last name

jkempster34:
### Update the `users:create` Task to accept name attributes
Users now have a `first_name` and `last_name`. 

This updates the `users:create` Task, so that they can be passed as Task arguments, and updates the relevant documentation.

### Validate presence of name attributes on User
Currently, the `first_name` and `last_name` properties of the User model are optional.

This adds validation to check that all users do have `first_name` and `last_name`. 

There is no need for any data migrations as we don't yet have a deployed application.

### Remove fallbacks for when the name attributes of User are absent
We have various bit of logic dotted around to handle users that don’t have `first_name` and `last_name`.

Now that these are required, we can safely remove any fallbacks.

## Checklist

### DIFF
--- a/spec/requests/sign_in_spec.rb
+++ b/spec/requests/sign_in_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.describe "Sign in" do
-  let(:user) { User.create!(email: "user@education.gov.uk") }
+  let(:user) { create(:user) }
 
   context "when the user is signed out" do
     it "redirects them to the sign in page and shows a helpful message" do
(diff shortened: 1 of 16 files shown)


jkempster34:
[Require Users to have `first_name` and `last_name`](https://trello.com/c/zIWkb4mR/582-require-users-to-have-firstname-and-lastname)

spec/requests/sign_in_spec.rb
```diff
@@ -27,10 +27,7 @@
   end
 
   context "when the users is not known by the application" do
-    before do
-      User.create!(email: "another.user@education.gov.uk")
```
jacksonj04, line 31: This is a change in the behaviour of the test; what was the significance (or given the fact that deleting it still passes, what should the significance have been) of `another.user@education.gov.uk`?

jacksonj04 (approved):
I love PRs which delete more than they add.

One question about a changing test behaviour, but other than that this looks good and wholesome to me.

spec/requests/sign_in_spec.rb
jkempster34: Ah, yeah, I forgot about that. That line doesn't seem to do anything. 

I'll double check what the test is trying to assert and rewrite if necessary 👀

PR #193, opened 2022-09-28
(592) Add "subheading" action type

jacksonj04:
## Changes

This introduces a new type of action, `subheading`, which can be used to place headings in a task's list of actions (ie to differentiate subsections within a task).

## Screenshots

localhost_3000_projects_3F805EC1-E0ED-4A6A-A927-6556F87CD2F4_tasks_3DE5C874-1391-4CD6-9A65-7BA87851AE70

## Checklist


spec/factories/action_factory.rb
```diff
@@ -1,5 +1,6 @@
 FactoryBot.define do
```
mec, line 1: I would love to see this when this property was introduced, but not a deal breaker.

mec (approved):
Smashed it! Two comments that you can take or leave, otherwise I love how simple this has become! 🚢

PR #196, opened 2022-09-30
(589) Task level notes

jkempster34:
### Add optional relationship between Task and Note
We want to add task-level notes.

We are going to use the existing Note model with an optional relationship to a Task.

### Add new note from task page
This allows users to add a new note from the task page which is associated to the task.

For now, the only difference is that the note has a task ID. The following commits will update any behaviours.

### Add method to distinguish task-level note
Add a helper method so that we can easily find out if a note is a task level note.

### Users get redirected to the task page after creating a task level note
Users should be redirected back to the task page where they started adding their task level note.

Update the redirect in the controller so that it is variable backed on the note type.

### Only show project level notes on the notes page
Currently, we show all notes on the notes page. Now that we have task level notes, we want to only show project level notes on the notes page.

Add a scope for project level notes, and surface these only.

### Show task level notes
We want to display the task level notes on the page of the task itself.

### Users can edit task level notes
Add a link to edit notes onto the task page. Update the redirect destination to match note creation.

### Users can delete task level notes
Users should be redirected to the task page when deleting a task level note, as they are when creating or updating it.

### Make sure the back links work for notes and task level notes
We want to make sure that the "Back" links take the user to the correct place:
- The notes index page when returning from editing or creating a project level
  note.
- The individual task page when returning from editing or creating a task level
  note.

Add a helper so that we can change the destination based of the type of note.

## Checklist

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -19,7 +19,6 @@ class NotesController < ApplicationController
     if @note.valid?
       @note.save
 
-      redirect_path = @note.task_level_note? ? project_task_path(@project, @note.task.id) : project_notes_path(@project)
       redirect_to(redirect_path, notice: I18n.t("note.create.success"))
     else
       render :new
@@ -39,7 +38,7 @@ class NotesController < ApplicationController
 
     if @note.valid?
       @note.save
-      redirect_to project_notes_path(@project), notice: I18n.t("note.update.success")
+      redirect_to redirect_path, notice: I18n.t("note.update.success")
     else
       render :edit
     end
@@ -57,6 +56,10 @@ class NotesController < ApplicationController
   def confirm_destroy
     @note = Note.find(params[:note_id])
     authorize @note
+  end
+
+  private def redirect_path
+    @note.task_level_note? ? project_task_path(@project, @note.task.id) : project_notes_path(@project)
   end
 
   private def find_project
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -1,20 +1,17 @@
 class NotesController < ApplicationController
   before_action :find_project
   before_action :find_project_level_notes, only: :index
-  after_action :verify_authorized
-
-  def index
-    authorize Note
-  end
+  after_action :verify_authorized, except: :index
+  after_action :verify_policy_scoped, only: :index
 
   def new
-    authorize Note
     @note = Note.new(project: @project, user_id:, task_id: params[:task_id])
+    authorize @note
   end
 
   def create
-    authorize Note
     @note = Note.new(project: @project, user_id:, **note_params)
+    authorize @note
 
     if @note.valid?
       @note.save
@@ -67,7 +64,7 @@ class NotesController < ApplicationController
   end
 
   private def find_project_level_notes
-    @notes = Note.includes([:user]).where(project: @project).project_level_notes
+    @notes = policy_scope(Note).includes([:user]).project_level_notes(@project)
   end
 
   private def note_params
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -19,7 +19,8 @@ class NotesController < ApplicationController
     if @note.valid?
       @note.save
 
-      redirect_to project_notes_path(@project), notice: I18n.t("note.create.success")
+      redirect_path = @note.task_level_note? ? project_task_path(@project, @note.task.id) : project_notes_path(@project)
+      redirect_to(redirect_path, notice: I18n.t("note.create.success"))
     else
       render :new
     end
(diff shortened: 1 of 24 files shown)


jkempster34:
[Add task level notes](https://trello.com/c/ffD3LAni/589-add-task-level-notes)

app/views/tasks/shared/_task_notes.html.erb
```diff
@@ -0,0 +1,5 @@
+<aside>
+  <h2 class="govuk-heading-m"><%= t("note.show.task_notes.title") %></h2>
+
+  <%= govuk_button_link_to "Add note", new_project_note_path(task_id: @task.id), secondary: true %>
```
jacksonj04: The `"Add note"` here could probably come from the language file.

app/controllers/notes_controller.rb
```diff
@@ -59,6 +58,10 @@ def confirm_destroy
     authorize @note
   end
 
+  private def redirect_path
```
jacksonj04, line 61: I'd be tempted to make this a method of the model itself, eg `@note.path_to_note`, but that's maybe just personal preference on the thick model / thick controller scale.

app/views/tasks/shared/_task_notes.html.erb
```diff
@@ -13,6 +13,10 @@
         <h3 class="govuk-heading-s govuk-!-padding-top-0"><%= note.user.full_name %></h3>
         <p class="govuk-body"><%= render_markdown(note.body) %></p>
 
+        <% if policy(note).edit? %>
+          <%= govuk_link_to "Change", edit_project_note_path(@task.project, note) %>
```
jacksonj04, line 17: Another string which could live in the language file.

spec/requests/notes_controller_spec.rb
```diff
@@ -84,6 +89,21 @@
...
         expect(Note.count).to be 1
         expect(Note.last.body).to eq(new_note_body)
+        expect(Note.last.task).to be_nil
+      end
+
+      context "when the note is a task level note" do
+        let(:task) { create(:task) }
+        let(:params) { {note: {body: new_note_body, task_id: task.id}} }
+
+        it "saves the note and redirects to the index view with a success message" do
```
jacksonj04, line 99:
```suggestion
        it "saves the note and redirects to the task view with a success message" do
```

spec/requests/notes_controller_spec.rb
```diff
@@ -183,6 +204,20 @@
         expect(Note.count).to be 1
         expect(Note.last.body).to eq(new_note_body)
       end
+
+      context "when the note is a task level note" do
+        let(:task) { create(:task) }
+        let(:params) { {note: {body: new_note_body, task_id: task.id}} }
+
+        it "saves the note and redirects to the index view with a success message" do
```
jacksonj04, line 212:
```suggestion
        it "saves the note and redirects to the task view with a success message" do
```

spec/requests/notes_controller_spec.rb
```diff
@@ -215,6 +250,17 @@
 
       expect(Note.where(id: note_id)).to_not exist
     end
+
+    context "when the note is a task level note" do
+      let(:note) { create(:note, :task_level_note, user: user) }
+
+      it "deletes the note and redirects to the index view with a success message" do
```
jacksonj04, line 257:
```suggestion
      it "deletes the note and redirects to the task view with a success message" do
```

app/controllers/notes_controller.rb
```diff
@@ -19,7 +19,8 @@ def create
     if @note.valid?
       @note.save
 
-      redirect_to project_notes_path(@project), notice: I18n.t("note.create.success")
+      redirect_path = @note.task_level_note? ? project_task_path(@project, @note.task.id) : project_notes_path(@project)
```
jacksonj04: I wonder if this should redirect to a page fragment (ie `/task/123abc#task-def456`) so people aren't stuck with scrolling? Perhaps could also make use of [`:target` CSS](http://jsfiddle.net/davidThomas/NGyVL/) to highlight, but that's a bit of a refinement which may need testing.

jacksonj04 (approved):
Nothing substantiative to report. A couple of copypaste errors in test titles, couple of language file suggestions, one non-critical quality of life thing (maybe), but nothing blocking.

app/views/tasks/shared/_task_notes.html.erb
mec: I've also made it 'Edit note'

app/controllers/notes_controller.rb
mec: This is a great idea, but I wonder if it is a future improvement when we know a little more about the volume of notes that get added to tasks?

app/controllers/notes_controller.rb
mec: Yeah, interesting! I think there are lots of ways to do this that possibly come down to preference, as this is Joseph's work I think I will leave this as is, it works for me.

jacksonj04 (approved):
🚀

PR #199, opened 2022-09-30
Use mcr.microsoft.com/azure-sql-edge image instead of mssql

Stretch96:
## Changes

* The mssql image isn't compatible with M1 Macs (ARM infrastructure)
* We're currently using Azure sql, so this may be preferable

## Checklist

### DIFF
--- a/.env.database.example
+++ b/.env.database.example
@@ -3,4 +3,4 @@
 #
 # SA_PASSWORD must be at least 8 characters including uppercase, lowercase letters, base-10 digits and/or non-alphanumeric symbols
 ACCEPT_EULA=Y
-SA_PASSWORD=
+MSSQL_SA_PASSWORD=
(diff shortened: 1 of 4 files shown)


mec:
@Stretch96 couple of things to say about this:

- are we sure the azure-sql-edge image is what we think it is? I have an M1 Mac and had to use it, but what worried me is that it could also be this product https://learn.microsoft.com/en-us/azure/azure-sql-edge/overview which only has a subset of features of MSSQL server: https://learn.microsoft.com/en-us/azure/azure-sql-edge/features?source=recommendations

I am no MSSQL server expert and none of the unsupported features look like anything we would need to use, but I do wonder if there is a way to have more confidence (and that might come via you!!)?

- the tests keep intermitently failing like they are here, we feel like this could be because the MSSQL server takes a long time to actually become available - we were planning to ask you for advice about this.

mec:
@Stretch96 I am going to incorporate this into another piece of work that includes an ADR on why we have to switch image and what it means, I'll close this and open a new PR.

PR #201, opened 2022-10-03
Run prettier

jacksonj04:
## Changes

This resolves an uncaught yaml formatting issue.


mec:
Do we know why this is not being caught in CI?

mec (approved):

PR #217, opened 2022-10-05
(643) Separate linting/formatting and application tests under CI

jacksonj04:
## Changes

At the moment, our test suite under CI doesn't run linting and formatting checks on all files. This is because all our tests run in the Docker container, which only includes files we explicitly copy in to the image.

To solve this, this PR separates the linting/formatting/schema tests and the application tests into two distinct processes. The linting can run directly in a GitHub worker (which completes a lot faster, giving us a fail-fast error for the more mundane things like style issues), and then the application tests run in the full-blown Docker container for parity with production. It also gives us more immediate visibility over which part of the test suite has failed.

## Checklist

## Next steps

- [x] Change the GitHub required tests configuration

### DIFF
--- a/.github/workflows/continuous-integration-tests.yml
+++ b/.github/workflows/continuous-integration-tests.yml
@@ -8,12 +8,32 @@ on:
       - develop
 
 jobs:
-  rails-test:
-    name: Rails test
+  linting-and-formatting:
+    name: Linting and formatting tests
     runs-on: ubuntu-latest
 
     env:
       RAILS_ENV: test
+      ONLY_LINTING: "true"
+      BUNDLE_ONLY: "linting"
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - uses: ruby/setup-ruby@v1
+      - uses: actions/setup-node@v3
+
+      - name: Test
+        run: script/no-docker/test
+
+  application:
+    name: Application tests
+    runs-on: ubuntu-latest
+
+    env:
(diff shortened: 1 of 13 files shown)


mec:
I like where this is going, I struggle a bit with the bash scripts! 

This is going to be a bit rambly!

I think I would like to understand what we want from this? 

If we want speed, perhaps we run the Ruby linters in the Docker image and the non-ruby ones in a worker - then we don't have to bundle both times which seems to be the longest part?

If the concern is we keep missing files because they don't get copied into the image, perhaps we just make sure they are copied?

For me the mixing of `if [ -z "$VAR" ]` and `if [ -n "$VAR" ]` in this context makes things hard to parse, which isn't in your work, but now might be a good time to sort all this? I also think I would prefer two blocks rather than all the if statements - might just be a me thing though?

I don't think I understand the auto fix option, if it just fixed the errors but never commits then - what is the point? They are fixed in the image but I run the linters locally and see errors - that feels frustrating to me?

It looks like this [run](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/actions/runs/3190366091/jobs/5216020647#step:6:349) was in auto fix, but I cannot see how or why  - this has bitten us before where CI passes because it fixes but locally it fails?!?

Happy to have a chat about this or let me know your thoughts here?

jkempster34 (approved):
I think this all looks reasonable to me 👍 Thanks for fixing!

PR #218, opened 2022-10-05
Fix incorrect release comparison links

jacksonj04:
The links in the CHANGELOG to compare two versions were to invalid destinations. They now compare releases correctly.


mec (approved):
Thanks for sorting this - been on my list to look at!

PR #222, opened 2022-10-07
Changes to legal docs sections

titlescreen:
- Clear legal documents becomes Clear and sign legal documents
- Add sign and seal actions to Supplemental funding agreement

## Changes

## Checklist

image

localhost_3000_projects_9BAA4077-3CC7-4B34-8E1A-8027F95803A5_tasks_1FA679EF-5746-4209-86C0-792695217D82

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Supplemental funding agreement now has sign and seal actions.
+- Clear legal documents becomes Clear and sign legal documents.
 - Only project level notes are shown on the notes page.
 - The back link destination is the task page for task level notes, or the notes
   page for project level notes.
--- a/app/workflows/lists/conversion/sections/clear-legal-docs.yml
+++ b/app/workflows/lists/conversion/sections/clear-legal-docs.yml
@@ -1,4 +1,4 @@
-title: Clear legal documents
+title: Clear and sign legal documents
 tasks:
   - title: Land questionnaire
     hint: |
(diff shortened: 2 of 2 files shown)


jacksonj04 (changes requested):
I'm happy with these content changes, *but* we've got an unexpected visual change due to how we apply the reduced padding for clear legal documents, and as such I've given this a "request changes" until we figure out the way around it.

### Before

localhost_3000_projects_3F805EC1-E0ED-4A6A-A927-6556F87CD2F4_tasks_3DE5C874-1391-4CD6-9A65-7BA87851AE70

### After

localhost_3000_projects_41E71D1D-4F8E-496C-9E3C-546947AA0FA3_tasks_D72AB6CD-3BE3-47DF-9F08-9A306B902E2B

jacksonj04:
Ways to fix:

- We just update the code so it looks for `Clear and sign legal documents` instead of `Clear legal documents`
- We add a new `reduce_padding: true` parameter to the _task_ which does this layout change instead, which makes it a function of the data rather than the text (and thus more robust). It also means we can apply this reduced padding to a task in any section, rather than every task in the legal documents section.
- We add a new `reduce_padding: true`, `padding: reduced` or similar parameter to the _action_ which applies different padding on an action-by-action basis. This involves adding more configuration to the workflow files, but does mean we can do this reduction of padding more selectively.

My personal preference is the last option and being more explicit action-by-action, but that's because I'm a big fan of making sure things are generalised.

jacksonj04 (approved):

titlescreen:
After a chat we have all decided to do this

We add a new reduce_padding: true, padding: reduced or similar parameter to the action which applies different padding on an action-by-action basis. This involves adding more configuration to the workflow files, but does mean we can do this reduction of padding more selectively.

@SteveMilnes and I will make the other content pages, then the parameter can be added in retroactively.

PR #233, opened 2022-10-10
(565) Projects pagination

jkempster34:
### Add Pagy for pagination
We want to implement pagination using the GOV.UK components library, which is
compatible with Pagy.

This adds and configures Pagy. The default number of items per page is set to
10, and the default size parameter is set to match the Design System
documentation.

### Paginate the projects index page
For now, we send a request to the API for each project on the project index
page. These requests are very slow, resulting in a very, very slow page load time
when there are multiple projects. We want to paginate the projects in order to
reduce the load.

This paginates the projects index page. It also moves the `Add a new project`
button to the top of the page, because it was looking a bit odd next to the
pagination component.

### Fix the total number of projects shown in the heading
We need to update the way we calculate the total number of projects now that we
paginate them (else we are just showing the number of projects per page).

This uses the Pagy `count` method which returns the total number of records.

## Screenshot

image

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Users can add, edit, and delete task level notes. These are displayed on the
   task page.
 - Added a privacy notice and link to the notice in the footer
+- Pagination has been added to the projects index page. 10 projects are shown
+  per page.
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


jkempster34:
[Paginate the project list](https://trello.com/c/v1rrjm0v/565-paginate-the-project-list)

jacksonj04 (approved):
Love me some pagination. Ship it!

PR #235, opened 2022-10-10
Update containerapp after deploying image to ACR (development)

Stretch96:
## Changes

* Once the image has been deployed to ACR, we need to tell ACA to update the app.
* It uses the sha as the image tag, so that it will always change, even if the release version is set to a branch name (eg. main)

## Checklist


mec (approved):
🚢

PR #237, opened 2022-10-10
(673) Chore: support for arm macs

mec:
## Changes

The official MSSQL container image is not built with ARM processor support and so cannot be used on newer Macs.

The only work around at the moment is to use the 'Azure SQL Edge` image - whilst this is MSSQL server, it is a different product with a different feature set.

Using this image let's up support new Macs but offers less parity with our live environments.  For the moment, we will accept this. There is an ADR to record this decision in this work.

Whilst here, I removed the script support for running a 100% Docker local development enviroment, no one on the team uses this method and the scripts and compose files are very out of date and not debt we should carry - if we want to add local Docker support later, we can.

https://trello.com/c/xrufiJem

## Checklist


.env.database.example
```diff
@@ -3,4 +3,4 @@
 #
 # SA_PASSWORD must be at least 8 characters including uppercase, lowercase letters, base-10 digits and/or non-alphanumeric symbols
```
jacksonj04, line 4:
```suggestion
# MSSQL_SA_PASSWORD must be at least 8 characters including uppercase, lowercase letters, base-10 digits and/or non-alphanumeric symbols
```

doc/decisions/0013-use-the-microsoft-azure-sql-edge-container-image.md
```diff
@@ -0,0 +1,38 @@
...
+different product offering container image, 'Azure SQL Edge' - this image
+contains a running MSSQL server amoungst other things, and can be used as a
+replacement.
+
+Azure SQL Edge does not have complete feature parity with MSSQL server [2], we
+believe it has enough features for our use case.
+
+## Decision
+
+We will switch all out Containers to use the Azure SQL Edge image.
```
jacksonj04, line 27:
```suggestion
We will switch all our Containers to use the Azure SQL Edge image.
```

jacksonj04 (approved):
Couple of tiny doc/comment tweaks, but this looks to run on my M1.

PR #238, opened 2022-10-10
(658) Make padding a configurable parameter for actions

jacksonj04:
TRELLO-Kt3hVR9A

## Changes

A series of changes which remove the unique behaviours for tasks within the "clear legal documents" section, and instead replaces them with the ability to adjust padding on each action individually.

Each action now by default has a `normal` padding between it and the action above, but this can be explicitly adjusted action-by-action by using the new `padding` parameter and setting it to `reduced`:

```yaml
- type: subheading
  title: Signed by the Secretary of State
- type: single-checkbox
  padding: reduced
  title: Sent to Academies Operational Practice Unit
```

This gives content designers more granular control over the rendering of each task, as well as providing a pattern to follow for adding new parameters to actions in future.

#270 contains related changes which narrow the scope of each schema file, to make changes to the schema more contained and easier to understand.

## Checklist


jacksonj04:
https://trello.com/c/Kt3hVR9A/658-add-optional-padding-parameter-to-tasklist

app/assets/stylesheets/application.scss
```diff
@@ -21,20 +21,25 @@ $govuk-global-styles: true;
...
-.task-checkboxes,
-.clear-legal-documents-task-checkboxes {
   .govuk-checkboxes__item {
     label {
       @extend .govuk-\!-font-weight-bold;
     }
   }
 }
 
+.task-action__wrapper:first-of-type {
```
jkempster34, line 31: Is ti worth scoping these within the `task-checkboxes` class (or other similar container class) to keep things organised?

app/controllers/tasks_controller.rb
```diff
@@ -3,7 +3,6 @@ class TasksController < ApplicationController
   before_action :find_task_level_notes, only: :show
 
   def show
-    render "tasks/clear_legal_documents/show" if @task.clear_legal_documents_type?
```
jkempster34, line 6: Nice!

app/models/task.rb
```diff
@@ -17,10 +17,6 @@ def completed_actions_count
     @completed_actions_count ||= actions.where(action_type: "single-checkbox").where(completed: true).count
   end
 
-  def clear_legal_documents_type?
-    section.title == LEGAL_DOCUMENT_SECTION_TITLE
```
jkempster34, line 21: There is a `LEGAL_DOCUMENT_SECTION_TITLE` constant at the top of this file that can go in the bin too

app/assets/stylesheets/application.scss
```diff
@@ -21,20 +21,25 @@ $govuk-global-styles: true;
...
+  margin-top: 0;
+}
+
+.task-action__wrapper__padding-normal {
+  margin-top: govuk-spacing(9);
```
jkempster34, line 36: I think this padding might be too much! Looking at the prototype, `4` looks closer

app/assets/stylesheets/application.scss
```diff
@@ -21,20 +21,25 @@ $govuk-global-styles: true;
...
+}
+
+.task-action__wrapper__padding-reduced {
```
jkempster34, line 39: You can do some Sass magic here:

```css
.task-action__wrapper {
  &__padding-normal {
    margin-top: govuk-spacing(9);
  }
  
  &__padding-normal {
    margin-top: govuk-spacing(2);
  }
}
```

app/assets/stylesheets/application.scss
jkempster34: Also I think the reduced padding should be 0, but you would have to set `.govuk-form-group: margin-bottom: 10px` as each checkbox is now its own group.

jkempster34 (changes requested):
Ran this and it looks great! Left some comments but nothing major 😎

app/assets/stylesheets/application.scss
jacksonj04: Ooh, I thought that only worked for things like identifiers and selectors rather than straight string concatenation, TIL.

jkempster34 (approved):
Looks great! 

I still think the `margin-top: govuk-spacing(9);` is too large and should be `margin-top: govuk-spacing(4);`, but I'm not sure if you have had conversations to the contrary!

jacksonj04:
@jkempster34 I've not had any conversations to the contrary, but I also think `9` is a match for what was there before and the prototype:

Screenshot 2022-10-25 at 11 42 05

(Mine on the left, prototype on the right)

There's a bit of extra space between form items even in `reduced` state due to each checkbox now getting its own margin from the form group like you mentioned, but I think the space between `normal` elements is correct. Are we talking about a different margin?

jkempster34:
> @jkempster34 I've not had any conversations to the contrary, but I also think `9` is a match for what was there before and the prototype:
> 
> Screenshot 2022-10-25 at 11 42 05
> 
> (Mine on the left, prototype on the right)
> 
> There's a bit of extra space between form items even in `reduced` state due to each checkbox now getting its own margin from the form group like you mentioned, but I think the space between `normal` elements is correct. Are we talking about a different margin?

Is the `9` not part of the non-legal documents sections? One min I'll get a comparison

jkempster34:
> > @jkempster34 I've not had any conversations to the contrary, but I also think `9` is a match for what was there before and the prototype:
> > Screenshot 2022-10-25 at 11 42 05
> > (Mine on the left, prototype on the right)
> > There's a bit of extra space between form items even in `reduced` state due to each checkbox now getting its own margin from the form group like you mentioned, but I think the space between `normal` elements is correct. Are we talking about a different margin?
> 
> Is the `9` not part of the non-legal documents sections? One min I'll get a comparison

@jacksonj04 

App:

image

Prototype:

image

PR #245, opened 2022-10-12
(687) Remove duplicate service name

mec:
## Changes

Previously we showed the service name on the Project index.

As the service name is in the header, this was duplication.

We also used an h1 for two elements, which I don't believe is good practice.

## Screen shot

image

## Checklist


jkempster34 (approved):
Great!

PR #249, opened 2022-10-14
Fix: missing not applicable options

mec:
## Changes

Our "legal documents' section renders a different view which includes the ability to show the 'not applicable' option.

We really don't like this and have work planned to change this very (very) soon, however, we use the section title to identify when to render a legal document task over a generic task.

Recently we changed the title of this section to:

Clear and sign legal documents

and did not update the condition.

Here we do that, even though it is horrific.


jacksonj04 (approved):
Aaaaaah!

PR #250, opened 2022-10-17
(675) Actions now submit state individually

jacksonj04:
## Changes

Previously, the state of all actions was submitted as a single fieldset with an array of checked values. This commit makes a subtle change to the behaviour so that each action is submitted as an individual field (assuming it is checked).

This means that we can better isolate the behaviour of each action in the form. Initially this is to support improvements to form layout, but in future this will also make it easier for some actions to have non-single-checkbox behaviours, such as text inputs, radios or multiple choice.

The `task_params` method is removed to simplify the flow of the form logic, since we are no longer attempting to mass assign values.

## Checklist


app/controllers/tasks_controller.rb
```diff
@@ -8,7 +8,7 @@ def show
 
   def update
     @task.actions.each do |action|
-      action_completed = action.id.in?(task_params[:completed_action_ids] || [])
+      action_completed = action.id.in?(params.dig("task", "actions") || [])
```
jacksonj04, line 11: This uses `params.dig` over `params[:task]` so that it neatly returns `nil` on failure rather than throwing an exception.

mec: I think we can take this refactor a little further, at the moment this is a bit of an odd mix, we are throwing arrays about here when I don't think we need to.

If we want to keep this `each` then `action.update(completed: action_completed)` might read better?

A further refinement might be something like:

`Action.update([array, of, all, task, ids], completed: [array, of, the, response, values])` 

to do everything in one DB call, which I think would be my preference.

We can also get the form builder to return both checked and uncheked states, which I think will make this even better as the unset boxes will also be in the params:

https://govuk-form-builder.netlify.app/form-elements/checkboxes/#generating-a-single-checkbox

mec (changes requested):

app/controllers/tasks_controller.rb
jacksonj04: If we use the unchecked state as well, we can actually get the update down to an even cleaner one-line, since we can just use the value of the checkbox directly from the form and it'll be cast to a boolean (so `1`/`0` work perfectly).

I'm less certain about the single DB call, since that only works as long as everything is a checkbox so we're going to end up taking it back out again in future. I have, however, fixed that if a task is marked N/A it would be updating every checkbox with a DB hit and then unsetting them all again!

mec (approved):
This is much better, thank you! I think we could go further still, but that is probably turning in to gold plating!

I can't see why bullet is making you include Task either, it can raise false positives, I am not worried about how we have this though.

PR #259, opened 2022-10-19
Add human-readable identifier slugs to tasks

jacksonj04:
TRELLO-eiQP2Dbt

## Changes

Every task in a workflow now requires a `slug` parameter with a human-readable identifier, restricted to URL-friendly characters. These are added to the database during task-list creation, but are not yet used elsewhere in the codebase.

NB: The uniqueness of these identifiers is _not_ enforced at this point, either by the schema or the database.

## Checklist

### DIFF
--- a/app/services/task_list_creator.rb
+++ b/app/services/task_list_creator.rb
@@ -22,6 +22,7 @@ class TaskListCreator
   private def create_tasks_in_section(section_data, section)
     section_data.fetch("tasks").each_with_index do |task_data, index|
       task = Task.create(
+        slug: task_data.fetch("slug"),
         title: task_data.fetch("title"),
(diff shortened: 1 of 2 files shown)


jkempster34 (approved):
This all looks good to me, good idea adding the Regex for the slug!

Is there any consequence to having previous project slugs as `task.title.parameterize` and new slugs as the ones you have defined?

jacksonj04:
There _might_ be a future consequence, but we're going to have to address that by necessity when we start trying to build these identifiers into application logic. Since they could potentially change over time as the process changes, we either need to handle forward data migration for each project (which starts to introduce the horror of workflow versioning), or build the logic in such a way that it too is stashed along with the rest of the things needed to render a project.

PR #267, opened 2022-10-20
(686) Styled error pages

jkempster34:
### Add High Voltage for static pages
High Voltage simplifies how we add static pages.

Add the gem and configure to:
- Use "top-level routes", so that the `pages` directory will not be included.
  E.g., `http://www.example.com/privacy`, instead of
  `http://www.example.com/pages/privacy`.
- Override the PagesController so that we can have more control. E.g., applying
  authorisation if we want to.

### Add page not found error page
Add a `page not found` error page following the suggested GOV.UK Design System
pattern.

### Add internal server error page
Add an `internal server error` page following the suggested GOV.UK Design
System pattern.

## Screenshots
image

image

## Checklist


config/routes.rb
```diff
@@ -7,6 +7,10 @@
   # Static pages
   get "/privacy", to: "static_pages#privacy"
 
+  # High voltage configuration. Matches routes from the root of the domain. Uses
```
jacksonj04: Would this be better at the bottom of the stack so we don't evaluate every request for another page against this?

spec/features/users_can_view_static_pages_spec.rb
```diff
@@ -1,9 +1,10 @@
 require "rails_helper"
 
-RSpec.feature "Users can view a list of projects" do
```
jacksonj04, line 3: Nice catch

app/views/pages/internal_server_error.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">Sorry, there is a problem with the service</h1>
+    <p class="govuk-body">Try again later.</p>
+
+    <p class="govuk-body">
+      <a href="mailto:regionalservices.rg@education.gov.uk" class="govuk-link">Contact the Regional Services support team</a> if you have any questions.
```
jacksonj04, line 7: Given that we've used this email address in a few places, is it worth making it either an env variable or baking it into a constant somewhere? I'm not sure the language file is the right place, but it definitely feels more like something we should define and reuse than just type.

jacksonj04 (approved):

PR #269, opened 2022-10-21
Chore: Set the host name in production

mec:
## Changes
We want to prevent DNS rebinding and other Host header attacks so we set a allow list of host names, which for now is only one.

We'll get the host name from the environment the deployment is in.

https://guides.rubyonrails.org/configuring.html#actiondispatch-hostauthorization

## Checklist

### DIFF
--- a/doc/environment-variables.md
+++ b/doc/environment-variables.md
@@ -1,8 +1,25 @@
 # Environment variables
 
+We have three production environment si.e. where Rails is running in
+'production' with RAILS_ENV="production":
+
+- development
+- test
+- production
+
+These environments require extra variables to function, these are stored in the
+Terraform tfvar files.
+
 `SQL_SERVER_SCHEMA_NAME`
 
-This is the SQL Server schema name used in production and
-pre-production/staging. Must be set in those environments, not used locally, see
+This is the SQL Server schema name used in production environments and must be
+set in those, not used locally, see
 [SQL Server documentation](./microsoft-sql-server.md) for more information on
 schema in this context.
+
+`HOSTNAME`
+
+This is the host name for the running application in production environments, we
+set this into the `Rails.application.config.hosts` array, see:
+
+https://guides.rubyonrails.org/configuring.html#actiondispatch-hostauthorization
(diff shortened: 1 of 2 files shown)


jacksonj04:
Is `HOSTNAME` something Azure gives us for free, or do we also want to make sure we add it to the `.env` examples and the terraform so it will fail a deploy if it doesn't exist?

mec:
> Is `HOSTNAME` something Azure gives us for free, or do we also want to make sure we add it to the `.env` examples and the terraform so it will fail a deploy if it doesn't exist?

Not that I know of, we may be able to read the container hostname and pass it through - but not at this stage.

As this is only for prouction environments, so development, test and production we can't use .env files, all the variables are in the tfvar files that are managed through Terraform and stored in Azure - once thing settle down we can show you how all this happens.

Documenting them is a good plan though, we have https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/main/doc/environment-variables.md for variables that are not used locally, I can add HOSTNAME there.

doc/environment-variables.md
```diff
@@ -1,8 +1,25 @@
 # Environment variables
 
+We have three production environemnts i.e. where Rails is running in
+'prodcution' with RAILS_ENV="production":
```
jacksonj04, line 4:
```suggestion
'production' with RAILS_ENV="production":
```

doc/environment-variables.md
```diff
@@ -1,8 +1,25 @@
...
-pre-production/staging. Must be set in those environments, not used locally, see
+This is the SQL Server schema name used in production environments and must be
+set in those, not used locally, see
 [SQL Server documentation](./microsoft-sql-server.md) for more information on
 schema in this context.
+
+`HOSTNAME`
+
+This is the host name for the running application in production environments, we
+set this into the `Rails.applicaiton.config.hosts` array, see:
```
jacksonj04, line 23:
```suggestion
set this into the `Rails.application.config.hosts` array, see:
```

jacksonj04 (approved):
Couple of quick typos, but otherwise 🛫

PR #270, opened 2022-10-21
(743) Split "action" into its own schema definition file

jacksonj04:
TRELLO-U4zyQ5RH

## Changes

This is originally part of #238, but it was adding a lot of noise to the changes so it's been sliced out as a standalone PR.

Since the "action" object is starting to become more complex and carry additional options, move it into its own definition file so that our schemas are smaller, more focussed and more easily understood.

This does not change the structure or validity of any of the workflow files; instead it changes how we organise the definition of what a workflow file should look like.

## Checklist


jkempster34 (approved):
Looks good. Thanks 🙌

PR #271, opened 2022-10-21
(714) Add project closed date to the project index view

jacksonj04:
TRELLO-oebBbkgX

## Changes

- To surface the fact that a project has been closed whilst in the list view, display the closed date where it exists.
- Since it's unlikely for people to want access to closed projects as often, sort them to the bottom of the list view.

## Checklist

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -26,6 +26,12 @@ class Project < ApplicationRecord
 
   scope :by_target_completion_date, -> { order(target_completion_date: :asc) }
 
+  # This works under MSSQL because it puts NULL at the front, and we can't make it more robust because TinyTDS won't
+  # play nicely with things like IS NULL and NULLS LAST. If you're running this under a different database and the
+  # order is suddenly inverted, go check out https://michaeljherold.com/articles/null-based-ordering-in-activerecord/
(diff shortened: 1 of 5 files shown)


app/models/project.rb
```diff
@@ -25,6 +25,12 @@ class Project < ApplicationRecord
 
   scope :by_target_completion_date, -> { order(target_completion_date: :asc) }
 
+  # This works under MSSQL because it puts NULL at the front, and we can't make it more robust because TinyTDS won't
+  # play nicely with things like IS NULL and NULLS LAST. If you're running this under a different database and the
+  # order is suddenly inverted, go check out https://michaeljherold.com/articles/null-based-ordering-in-activerecord/
+  # and see if you can use Arel to build a proper query.
+  scope :by_closed_state, -> { order(:closed_at) }
```
jkempster34, line 32: This seems to work as an option from [here](https://stackoverflow.com/a/50261786), but I agree that what you have done is simpler.

```
  scope :by_closed_state, -> { Project.order(Arel.sql('CASE WHEN closed_at IS NULL THEN 1 ELSE 2 END, closed_at DESC')) }
 ```

PR #272, opened 2022-10-21
Chore: Configure ActionMailer in production

mec:
## Changes
Sending emails will not work without being configured.

Here we add the default hostname so links to the application in emails work and configure ActionMailer to use GOV.UK Notify to actually send emails - this is already configured in local development.

We don't have Sidekiq configured in production environments just yet, so for now, these emails are sent asynhronusly, see:

https://guides.rubyonrails.org/action_mailer_basics.html#calling-the-mailer

## Checklist


jacksonj04 (approved):
🚀

PR #279, opened 2022-10-27
(740) Add errors when project dates are invalid

jkempster34:
## Changes

Currently, when an invalid date such as 42,01,2022 or -1,01,2022 is set, we see a MultiParameter error.

Unfortunately, there doesn't seem to be any particularly nice way to fix this.

This overrides the setter for the Project date attributes. We rescue from any invalid dates, and add a custom error. Simply setting the error in the `rescue` won't work as the date is nillfied and we will see the error associated with the `presence` validation. 

This solution does not show the user their previous incorrect input. However, to enable that does not seem worth the effort at this point.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fine-tuned padding for actions in various type/state combinations to closer
   match the prototype.
 
+#### Fixed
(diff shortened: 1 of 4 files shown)


jkempster34:
[Improve date field validation](https://trello.com/c/MNdanj62/740-improve-date-field-validation)

jacksonj04 (approved):

PR #286, opened 2022-11-01
Add environment banner

beeLorna:
## Changes

We want to know which environment we are currently in to help with development purposes. 
We have added an environment banner which displays this. 
We are currently using the sentry environment to control this.

Screenshot 2022-10-31 at 15 25 29

## Checklist

### DIFF
--- a/app/helpers/environment_banner_helper.rb
+++ b/app/helpers/environment_banner_helper.rb
@@ -1,0 +1,21 @@
+module EnvironmentBannerHelper
+  ENVIRONMENT_COLOURS = {
+    local_development: "purple",
+    development: "turquoise",
+    test: "orange"
+  }.freeze.with_indifferent_access
+
+  def environment_banner
+    return if ENVIRONMENT_COLOURS.keys.none?(sentry_env)
+
+    govuk_tag(
+      text: "#{sentry_env.tr("_", " ").upcase} ENVIRONMENT",
+      colour: ENVIRONMENT_COLOURS[sentry_env],
+      classes: "environment-banner"
+    )
+  end
+
+  private def sentry_env
+    ENV.fetch("SENTRY_ENV")
+  end
+end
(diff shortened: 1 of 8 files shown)


app/helpers/environment_banner_helper.rb
```diff
@@ -0,0 +1,21 @@
+module EnvironmentBannerHelper
+  ENVIRONMENT_COLOURS = {
+    local_development: "purple",
+    development: "turquoise",
+    test: "orange"
+  }.freeze.with_indifferent_access
```
jacksonj04, line 6: TIL about `with_indifferent_access`!

spec/helpers/environment_banner_helper_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    context "when SENTRY_ENV is test" do
+      before { allow(ENV).to receive(:fetch).with("SENTRY_ENV").and_return("test") }
+
+      it { expect(subject).to eql(environment_tag("orange", "TEST ENVIRONMENT")) }
+    end
+  end
+
+  private def environment_tag(colour, environment_name)
+    "<strong class=\"govuk-tag govuk-tag--#{colour} environment-banner\">#{environment_name}</strong>"
```
jacksonj04, line 33: I think this could use https://govuk-components.netlify.app/components/tag/ with `classes`

jacksonj04 (approved):
One possible (non-blocking) improvement, but otherwise LGTM!

beeLorna:
> Actually a followup I've just remembered: we probably want to add some documentation that this is dependent on `SENTRY_ENV` and not `RAILS_ENV`, and the difference between the two (eg test is a `production` `RAILS_ENV` but a `test` `SENTRY_ENV`)

What's the best way to document this?

spec/helpers/environment_banner_helper_spec.rb
beeLorna: We did actually look at using this format, but it meant it was difficult to check the expected responses within the helper tests. But will keep this format in mind for future work 👍

PR #290, opened 2022-11-01
Fix double links in task list

jkempster34:
## Changes

We (I) have made a mistake when copying over the task list html, and nested our links within an empty anchor tag (`href="#"`). This has been raised during an accessibility audit, because the assistive technology used is reading out all of the blank links.

To fix, remove the empty link, and move the `aria-describedby` attribute to the correct task link.
## Checklist


jacksonj04 (approved):
🚀

PR #291, opened 2022-11-01
(765) Allow test suite to use real browser for capybara tests

beeLorna:
## Changes

Set up selenium to work within CI

This is our first step in automating our accessibilty tests.
The next step is to make our test feature test be more specific for an accessibility test and to install axe-core-selenium gem as the accessibility tool.

## Checklist


mec (approved):
If I were being super, super picky,  I would like to see this split into small commits with some explanation in the commit messages about why we made the decisions we did.

I will leave it up to you though as this work still stands 👍

PR #294, opened 2022-11-04
(chore) Enable Sidekiq in production

mec:
Our infrastructure is in place so we can enable Sidekiq in production environments.


jacksonj04 (approved):
Let's go queueing!
🧍🏼🧍🏼🧍🏼🧍🏼🧍🏼🧍🏼🧍🏼🧍🏼

PR #297, opened 2022-11-04
(739) Projects are no longer created if task list creation fails

jacksonj04:
TRELLO-xPjY0AZI

## Changes

Previously, since we created projects before attempting to populate the task list, it was possible to reach a condition (if task list creation failed for any reason) where there would be a project with no tasks, or only part of the task list.

This change wraps the entire project creation process in a database transaction, so it either succeeds or fails as an entirety.

## Checklist


lozette (approved):
LGTM!

PR #303, opened 2022-11-07
(768) Form model for project

jkempster34:
### Move Project validation into a form model
The project model and controller are getting a bit thick.

This moves the project validation into a form model. For now, the form model is
responsible for creating the handover note, and the logic surrounding the
creation of a new project remains in the controller.

### Move project creation into a service
Move the logic around project creation into a service.

In order to do this, we need to create a note form model as well. We may want
to migrate the NotesController to use this to remove any duplication.

## Checklist


jkempster34:
[Refactor ProjectController](https://trello.com/c/GAI7lOH0/768-refactor-projectcontroller)

mec:
I've made this a draft whilst we look at the issue in development that may or may not relate to the previous version of this.

mec:
The issue in development seems to have been resolved by reverting the work this incorporates, i.e the `target_conversion_date` and `advisory_board_date` setter overrides.

There is no guarentee this was the cause, there are just too many variables here, however given the time constraint, I feel it's best we go ahead and assume that to be the case.

The author of this work is on holiday this week, so this work risks becoming quite stale. 

On balance, we will close this PR and introduce a new one that attempts to meet the same goals: 

- a specific validation context for creating new projects
- a way to pre-validate the multiparameter date fields
- a way to create the new project and note
- a place to trigger notifications

Which are from the card: https://trello.com/c/GAI7lOH0

PR #307, opened 2022-11-09
(792) Add type to project model

jacksonj04:
TRELLO-jXNyxFiH

## Changes

The project model now stores a `project_type` to distinguish one workflow from another. All created projects default to `conversion` to keep the behaviour the same.

## Checklist


mec:
We've now learnt that there may be only tiny diferences between the regional delivery officer and regional case worker conversion projects, until we confrim, I will close this - we can pick it back up once we have more confidence in the direction this is all going.

PR #308, opened 2022-11-09
(chore) Configure simple cov to output json

mec:
## Changes
We would like Sonarcloud to pick up our coverage, according to this:

https://docs.sonarcloud.io/enriching/test-coverage/test-coverage-parameters/

Outputting json is required so this configures simple cov.

At this point, I have not idea how sonarcloud can read the file, but this might be a 'it just works' things so we are happy to try as this has no other impact that we can see.


mec:
I kind of suspected this would not work. I think we have to run sonar cloud in a github action so we can ensure it runs after simple cov collects our coverage stats.

I will close this for now.

PR #309, opened 2022-11-11
After opening redact funding docs becomes redact documents

titlescreen:
Change copy as it is not only funding documents that can be redacted and set for publishing on GOV.UK

## Changes

- Title and slug change to be Redact and send documents to be published on GOV.UK
- guidance no longer mentions funding docs only
- guidance indicates that documents may end up published on GOV.UK
- I also changed the slug and it did not seem to break anything but I thought I'd better highlight that

localhost_3000_projects_CAEF0ED1-E49A-4D9E-8529-FBEB9DF51418_tasks_B2362B81-C138-4DE0-B12B-55F64DB308D6

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,9 +20,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 #### Changed
 
-- After opening => redact and send task and actions no longer explicitly mention
-  funding documents only as other documents may need to be sent and publised on
-  GOV.UK
 - Fine-tuned padding for actions in various type/state combinations to closer
   match the prototype.
 - The support email has been swapped out for a complete team shared inbox.
@@ -36,6 +33,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Content
 
+- After opening => redact and send task and actions no longer explicitly mention
+  funding documents only as other documents may need to be sent and publised on
+  GOV.UK
 - Tenancy at will adds action to explicitly indicate that the email from the
   school has been received.
 - Subleases adds action to explicitly indicate that the email from the school
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 #### Changed
 
+- After opening => redact and send task and actions no longer explicitly mention
+  funding documents only as other documents may need to be sent and publised on
+  GOV.UK
 - Fine-tuned padding for actions in various type/state combinations to closer
   match the prototype.
 - The support email has been swapped out for a complete team shared inbox.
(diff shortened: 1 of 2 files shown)


app/workflows/lists/conversion/sections/after-opening.yml
```diff
@@ -9,12 +9,11 @@ tasks:
     actions:
       - title: Email the regional delivery officer
 
-  - slug: redact-and-send-funding-agreement-documents
-    title: Redact and send funding agreement documents
+  - slug: redact-and-send-documents-to-be-published-on-gov-uk
+    title: Redact and send documents to be published on GOV.UK
     hint: |
```
jkempster34, line 14: We have some wayward pipes `|` that could be removed while we are here

app/workflows/lists/conversion/sections/after-opening.yml
```diff
@@ -37,7 +36,7 @@ tasks:
       - title: Send relevant documents to the funding team
         guidance_summary: Help sending documents
         guidance_text: |
```
jkempster34, line 38: Another one

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 #### Changed
 
+- After opening => redact and send task and actions no longer explicitly mention
```
jkempster34, line 22: I think this should go in the `Content` section ideally

jkempster34 (changes requested):
👋 I think it would be nice to make just a couple of small changes to tidy up

app/workflows/lists/conversion/sections/after-opening.yml
titlescreen: Just to refresh my brain the Pipe means something like "output this as literal HTML" so the markdown becomes html?

app/workflows/lists/conversion/sections/after-opening.yml
titlescreen: I think this one needs to stay on the guidance text as it has markdown for a link in it.

CHANGELOG.md
titlescreen: Yeah good shout, not sure why I forgot about the content section this time 🤷‍♀️ 😁

app/workflows/lists/conversion/sections/after-opening.yml
jkempster34: It means to preserve newlines.

If we do this (no pipe):
```yaml
hint:
  List:
  * thing
  * thing
```

We will get the following output (there won't be any formatting because the new line is necessary):
```
List: *thing *thing
```

So, we need the pipe when we want spaces between paragraphs, or to use bullet points or numbered lists.

app/workflows/lists/conversion/sections/after-opening.yml
jkempster34: This actually doesn't need the pipe. We only need it for newlines

jkempster34 (approved):
Approved as the pipe doesn't matter for now!

Hope my explanation made sense

PR #311, opened 2022-11-14
(refactor) Extract and tidy validations

mec:
## Changes
Ahead of work to introduce a form object for creating new projects, here we extract the validation from the Project model into validators that can be reused.


jkempster34 (approved):
I think there is an argument that this adds an avoidable level of indirection, but I think for the purpose of thinning out the model it looks good

✅

PR #314, opened 2022-11-14
Rename `closed_at` column on Projects to `completed_at`

lozette:
## Changes

We are going to change the wording of "closed" projects on the front end to "completed". For clarity, we have decided to keep the front and back end terminology the same, which means renaming the `closed_at` column on the Project model to `completed_at`.

This change also means changes to the relevant controllers, methods, tests, views and text strings.

Part of https://trello.com/c/liZ2Yqcv/795-build-a-way-to-visualise-closed-project

## Checklist


spec/requests/project_complete_spec.rb
```diff
@@ -0,0 +1,30 @@
+require "rails_helper"
+
+RSpec.describe "Closing projects", type: :request do
```
jkempster34, line 3: RSpec.describe "Completing projects"

jkempster34 (approved):
Two more sorry (not important ones) 🙈 

Approved so that I'm not blocking you! 🚀

mec (approved):
Wow! This is why getting the language right early matters! Thanks for taking the time to switch this. 🚢

PR #315, opened 2022-11-14
Set up accessibility tool

beeLexington:
## Changes

`axe-core-rspec` gem has been used as the accessibility tool for our automated accessibilty tests.

Next step is to write accessibilty tests for the `axe-core-rspec` gem to run against.

## Checklist

### DIFF
--- a/.github/workflows/continuous-integration-tests.yml
+++ b/.github/workflows/continuous-integration-tests.yml
@@ -8,6 +8,41 @@ on:
       - develop
 
 jobs:
+  accessibility-checks:
+    name: Accessibility checks
+    runs-on: ubuntu-latest
+
+    env:
+      RAILS_ENV: test
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - id: cache-docker
+        uses: actions/cache@v3
+        with:
+          path: /tmp/docker-save
+          key:
+            docker-save-${{ hashFiles('Dockerfile', 'Gemfile.lock',
+            'package-lock.json') }}
+
+      - if: steps.cache-docker.outputs.cache-hit == 'true'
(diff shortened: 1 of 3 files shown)


mec:
@beeLorna we chatted about making the axe specs not run as part of the default test suite, we want to only run the axe tests when we ask for them sepcifically so they do not block us from merging work. I am hoping we can use Rspec filters and tags to achieve this: https://relishapp.com/rspec/rspec-core/v/3-12/docs/filtering/exclusion-filters

Give me a yell if you want to chat again.

spec/features/test_for_ci_spec.rb
```diff
@@ -1,4 +1,9 @@
 require "rails_helper"
+require "axe-rspec"
+
+RSpec.configure do |c|
+  c.filter_run_excluding broken: true
```
mec, line 5: I am not sure `broken` is the right name here? These are going to be accessibility test so perhaps `accessibility: true` would be better?

spec/features/test_for_ci_spec.rb
```diff
@@ -7,9 +12,9 @@
     sign_in_with_user(user)
   end
 
-  scenario "the test runs", driver: :headless_firefox do
+  scenario "the test runs", driver: :headless_firefox, broken: true do
```
mec, line 15: With the name fixed, we could move `driver: :headless_firefox` and `broken: true`) up to line 8 so the whole spec file runs with those options.

mec (changes requested):
NIce! I think we can make this a little closer to the goal.

I would like to see how we can add a workflow or job to CI that runs only these accessibility specs as part of this work, let me know if you want to work on that together?

mec (approved):

PR #319, opened 2022-11-15
Build a way to visualise completed projects on the Project index page

lozette:
## Changes

We want to make completed projects more easily findable on the index page.

Previously, completed (or "closed") projects were pushed to the end of the
project list, so it wasn't entirely obvious where they were. This commit splits
the index page into two actions, and presents them on separate tabs on the
project index page.

The first tab is "In progress" projects, i.e projects which are open (do not
have a `completed_at` date). The second tab is for "Completed" projects. Both
tabs have their own route & action.

As a side effect of this, the layout for the index page has been put into a
partial, to be included in the views for both the `index` and `completed`
actions.

The relevant tests have been updated, and a new test for the completed project
view has been added.

Screenshot 2022-11-15 at 11 45 52

Trello: https://trello.com/c/liZ2Yqcv/795-build-a-way-to-visualise-closed-project 

## Checklist


lozette:
[Build a way to visualise closed project](https://trello.com/c/liZ2Yqcv/795-build-a-way-to-visualise-closed-project)

config/routes.rb
```diff
@@ -22,6 +22,8 @@
     get "/delete", action: :confirm_destroy
   end
 
+  get "completed", to: "projects#completed"
```
lozette, line 25: I'm not 100% convinced this is the best route. Would `projects/completed` make more sense?

spec/models/project_spec.rb
```diff
@@ -301,4 +301,34 @@
       end
     end
   end
+
+  describe "completed scope" do
```
jkempster34, line 305: Might be nice to wrap these in a `describe "Scopes"` which is what we have done in the other spec files

config/routes.rb
jkempster34: I think it would make more sense!

jkempster34 (changes requested):
Looks great, like the tabs! :pinched_fingers:

Found a few things:
- One of the commit titles is a bit too long for GitHub and is wrapping
- I defo prefer `projects/completed` to `completed`
- There is some code that we can remove now. The `by_completed_state` scope, and where it's called in the `project_policy`

app/models/project.rb
```diff
@@ -33,6 +33,9 @@ class Project < ApplicationRecord
   # and see if you can use Arel to build a proper query.
   scope :by_completed_state, -> { order(:completed_at) }
 
+  scope :completed, -> { where.not(completed_at: nil) }
+  scope :open, -> { where(completed_at: nil) }
```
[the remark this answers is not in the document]
lozette: Oh not seen that before but not a fan, I think it makes it more obscure!

jkempster34:
> Added text to display when the tabs are empty
> 
> Screenshot 2022-11-15 at 15 47 22

Nice! Is it worth matching it to the `No notes` messages and putting it in an inset text?

image

lozette:
> 

I've added it as per https://lucid.app/lucidspark/59a224c0-3bb3-4a0f-ace4-199138d009f5/edit?invitationId=inv_a2eb4a53-99fa-4945-93a6-aeb92467c134# so I'm not sure - @Rich-Cooley any thoughts?

app/models/project.rb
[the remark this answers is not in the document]
jkempster34: Agree!

app/views/projects/index/_project_index.html.erb
```diff
@@ -0,0 +1,39 @@
...
+    <% if policy(:project).new? %>
+      <p><%= link_to t("project.index.new_button.text"),
+               new_project_path,
+               class: "govuk-button",
+               data: {module: "govuk-button"} %></p>
+    <% end %>
+
+    <%= render partial: "shared/project_index_sub_navigation" %>
+
+    <%- if @projects.empty? %>
```
jkempster34, line 19: We haven't used these `-` anywhere else, I'm not sure what they do, or do they just do the same as omitting the `-`?

config/routes.rb
```diff
@@ -22,6 +22,8 @@
     get "/delete", action: :confirm_destroy
   end
 
+  get "projects/completed", to: "projects#completed"
```
jkempster34, line 25: I think we can put this inside the block below like:

```ruby
resources :projects, only: %i[index show new create] do
  collection do
    get "completed"
  end
```

jkempster34 (approved):
Left two picky stylistic comments, and one about the routing (I looked into this recently to tidy it up).

Looks great 💯

app/views/projects/index/_project_index.html.erb
lozette: You know what I think these are quite old erb style and I haven't used Ruby for a year so I'm out of date 😆

PR #306, opened 2022-11-09
Revert "Add errors when project dates are invalid"

mec:
## Changes

The development environment is currently broken, when creating a new project we see the following errors:

```
2 error(s) on assignment of multiparameter attributes [error on
  assignment [1, 1, 2023] to target_completion_date (super: no
  superclass method `target_completion_date=' for #<Project > Did you
  mean?  target_completion_date_is_in_the_future),error on assignment
  [1, 10, 2022] to advisory_board_date (super: no superclass method
  `advisory_board_date=' for #<Project >)]
```

This behaviour cannot be replicated locally, right now our next best step is to remove this work and test in development.

Later work moves the multiparameter validation introduced here, so ruling this out now would be helpful.

This reverts commit f05b64bbc21e649bb67811e5bab5ee816f8c3b86.

PR #321, opened 2022-11-16
(768) Create project form object

mec:
## Changes

The motivation for this work is a bug we have where a user enters an invalid value into the three part date field, e.g. a month value as 31 or letters instead of numbers.

Screenshot 2022-11-16 at 13 59 20

Screenshot 2022-11-16 at 13 59 28

This is a pattern from the [GOV.UK design system](https://design-system.service.gov.uk/components/date-input/) so it's surprising to have to deal with it ourselves rahter than being handled by the [GOV.UK formbuilder](https://govuk-form-builder.netlify.app/) that we use

It is also pretty hard to fix as the 3 param dates is handled by Rails magic™ the three params are cast into a Date object, but as we have learnt, if the params cannot be cast, the application crashes.

We think this is because in generic Rails a Date object exposed in a form is made up of select boxes which the user cannot submit incorrect params - but we are not 100% on this, it feels pretty brittle...

Whilst we had a previous [fix](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/306), this led us to crashes in development and whilst we are not clear on the issue, we have to assume this was the  cause as reverting the change fixed the problem.

Here we are introducing a form object, a design pattern that places a 'form object' between the form being submitted and the models being updated. This gives us a number of good things:

- a context to validate a new project in
- a place to deal with the multiparameters from dates in the form
 - a different option to handle creating multiple objects from one form, a Project and Note in this case
- a place to trigger any notifications
- a way to thin out the controller

At the cost of duplication.

I prefer this approach as it seperates the model, which should accept a date object for attributes representing dates from the form submission with has the three parameters - it seperates the concerns.

This work is proving to be large, so I want to open this PR now which introduces the form object, validation and multiparameter handling, but does not use it to save anything. 

We can follow up with more changes to actually use the new form object.

https://trello.com/c/GAI7lOH0

## Checklist

### DIFF
--- a/app/forms/create_project_form.rb
+++ b/app/forms/create_project_form.rb
@@ -1,0 +1,81 @@
+class CreateProjectForm
+  include ActiveModel::Model
+  include ActiveRecord::AttributeAssignment
+
+  include ::MultiparameterDate
+
+  attr_accessor :urn,
+    :incoming_trust_ukprn,
+    :establishment_sharepoint_link,
+    :trust_sharepoint_link,
+    :advisory_board_conditions,
+    :note_body,
+    :user,
+    :attributes_with_empty_values,
+    :attributes_with_invalid_values
+
+  attr_reader :target_conversion_date,
+    :advisory_board_date
+
+  validates :urn,
+    :incoming_trust_ukprn,
+    :establishment_sharepoint_link,
+    :trust_sharepoint_link, presence: true
+
+  validates :urn, urn: true
+  validates :incoming_trust_ukprn, ukprn: true
+
+  validates :target_conversion_date, date_in_the_future: true
+  validates :advisory_board_date, date_in_the_past: true
+
+  validate :multiparameter_date_attributes_values
+  validate :multiparameter_date_attributes_empty
+
+  def initialize(params = {})
+    @attributes_with_empty_values = []
+    @attributes_with_invalid_values = []
+    super(params)
+  end
+
+  def target_conversion_date=(value)
+    if month_for(value).nil? && year_for(value).nil?
+      @attributes_with_empty_values << :target_conversion_date
+      return
+    end
+
+    if date_parameter_valid?(value)
+      @target_conversion_date = date_from_multiparameter_hash(value)
+    else
+      @attributes_with_invalid_values << :target_conversion_date
+    end
+  end
+
+  def advisory_board_date=(value)
+    if value.nil? || (day_for(value).nil? && month_for(value).nil? && year_for(value).nil?)
+      @attributes_with_empty_values << :advisory_board_date
+      return
+    end
+
+    if date_parameter_valid?(value)
+      @advisory_board_date = date_from_multiparameter_hash(value)
+    else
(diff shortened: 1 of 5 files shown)


lozette (approved):
This looks good although complex!

mec:
I am in two minds about this, we still need to deal with the multiparameter dates, but this will need to be updated for the new `Conversion::Project` work.

@lozette comment on the complexity also worries me - there may be a simpler way to handle this, but we have sunk a great deal of time into the issue already!

I'll leave this here until #344 goes in and then we can make a decision.

peteryates:
Sorry, just got round to taking a proper look at this.

The way the form builder deals with dates is exactly the same as [Rails itself does](https://guides.rubyonrails.org/form_helpers.html#using-date-and-time-form-helpers). You shouldn't have to do any extra work to pull a date out of params, you shouldn't even need to be aware they're passed in using the `1i`/`2i`/`3i` format.

Doing a `params.require(:some_record).permit(:subscribed_on)` should give you a valid date. As I mentioned on Slack, the date might not always be what you expect (as `Time.new(2022, 9, 31) => 2022-10-01`) but for the most part it works identically to Rails' own implementation.

mec:
> Sorry, just got round to taking a proper look at this.
> 
> The way the form builder deals with dates is exactly the same as [Rails itself does](https://guides.rubyonrails.org/form_helpers.html#using-date-and-time-form-helpers). You shouldn't have to do any extra work to pull a date out of params, you shouldn't even need to be aware they're passed in using the `1i`/`2i`/`3i` format.
> 
> Doing a `params.require(:some_record).permit(:subscribed_on)` should give you a valid date. As I mentioned on Slack, the date might not always be what you expect (as `Time.new(2022, 9, 31) => 2022-10-01`) but it mimics Rails' own implementation but with `<input>` elements rather than `<select>`.

@peteryates no problem at all - thanks for having a look!

Yep, the form builder does produce what Rails would, our problem is because users can add a string instead of an integer, or a incorrect value, e.g. 31 for a month because the fields are text input rather than selects, this raises an `ActiveRecord::MultiparameterAssignmentErrors` rather than an invalid form submission - which is not ideal for our users.

The only way we've found to rectify this is overriding the setters as here. 

The issue seems to be that conversion from the hash we get to the `Date` object happens when Rails is mass assigning and so it is difficult to actually get at in the context of the object rather than the controller, so making the object invalid is hard.

As you pointed out, the default Rails behaviour is a set of three `<select>` which gets around the problem, albeit not brilliantly! Another option would be to reach for a client side JS solution, which is something we generally tried to avoid or even using `<input type="number">` none of which feel right in our context.

The best write up I have found is here:

https://thoughtbot.com/blog/ruby-on-fails

Long term I wonder if there is a different approach to the markup in the form to make this a bit more approachable to get to during validation, but we don't have the time in our delivery plan right now!

mec:
We are closing this PR for now as we now have other work to introduce a form object for creating new conversion projects. 

Once we have the basic form object working, we will come back and fix the multiparameter date submission.

PR #325, opened 2022-11-17
Show a specific timeout error when the Academies API times out

lozette:
## Changes

If the Academies API times out, show an error page to the user, indicating they might be able to try again.

Previously, the users were being shown a less user-friendly error.

Screenshot 2022-11-17 at 16 50 37

## Checklist

### DIFF
--- a/app/views/pages/api_client_timeout.html.erb
+++ b/app/views/pages/api_client_timeout.html.erb
@@ -1,0 +1,14 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l"><%= t("pages.api_client_timeout.title") %></h1>
+    <p class="govuk-body">
+      <%= t("pages.api_client_timeout.body_text") %>
+    </p>
+
+    <p class="govuk-body">
+      <%= t("pages.api_client_timeout.email_help",
+            team_name: t("service_name"),
+            email: support_email(Rails.application.config.support_email)).html_safe %>
+    </p>
+  </div>
+</div>
--- a/spec/requests/projects_controller_spec.rb
+++ b/spec/requests/projects_controller_spec.rb
@@ -85,5 +85,15 @@ RSpec.describe ProjectsController, type: :request do
         expect(Project.count).to be 0
       end
     end
+
+    context "when the Academies API times out" do
+      before do
+        mock_timeout_api_responses(urn: 123456, ukprn: 10061021)
+      end
+
+      it "redirects to an informational client timeout page" do
+        expect(perform_request).to render_template("pages/api_client_timeout")
+      end
+    end
   end
 end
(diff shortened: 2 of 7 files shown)


app/views/pages/api_client_timeout.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l"><%= t("pages.api_client_timeout.title") %></h1>
+    <p class="govuk-body">
+      <%= t("pages.api_client_timeout.body_text") %>
+    </p>
+
+    <p class="govuk-body">
+      <%= t("pages.api_client_timeout.email_help",
```
jkempster34, line 9: The other error pages don't use the locales, but maybe they should because we use locales a lot generally.

Should we add a ticket to switch over the other error pages?

app/views/pages/api_client_timeout.html.erb
```diff
@@ -0,0 +1,14 @@
...
+            team_name: t("service_name"),
+            email: support_email(Rails.application.config.support_email)).html_safe %>
```
jkempster34, line 11: I've seen this somewhere else. I think it might be nice to improve the `support_email` helper with `support_email(name: nil)` for when we don't want to pass a name, and document it in the helper specs.

spec/support/academies_api_helpers.rb
```diff
@@ -23,4 +23,23 @@ def mock_successful_api_trust_response(ukprn:, trust: nil)
     allow(test_client).to receive(:get_trust).with(ukprn).and_return(fake_result)
     allow(AcademiesApi::Client).to receive(:new).and_return(test_client)
   end
+
+  def mock_timeout_api_responses(urn:, ukprn:)
```
jkempster34, line 27: 🙌

spec/requests/projects_controller_spec.rb
```diff
@@ -85,5 +85,17 @@
         expect(Project.count).to be 0
       end
     end
+
+    context "when the Academies API times out" do
+      before do
+        mock_timeout_api_responses(urn: 123456, ukprn: 10061021)
+
+        perform_request
```
jkempster34, line 93: This should be able to go

jkempster34 (commented):
We have been adding feature specs for the static pages. I think its debatable how useful they are, but we should probably add one just for consistency. 🤔 

They are in `features/users_can_view_static_pages_spec`.

app/views/pages/api_client_timeout.html.erb
lozette: Yeah I think a tech debt ticket to get everything translated would be good

spec/requests/projects_controller_spec.rb
lozette: So it can!!

jkempster34:
> We have been adding feature specs for the static pages. I think its debatable how useful they are, but we should probably add one just for consistency. 🤔
> 
> They are in `features/users_can_view_static_pages_spec`.

@lozette Did you see this comment?

Apart from that I'm ready to ✅ both PRs. Thanks for improving the support email function!

lozette:
Missed that @jkempster34 but it's added now!

jkempster34 (approved):
LGTM 🎺 

Might need to update the CHANGELOG!

PR #328, opened 2022-11-18
(chore) Update Sentry configuration

mec:
## Changes
We have decided not to send Sentry data in our local development
environments. Sentry is billed per 'event' so we want to minimise how
many we send.
    
Here we remove the two environment variables relating to Sentry for
local development and guard against configuring sentry unless we are in
a production environment i.e one of our live environments.
    
We have to update the environment banner helper to provide a default
when the SENTRY_ENV is not present.

https://trello.com/c/T6ghR2oV


app/helpers/environment_banner_helper.rb
```diff
@@ -16,6 +16,6 @@ def environment_banner
   end
 
   private def sentry_env
-    ENV.fetch("SENTRY_ENV")
+    ENV.fetch("SENTRY_ENV", "local_development")
```
jkempster34, line 19: Just wanted to point out a potential bug.

If SENTRY_ENV is absent in any of the live environments, we will see the `LOCAL DEVELOPMENT ENVIRONMENT` banner. This might not be a problem if we are confident that SENTRY_ENV will never be absent

jkempster34 (approved):
Approved ✅  but left a comment about a potential bug, might be fine however!

app/helpers/environment_banner_helper.rb
lozette: +1, possibly we should refactor the banner to use a separate env variable (in a separate ticket)

lozette (approved):

app/helpers/environment_banner_helper.rb
mec: Totally valid! Thanks both. I guess it's a long shot as we have all three live environments now and the `SENTRY_ENV` is set for all of them.

This feels like a card for a proper backlog? As I have the approvals - i'll go do this and link it here before merging.

mec: https://trello.com/c/SSBwCdx9

PR #339, opened 2022-11-23
(chore) Add allowed hosts to Rails hosts

mec:
## Changes
Once we put the applicaiton behind a CDN we found that Rails built in host authorization kicked in, preventing the application from responding via the CDN.

Adding this, allows us to specify all of the possible hostnames the host authorization validation may encounter.

We retain the sepearate `HOSTNAME` to specify the external host name users use when accessing the application as this is used to build links.


Stretch96 (approved):
lgtm

PR #344, opened 2022-11-25
Add conversion projects

lozette:
## Changes

This PR creates the concept of a Conversion Project. Eventually, the application will feature other types of projects, e.g. Transfer projects, so we needed to separate out the concept of different types of projects.

The current project journey we have is for "voluntary" conversion projects. We also need to add "involuntary" conversion projects. In order to keep this PR relatively small we have decided to:

- Keep the current Project journey
- Introduce Conversion projects "under the hood" (users cannot currently access this journey)
- Introduce Involuntary conversion projects as a subtype of conversion projects, with their own form and tasklist

## Checklist


lozette:
[The application supports two Converion journeys](https://trello.com/c/U3QFK7Lg/858-the-application-supports-two-converion-journeys)

spec/features/users_can_create_involuntary_conversion_projects_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      expect(page).to have_content(I18n.t("project.show.title"))
+      expect(page).to have_content("Involuntary task")
+
+      click_on("Project information")
+
+      expect(page).to have_content(two_weeks_ago.to_formatted_s(:govuk))
+    end
+  end
+
+  context "when some required project information is not valid" do
```
jkempster34, line 53: We have been putting the `invalid` tests in the request specs so we can keep the test suite as fast as possible. I think we are trying to reserve feature specs for the happy path. 

https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/5d1f9c4552a94a8d8341b8eaa8d63dbd301c7fc4/spec/requests/notes_controller_spec.rb#L74

spec/requests/create_new_conversion_project_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "rails_helper"
+
+RSpec.describe "Create new conversion project" do
```
jkempster34, line 3: I think this could be a `conversion_project_controller_spec`, to match some of the others, but let me know what you think!

config/routes.rb
```diff
@@ -22,6 +22,15 @@
     get "/delete", action: :confirm_destroy
   end
 
+  scope :projects do
```
jkempster34, line 25: Tried to look at how we could nest this inside the `resources :projects` block, and this works, but I'm not sure if it's correct because it's not really a collection.

```ruby
resources :projects, only: %i[index show new create] do
  collection do
    get "completed"
  end

  collection do
    namespace :conversion do
      namespace :involuntary do
        get "new", to: "projects#new"
        post "new", to: "projects#create"
      end
    end
  end
```

app/controllers/conversion/projects_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Conversion::ProjectsController < ProjectsController
+  def new
+    authorize Conversion::Project, policy_class: ProjectPolicy
```
jkempster34, line 3: Do we need to cover this in the request specs, like this https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/5d1f9c4552a94a8d8341b8eaa8d63dbd301c7fc4/spec/requests/notes_controller_spec.rb#L277

It makes sense to me that the unauthorised behaviour would be tested there 🤔, but maybe somewhere else?

jkempster34 (changes requested):
Left some comments/questions mostly about specs

spec/features/users_can_create_involuntary_conversion_projects_spec.rb
lozette: I added this test because without it, coverage was less than 100%. I've moved a similar test into requests and the coverage is still 100% but it feels a bit like a useless test. Not sure how to keep coverage happy without this slightly pointless test.

config/routes.rb
lozette: I think the way we have it is correct, as we're not using a `project` resource. The `scope` syntax works because we want to have `/projects/` on the URL but the controller folders aren't nested inside `projects`. I *think* that's the intention of scopes in routes.

jkempster34: 👍

spec/requests/create_new_conversion_project_spec.rb
jkempster34: I've always done these as a `controller` spec, but I might be wrong to do that. [Theres definitely a few people who do that but not loads](https://github.com/search?p=1&q=org%3ADFE-Digital+filename%3Acontroller+path%3Aspec%2Frequests+language%3ARuby+language%3ARuby&type=Code) 

I think from my point of view it's easy to ensure you have full coverage by testing the controller actions and just having the naming match. No extra thought required 🤷 "Where shall I put the test for the new controller action I have just written in the notes_controller" -> "notes_controller_spec"

Looks like we aren't meant to use controller spec regardless 
image

app/controllers/conversion/projects_controller.rb
lozette: Added!

jkempster34 (approved):
LGTM 🚢 🙌 

I still think with the spec names, whatever direction we go in, it would be nice to try to keep them consistent/refactor to be consistent.

So like:
```
requests/
  notes_controller_spec.rb
  projects_controller_spec.rb
```

or

```
  requests/
    notes/
      create_spec.rb
      destroy_spec.rb
    projects/
      index_spec.rb
```

or something else, maybe

PR #349, opened 2022-11-29
867 content updates for the voluntary conversions workflow

titlescreen:
## Changes

Iterate content with things we learned from private beta and regional delivery officers

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -413,6 +413,7 @@ PLATFORMS
   arm64-darwin-21
   x86_64-darwin-20
   x86_64-darwin-21
+  x86_64-darwin-22
   x86_64-linux
 
 DEPENDENCIES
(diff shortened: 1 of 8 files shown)


titlescreen:
I did a thing to make it work locally then pushed. There must be a better way or I'm doing it wrong, so please let me know the best practice.

titlescreen:
Thanks Joseph, that content looks good to me.

app/workflows/lists/conversion/voluntary/sections/ready-for-opening.yml
```diff
@@ -7,7 +7,7 @@ tasks:
       the Baseline email. Indicate on the Baseline that the school is on track
       to open.
     actions:
-      - title: Confirm the school is on track to open on the target date
+      - title: Confirm the school is on track to open on the provisional date
```
lozette, line 10: The previous commit swaps these over. I'd drop the previous commit as I assume it's not needed.

Gemfile.lock
```diff
@@ -413,6 +413,7 @@ PLATFORMS
   arm64-darwin-21
   x86_64-darwin-20
   x86_64-darwin-21
+  x86_64-darwin-22
```
lozette, line 416: Can this be moved into its own commit with its own message!

lozette (changes requested):
Can you drop one commit entirely eec587467d8213f7d3310fb2557f899402aa3bed as it is overwritten in a later commit.

PR #351, opened 2022-11-29
Refactor Projects into Voluntary Conversion Projects

lozette:
## Changes

In a previous PR https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/344 we added the concept of Conversion projects, and their subtype Involuntary conversion projects. We now need to add the concept of Voluntary conversion projects.

Voluntary conversion projects are actually what we currently think of as just "projects". Therefore this PR adds the concept of Voluntary conversion projects and redirects all previous journeys for "regular projects" down the Voluntary conversion project paths. The end user will see no difference in their user experience in the app.

## Checklist


mec (changes requested):
Nice work - thanks for taking the time to make this refactor easy to follow.

Whilst testing I did come across a bug that would be worth fixing here if we can:

The new task note form now needs to be explicitally told it is still able to use a Project:

https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/main/app/views/notes/new.html.erb

here on line 7:
`<%= form_for [@project, @note], url: project_notes_path(@project) do |form| %> `

other wise it will try to do a 'new conversion project note'

mec (approved):
🚢

PR #354, opened 2022-12-01
Fix: make url to edit notes and contacts more explicit

lozette:
## Changes

When we added the concept of Conversion Projects in a previous commit, we omitted to make the paths to edit/delete notes and contacts *explicitly* the paths for projects. Because all projects will (for now) be conversion projects, but notes and contacts are attached to "plain" projects, we need to be explicit that the forms for these objects needs to submit to `project_{object}_path`, not `conversion_project_{object}_path`

## Checklist


mec (approved):
🚢

PR #357, opened 2022-12-01
871 get bulk data

jkempster34:
### Add method to get multiple establishments by URNs to client
We want to be able to fetch multiple establishments in a single request on the
projects index page.

This adds a `get_establishments` method to the API client, which calls the new
`bulk` endpoint on the API.

It also refactors the `get_establishment` method to use the `bulk` endpoint to
keep things DRY, with the benefit that performance is better on the bulk
endpoint than on the single resource endpoint.

### Add method to get multiple trusts by UKPRNs to client
We want to be able to fetch multiple trusts in a single request on the projects
index page.

This adds a `get_trusts` method to the API client, which calls the new `bulk`
endpoint on the API.

It also refactors the `get_trust` method to use the `bulk` endpoint to keep
things DRY, with the benefit that performance is better on the bulk endpoint
than on the single resource endpoint.

### Add ProjectDataInitialiser service
We need to be able to map the establishments and trusts fetched from the
`get_establishments` and `get_trusts` methods of the client to the projects.

This adds a ProjectDataInitialiser service to perform this. The URN and UKPRN
must be added to the API response so that we are able to map them to the
projects. Setters are added to the project model so that the establishment and
incoming trust can be assigned.

### Call the ProjectDataInitialiser on the Projects index page
This calls the ProjectDataInitialiser service from the project index action.
The bulk establishments and trusts are assigned to the projects, and as these
are memoized, requests to fetch individual establishments and trust will no
longer be made.

The ProjectDataInitialiser is mocked in the feature specs. While this is not
ideal, mocking the endpoints was adding a significant amount of noise and
complexity to the tests.

## Checklist


app/controllers/projects_controller.rb
```diff
@@ -5,6 +5,8 @@ class ProjectsController < ApplicationController
   def index
     authorize Project
     @pagy, @projects = pagy(policy_scope(Project.open))
+
+    ProjectDataInitialiser.new.call(@projects)
```
jkempster34, line 9: Is there a better name for this?

mec (changes requested):
I've left a couple of comments, I mostly feel like there is good reason to split the service into two.

You mentioned the mocking in features earlier on today - I think I agree there is probably the opportunity to have fake projects with establishments and trust already assigned now that we can set the values - but let's pick that up as seperate work.

Great work still! I always learn when I review you work! ⭐

mec (approved):
🚢

PR #359, opened 2022-12-02
Add sign in with helper

mec:
## Changes

We have something similar for feature specs, this just reduces noise in
our specs that are not focused on how we signed in, but need to behave
as a user who is signed in.


jkempster34 (approved):
Looks good! ✅

PR #362, opened 2022-12-05
(886) Fix accessible autocomplete on user assignment pages

jkempster34:
### Fix accessible autocomplete on user assignment pages
We enhance our select elements on the user assignment pages with the
accessible autocomplete component.

Currently, we use the ID of the select element to attach the component.
However, because this ID is generated from the form builder, it has changed
with a recent renaming of attributes.

This adds a "target" class to the select element so any further renames will
not break the autocomplete.

## Checklist


mec (approved):
If this fixes something, could we have an entry in the change log?

PR #375, opened 2022-12-09
New Involuntary conversions notificaitons

mec:
## Changes

Right now, when a new involuntary conversion project is created, there
is no value in notifying a Regional caseworker team lead as they play
no role in the projects life.

Here we remove the notification for involuntary conversions and move the
coverage from the shared example and into specific specs.

https://trello.com/c/2bLWZ5Ig


jkempster34 (approved):
LGTM 🙌

PR #376, opened 2022-12-09
896 project created confirmation tells users what to do next

titlescreen:
## Changes

Now supports bespoke project creation banner for voluntary and involuntary routes. 

This also renders the flash message as `html_safe` so that message formatting can be defined in the translations file

image
image

## Checklist

### DIFF
--- a/app/views/shared/_flash.html.erb
+++ b/app/views/shared/_flash.html.erb
@@ -8,7 +8,7 @@
   <div class="govuk-notification-banner__content">
     <% flash.each do |type, msg| %>
       <div class="flash <%= type %>">
-        <p><%= msg %></p>
+        <p class="govuk-notification-banner__heading"><%= msg.html_safe %></p>
       </div>
     <% end %>
   </div>
--- a/config/locales/conversion_project.en.yml
+++ b/config/locales/conversion_project.en.yml
@@ -5,14 +5,22 @@ en:
       new:
         title: Add a new involuntary conversion project
         hint_html: <p class="govuk-body">Enter the school and trust information.</p><p class="govuk-body">This will create a conversion project for a school that has been issued a Directive academy order.</p>
+      create:
+        success:
+          Involuntary project successfully created
+          <p class="govuk-body">Add details of any contacts you have for the school, trust, local authority and solicitors.</p>
+          <p class="govuk-body">Use the task list to manage the project.</p>
     voluntary:
       route: Voluntary
       new:
         title: Add a new voluntary conversion project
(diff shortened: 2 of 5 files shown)


config/locales/conversion_project.en.yml
```diff
@@ -5,11 +5,22 @@ en:
       new:
         title: Add a new involuntary conversion project
         hint_html: <p class="govuk-body">Enter the school and trust information.</p><p class="govuk-body">This will create a conversion project for a school that has been issued a Directive academy order.</p>
+      create:
+        success:
+          Involuntary project successfully created
+          <p class="govuk-body">Add details of any contacts you have for the school, trust, local authority and solicitors.</p>
```
jkempster34, line 11: Is there a better way of doing this, or is this okay for now?

I suppose we could pass a hash

```ruby
redirect_to project_path(@project), 
notice: {
  title: I18n.t("conversion_project.voluntary.create.success.title"), 
  body: I18n.t("conversion_project.voluntary.create.success.body") }


<% flash.each do |type, msg| %>
        <p class="govuk-notification-banner__heading"><%= msg[:title].html_safe %></p>
        <p class="govuk-body"><%= msg[:body].html_safe %></p>

```

app/views/shared/_flash.html.erb
```diff
@@ -8,7 +8,7 @@
   <div class="govuk-notification-banner__content">
     <% flash.each do |type, msg| %>
       <div class="flash <%= type %>">
-        <p><%= msg %></p>
+        <p class="govuk-notification-banner__heading"><%= msg.html_safe %></p>
```
jkempster34, line 11: Just noticed this @Rich-Cooley 

I think you would be better declaring the header in the `en` file and removing this class.

This flash file affects all of the messages across the site. So all of the notes and contacts messages will now be bold.

```yml
      create:
        success:
          <p class="govuk-notification-banner__heading">Voluntary project successfully created<p>
          <p class="govuk-body">Add details of any contacts you have for the school, trust, local authority and solicitors.</p>
          <p class="govuk-body">A caseworker will be assigned this project.</p>
          <p class="govuk-body">There is nothing more for you to do.</p>
```

titlescreen: My initial thought is yes, when a notification banner is shown then it should be in the larger text… but then I think if I were to look in the yml file it would be more readable which part was the header and which part was the body. Personally I think I'd like to keep it how it is, and add that `<p class="govuk-notification-banner__heading">` to all the notification banner messages.

jkempster34: So they will go from this:
image

To this:
image

titlescreen: Ok I have sorted that as I'm not confident on finding all the error messages and this should give consistency like you suggested.

lozette (approved):

PR #378, opened 2022-12-12
Fix bug where empty URNs and UKPRNs raise an API error

jkempster34:
## Changes
Currently, when creating a project with an empty URN or UKPRN, the project
model will try to validate these against the API. The API will throw a bad
request error, which is re-raised in the API client, and the user will see an
error page.

Instead, we can conditionally validate that the URN or UKPRN returns an
establishment or trust only if they are present. This ensures that a
blank URN or UKPRN will not call out to the API for validation, and will
instead show the error message associated with `presence: true`.

## Checklist

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -21,7 +21,8 @@ class Project < ApplicationRecord
   validates :establishment_sharepoint_link, presence: true, url: {hostnames: SHAREPOINT_URLS}, on: :create
   validates :trust_sharepoint_link, presence: true, url: {hostnames: SHAREPOINT_URLS}, on: :create
 
-  validate :establishment_exists, :trust_exists, on: :create
+  validate :establishment_exists, on: :create, if: -> { urn.present? }
+  validate :trust_exists, on: :create, if: -> { incoming_trust_ukprn.present? }
 
   belongs_to :caseworker, class_name: "User", optional: true
   belongs_to :team_leader, class_name: "User", optional: true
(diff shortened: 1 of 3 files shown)


app/models/project.rb
```diff
@@ -21,7 +21,7 @@ class Project < ApplicationRecord
   validates :establishment_sharepoint_link, presence: true, url: {hostnames: SHAREPOINT_URLS}, on: :create
   validates :trust_sharepoint_link, presence: true, url: {hostnames: SHAREPOINT_URLS}, on: :create
 
-  validate :establishment_exists, :trust_exists, on: :create
+  validate :establishment_exists, :trust_exists, on: :create, if: proc { |project| project.errors.empty? }
```
mec, line 24: Is there a reason we don't make this about the values themselves rather than errors? Something like:

```
  validate :trust_exists, on: :create, unless: proc { |project| project.incoming_trust_ukprn.nil? }
  validate :establishment_exists, on: :create, unless: proc { |project| project.urn.nil? }
```

mec (changes requested):
I've got a question about the approach that would probably change the way we test this, let me kbnow your thoughts.

app/models/project.rb
jkempster34: Done 👍

lozette (approved):
Approved but you have a speeling mistake

spec/models/project_spec.rb
```diff
@@ -102,14 +104,16 @@
     end
 
     describe "#incoming_trust_ukprn" do
-      context "when no trust with that UKPRN exists in the API" do
+      context "when no trust with that UKPRN exists in the API and the UKPRN is prsent" do
```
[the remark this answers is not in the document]
jkempster34: Whoops, thanks!

PR #383, opened 2022-12-13
Use FormObject to create projects

lozette:
## Changes

Now that we have a FormObject pattern to create Involuntary & Voluntary conversion projects, change the form in the UI to submit the Project Create form to the FormObject. This will then handle creating the project, creating a note if one is required, and assigning the correct Tasklist to the project. 

## Checklist


mec (changes requested):
I left a couple of comments - I do think if we move the notification method to the form object it becomes simpler to test, might be worth a go?

I would also like to see us flatten out the form view now as we don't need to rely on `accepts_nested_attributes_for` in `Project` with the form object - I can help try that if you like?

PR #385, opened 2022-12-13
898 copy changes to external stakeholder kick off task

titlescreen:
## Changes

- Send introductory emails iterations 
- Proforma action is to check it is saved not check doc for errors
- External stakeholder hint prompts to talk about common problems
- Frame LA guidance around what if you do not have the proforma
- Check LA date hint becomes the action as it is more explicit

## Checklist

### DIFF
--- a/app/workflows/lists/conversion/involuntary/sections/kickoff.yml
+++ b/app/workflows/lists/conversion/involuntary/sections/kickoff.yml
@@ -69,10 +69,7 @@ tasks:
           The regional delivery officer should have sent the local authority proforma to the local authority to complete.
 
           If you haven't received it yet, check that the regional delivery officer sent it and then ask the local authority if they have completed and shared it.
-      - title: Confirm the provisional conversion date with the local authority
-        hint:
-          Check that the local authority has capacity to convert the school by
-          the provisional date.
+      - title: Check that the local authority has capacity to convert the school by the provisional date
         guidance_summary:
           What to do if the local authority does not have capacity
         guidance_text: |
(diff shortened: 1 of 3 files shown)


app/workflows/lists/conversion/involuntary/sections/kickoff.yml
```diff
@@ -40,7 +40,8 @@ tasks:
     title: External stakeholder kick-off
     hint:
       You should start this work after you've had the handover with the regional
-      delivery officer.
+      delivery officer. This is a good opportunity for you to talk to about
```
beeLorna, line 43: Is there an extra "to" in this sentence that shouldn't be there? I've crossed out the one that I feel is there in error.

> This is a good opportunity for you to talk ~~to~~ about common problems and how to avoid them.

or is a "them" needed to be added in, like:

> This is a good opportunity for you to talk to them about common problems and how to avoid them.

app/workflows/lists/conversion/voluntary/sections/kickoff.yml
```diff
@@ -38,7 +38,8 @@ tasks:
     title: External stakeholder kick-off
     hint:
       You should start this work after you've had the handover with the regional
-      delivery officer.
+      delivery officer. This is a good opportunity for you to talk to about
```
beeLorna, line 41: See comment for this same line within the involuntary kickoff.yml - same applies here too

beeLorna (approved):
There is just a comment on the structure of one of the sentences, but otherwise all looks good.
You'll also need to do a rebase against `main` now as well :)

app/workflows/lists/conversion/involuntary/sections/kickoff.yml
titlescreen: A nice catch, I will remove the extra to as I think having 'This is a good opportunity to talk to them about…' is adding happy words that do not add to the understanding of the sentence.

PR #386, opened 2022-12-13
Add `TaskList::Base` class and basic `Conversion::Voluntary::TaskList`

jkempster34:
### Add `TaskList::Base` class and basic `Conversion::Voluntary::TaskList`
We want an easy way to create different types of task list by defining a
task list layout composed of `TaskList::Task` objects.

The `TaskList::Base` class handles the mapping of the `TaskList::Task` objects
to the task list database table. This uses the `TaskList::Task` key/identifier,
which must be named consistently with the database columns. For example, a task
list containing a `TaskList::Task` named
`Conversion::Voluntary::Tasks::Handover` which has a `review` attribute, must
have a column named `handover_review`.

For now, the base class is tested through an instance of a
`Conversion::Voluntary::TaskList`. As we introduce more task list types, this
will likely be moved into a shared spec.

### Rename `TaskList::Task.key` to `identifier`
Rename the `key` method of the `Task` class so that:
- It's more consistent with the `Section` model.
- It's not confused with a key for a Ruby Hash.

### Remove `TaskList::Task.humanized_name`
We plan to use locales to store the task name instead.

## Checklist

### DIFF
--- a/app/models/conversion/voluntary/task_list.rb
+++ b/app/models/conversion/voluntary/task_list.rb
@@ -1,0 +1,16 @@
+class Conversion::Voluntary::TaskList < TaskList::Base
+  self.table_name = "conversion_voluntary_task_lists"
+
+  TASK_LIST_LAYOUT = [
+    {
+      identifier: :project_kick_off,
+      tasks: [
+        Conversion::Voluntary::Tasks::Handover
+      ]
+    }
+  ].freeze
+
+  def task_list_layout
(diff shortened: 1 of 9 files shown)


spec/models/task_list/section_spec.rb
```diff
@@ -0,0 +1,4 @@
+require "rails_helper"
```
mec, line 1: I know I added this - but we should drop it!

PR #387, opened 2022-12-13
Enable DNS zone

Stretch96:
* Allows creating a DNS zone if required

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -78,3 +78,13 @@ variable "enable_cdn_frontdoor" {
   description = "Enable Azure CDN FrontDoor. This will use the Container Apps endpoint as the origin."
   type        = bool
 }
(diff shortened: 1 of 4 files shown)


terraform/variables.tf
```diff
@@ -78,3 +78,13 @@ variable "enable_cdn_frontdoor" {
...
   type        = bool
 }
+
+variable "enable_dns_zone" {
+  description = "Conditionally create a DNS zone"
+  type        = bool
+}
+
+variable "dns_zone_domain_name" {
+  description = "DNS zone domain name. If created, records will automatically be created to point to the CDN."
```
DrizzlyOwl, line 88: Can we swap "If created" for "If specified"?

DrizzlyOwl (approved):

PR #389, opened 2022-12-13
(934) Send notification to caseworker when assigned to a project

beeLorna:
## Changes

When a caseworker is assigned to a project, they will receive an email notification informing them of the assignment along with a link to the project.
The email template has been set up in `Notify`

Currently an email is sent when a reassignment is made, even if it that reassignment is to the same caseworker that is already the current caseworker. I don't necessarily think this is a problem, but it's worth noting. 

## Checklist

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -31,6 +31,8 @@ class AssignmentsController < ApplicationController
     @project.update(caseworker_assigned_at: DateTime.now) if @project.caseworker_assigned_at.nil?
     @project.update(caseworker_params)
 
+    CaseworkerMailer.caseworker_assigned_notification(@project.caseworker, @project).deliver_later
+
     redirect_to project_information_path(@project), notice: t("project.assign.caseworker.success")
   end
 
--- a/spec/requests/assignments_controller_spec.rb
+++ b/spec/requests/assignments_controller_spec.rb
@@ -145,5 +145,13 @@ RSpec.describe AssignmentsController, type: :request do
       expect(project.reload.caseworker).to eq caseworker
       expect(project.reload.caseworker_assigned_at).to eq DateTime.now
     end
+
+    it "sends a notification to the caseworker" do
+      perform_request
+
+      expect(ActionMailer::MailDeliveryJob)
+        .to(have_been_enqueued.on_queue("default")
+        .with("CaseworkerMailer", "caseworker_assigned_notification", "deliver_now", args: [caseworker, Project.last]))
+    end
   end
 end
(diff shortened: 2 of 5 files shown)


app/mailers/caseworker_assignment_mailer.rb
```diff
@@ -0,0 +1,12 @@
+class CaseworkerAssignmentMailer < ApplicationMailer
```
jkempster34, line 1: Hiya, not a proper review because I've only glanced through!

I think this class would be best named `CaseworkerMailer` based on the guidance here: https://katanacode.com/blog/ruby-on-rails-action-mailer-how-to-name-mailer-methods/

We could also follow their convention of naming the methods and update ours to be:
- `TeamLeaderMailer#project_created_notification`
- `CaseworkerMailer#caseworker_assigned_notification`

Let me know what you think!

app/controllers/assignments_controller.rb
```diff
@@ -31,6 +31,9 @@ def update_caseworker
     @project.update(caseworker_assigned_at: DateTime.now) if @project.caseworker_assigned_at.nil?
     @project.update(caseworker_params)
 
+    caseworker = User.find(caseworker_params["caseworker_id"])
+    CaseworkerAssignmentMailer.new_assigned_caseworker(caseworker, @project).deliver_later
```
jkempster34, line 35: I think we can save the extra request here and do `CaseworkerAssignmentMailer.new_assigned_caseworker(@project.caseworker, @project).deliver_later`

spec/requests/assignments_controller_spec.rb
```diff
@@ -145,5 +145,13 @@
       expect(project.reload.caseworker).to eq caseworker
       expect(project.reload.caseworker_assigned_at).to eq DateTime.now
     end
+
+    it "sends a notification to the caseworker" do
+      perform_request
+
+      expect(ActionMailer::MailDeliveryJob)
+        .to(have_been_enqueued.on_queue("default")
+      .with("CaseworkerAssignmentMailer", "new_assigned_caseworker", "deliver_now", args: [caseworker, Project.last]))
```
jkempster34, line 154: Some indentation issues here

jkempster34 (changes requested):

jkempster34 (changes requested):
Just tested this out and it works great.

Left a few comments where things could be tidied up a touch!

spec/requests/assignments_controller_spec.rb
jkempster34: This is looking good to me, apart from this! ☝️

beeLorna: There isn't actually an indentation problem here - it's just github displaying it in a peculiar way.

jkempster34: Same on my IDE. 6 spaces on the first line, then 8 on the next

image

jkempster34 (approved):
🙌

PR #396, opened 2022-12-15
Correctly validate incomplete, invalid or missing multipart dates

lozette:
## Changes

When entering dates in the create project form, we want to be able to validate them in a meaningful way. Previously missing or incomplete dates were throwing MultipartAssignment errors on the front end (translating to a 500 error and a bad experience for the user).

This PR checks for:
- Blank dates
- Partially complete dates (e.g. missing a month or year)
- Invalid dates (like 31st February)
- "Dates" made up of strings of letters

All but the first of these result in the user being told the date entered is invalid. The first renders an error which tells the user the date is blank.  

## Checklist

### DIFF
--- a/app/forms/conversion/create_project_form.rb
+++ b/app/forms/conversion/create_project_form.rb
@@ -35,13 +35,13 @@ class Conversion::CreateProjectForm
 
   def provisional_conversion_date=(hash)
     @provisional_conversion_date = Date.new(year_for(hash), month_for(hash), day_for(hash))
-  rescue NoMethodError
+  rescue TypeError, NoMethodError
     nil
   end
 
   def advisory_board_date=(hash)
     @advisory_board_date = Date.new(year_for(hash), month_for(hash), day_for(hash))
-  rescue NoMethodError
+  rescue TypeError, NoMethodError
     nil
   end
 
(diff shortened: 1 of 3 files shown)


lozette:
[Fix multiparameter date submission](https://trello.com/c/GAI7lOH0/768-fix-multiparameter-date-submission)

app/forms/conversion/create_project_form.rb
```diff
@@ -41,31 +43,25 @@ def initialize(params = {})
...
-    @advisory_board_date = Date.new(year_for(hash), month_for(hash), day_for(hash))
+    @advisory_board_date = Date.new(value_at_position(hash, 1), value_at_position(hash, 2), value_at_position(hash, 3))
   rescue NoMethodError
     nil
-  rescue TypeError, Date::Error
+  rescue TypeError, Date::Error, NegativeValueError
     @attributes_with_invalid_values << :advisory_board_date
   end
 
-  private def day_for(value)
```
mec, line 48: Just interested in why you got rid of these?

mec (approved):
Phew! This all looks reasonable - why we have to go through it all is still nagging at me, but this is an excellent start to fixing the issue for our users! :star

app/forms/conversion/create_project_form.rb
lozette: Just to reduce duplication - I'd have to throw the NegativeDateError from each one and it felt better to roll them into one method. Although it's probably less readable now?

PR #399, opened 2022-12-16
Fix: Resubmitting a create conversion project form

lozette:
## Changes

We had a bug in which resubmitting a create conversion project form, after it had failed validation, raised a 500 error.

To reproduce:
- Fill in either an involuntary or voluntary create project form, making sure the form will fail validation
- Click submit
- Correct any validation errors
- Click submit

The app will then render a 500 error.

This is because the Conversion::Voluntary/Involuntary::ProjectsController create method is expecting params with a key of `conversion_project`. When the form is first submitted, the principal parameter key will be `conversion_project` because the form is initialised with a Conversion::Project object.

However, once the form fails validation and re-renders, the form will be initialised with a Conversion::Involuntary::CreateProjectForm or a Conversion::Voluntary::CreateProjectForm, and the parameter key on re-submission will be either `:conversion_involuntary_create_project_form` or `:conversion_voluntary_create_project_form`

To remedy this we have:
- Moved the `new` and `project_params` methods into the individual controllers. This means the forms are initialised with either a `Conversion::Voluntary::CreateProjectForm` or
`Conversion::Involuntary::CreateProjectForm` depending on which project is being created, and `project_params` can then expect the correct key

- Explicitly added the text for the form labels. Previously these were being added via Rails/I18n magic and assumed the key would contain `conversion_project` because that was the key to the form. As the form is now named either `conversion_involuntary_create_project_form` or
`conversion_voluntary_create_project_form` we needed to either change the keys (and have two sets of identical keys for each form type) or explicity add the labels. We have gone for the latter option to  reduce duplication.

## Checklist


lozette:
[Bug: Resubmitting an invalid conversion project form results in an error](https://trello.com/c/ACvdYAlw/964-bug-resubmitting-an-invalid-conversion-project-form-results-in-an-error)

mec (approved):
🚢

PR #408, opened 2022-12-19
Refactor GitHub workflow outputs

Stretch96:
* The `set-output` command is depreciated. This refactors the workflows to output them into the `GITHUB_OUTPUT` environment variable instead.
* ```Warning: The \`set-output\` command is deprecated and will be disabled soon.```

### DIFF
--- a/.github/workflows/build-and-push-image-test.yml
+++ b/.github/workflows/build-and-push-image-test.yml
@@ -13,7 +13,7 @@ jobs:
       version: ${{steps.release-version.outputs.version}}
     steps:
       - id: release-version
-        run: echo "::set-output name=version::release-9"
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-push-image-test.yml
```diff
@@ -44,8 +44,8 @@ jobs:
             VERSION=sha-${GITHUB_SHA}
             TAGS="$TAGS,${DOCKER_IMAGE}:${VERSION},${DOCKER_IMAGE}:${{needs.set-test-release-version.outputs.version}}"
```
mec, line 45: Is there no change when reading the values later? We still only need `outputs`?

mec (commented):
I have a quesiton about using the output.

DrizzlyOwl (approved):

PR #411, opened 2022-12-20
919 tenacy at will exceptional circumstances guidance

SteveMilnes:
## Changes

Updates to Tenancy at will guidance in voluntary and involuntary task list.

Content changes that explain the document is and should be rarely used. Links to form to contact policy team for advice if a Tenancy at will is needed.

https://trello.com/c/coz7WqIs/919-tenacy-at-will-can-only-be-used-in-exceptional-circumstances

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Confirm opening date with school becomes Share the grant certificate and
   information about opening with the trust
 - Tasks Continue button now states 'Save and return', tests updated too
+- Tenancy at will guidance now explains how rarely it is used and that users
+  should contact policy team for advice if one is needed. Voluntary &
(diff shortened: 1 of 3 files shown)


SteveMilnes:
[Tenacy at will can only be used in exceptional circumstances](https://trello.com/c/coz7WqIs/919-tenacy-at-will-can-only-be-used-in-exceptional-circumstances)

jkempster34 (approved):
@SteveMilnes Looks good to me! 🚢 

I've tidied up the commits a bit by squashing them into one (as we discussed this is quite difficult, so don't worry about it too much, but if we can deliver related changes in the same commit that would be best. Though it's important to try to keep it to one change per commit...)

I've also changed the commit title from "Added" to "Add". We want a short title in the "imperative form". Again not too important for this, but nice to try to stick to! As you can see below when they are all consistent, they read nicely together

image


Here is what it looks like in the app 
image

PR #412, opened 2022-12-20
Change validations on `Project` and `CreateProjectForm`

lozette:
## Changes

Now that we have a FormObject (CreateProjectForm) for creating projects, we no longer create them with the Project model directly. This means we can change some of the validations. We have:

- Loosened the `provisional_conversion_date` validation. It now only checks the date is in the future on creation (via the form). In the project model, the date only needs to exist, as the project should still be valid even if the conversion date has passed.
- Changed the validation of `advisory_board_date` - it should always be in the past, even if the project is being edited, so we should not check it only on create.
- The sharepoint links should always be validated, not just on create.
- Added the `trust_exists` and `establishment_exists` validations to the FormObject, as these need to be checked on project creation via the form.

TODO: Do we need to remove `trust_exists` and `establishment_exists` validations from the Project model? Are these attributes ever likely to be edited?

## Checklist

### DIFF
--- a/app/forms/conversion/create_project_form.rb
+++ b/app/forms/conversion/create_project_form.rb
@@ -1,41 +1,41 @@
 class Conversion::CreateProjectForm
   include ActiveModel::Model
+  include ActiveModel::Attributes
   include ActiveRecord::AttributeAssignment
 
   SHAREPOINT_URLS = %w[educationgovuk-my.sharepoint.com educationgovuk.sharepoint.com].freeze
 
   class NegativeValueError < StandardError; end
 
-  attr_accessor :urn,
-    :incoming_trust_ukprn,
-    :establishment_sharepoint_link,
-    :trust_sharepoint_link,
-    :advisory_board_conditions,
-    :note_body,
-    :user
+  attribute :urn, :integer
+  attribute :incoming_trust_ukprn, :integer
+
+  attr_accessor :establishment_sharepoint_link,
+  :trust_sharepoint_link,
+  :advisory_board_conditions,
+  :note_body,
+  :user
 
   attr_reader :provisional_conversion_date,
     :advisory_board_date
 
-  validates :urn,
-    :incoming_trust_ukprn,
-    :establishment_sharepoint_link,
-    :trust_sharepoint_link,
-    :provisional_conversion_date,
+  validates :provisional_conversion_date,
     :advisory_board_date,
     presence: true
 
-  validates :urn, urn: true
-  validates :incoming_trust_ukprn, ukprn: true
-
-  validates :provisional_conversion_date, date_in_the_future: true
-  validates :provisional_conversion_date, first_day_of_month: true
+  validates :provisional_conversion_date, date_in_the_future: true, first_day_of_month: true
   validates :advisory_board_date, date_in_the_past: true
 
   validates :establishment_sharepoint_link, presence: true, url: {hostnames: SHAREPOINT_URLS}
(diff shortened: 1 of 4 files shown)


lozette:
[Loosen Project validations](https://trello.com/c/a6sRLUgX/967-loosen-project-validations)

app/forms/conversion/create_project_form.rb
```diff
@@ -17,26 +17,25 @@ class NegativeValueError < StandardError; end
...
-    :provisional_conversion_date,
+  validates :provisional_conversion_date,
     :advisory_board_date,
     presence: true
 
-  validates :urn, urn: true
-  validates :incoming_trust_ukprn, ukprn: true
-
   validates :provisional_conversion_date, date_in_the_future: true
   validates :provisional_conversion_date, first_day_of_month: true
```
jkempster34, line 25: I wonder if these two could be squashed onto one line to match the rest of the validations

`validates :provisional_conversion_date, date_in_the_future: true, first_day_of_month: true`

spec/features/users_can_create_involuntary_conversion_projects_spec.rb
```diff
@@ -13,7 +13,7 @@
     let(:ukprn) { 10061021 }
     let(:two_weeks_ago) { Date.today - 2.weeks }
 
-    before { mock_successful_api_responses(urn: urn, ukprn: ukprn) }
+    before { mock_successful_api_responses(urn: any_args, ukprn: any_args) }
```
jkempster34, line 16: Just had a poke at this and it looks like URN and UKPRN are no longer cast to integers in the model, but we still need to set up the mock for the factory as well.

I think one option is to change from using attr_accessor to using the `attribute` method of `ActiveModel::Attributes` so that we can cast values, and have:

```ruby
include ActiveModel::Attributes

attribute :urn, :integer
```

If not, I think this will have to do!

jkempster34 (approved):
I found browsing through this a bit difficult with the form model hierarchy and the ActiveRecord model too, I think thats the nature of the pattern though.

Are the validations meant to be duplicated across the `Project` model and the form models? I suppose we need the contextual validations in the form models, but also the security of the validations on the `Project` model itself... 🤔 

I do reckon however, that we should probably remove `trust_exists` and `establishment_exists` from `Project` because we don't update UKPRN and URN currently anyway.

Left a couple of comments of potential tidying up, but nothing serious!

spec/features/users_can_create_involuntary_conversion_projects_spec.rb
lozette: I tried this when doing the work originally (I saw you'd used it in a closed PR) but I am a bit baffled by it, I can't figure out why I get this error: 

```Failure/Error: attribute :urn, :integer

NameError:
  undefined local variable or method `attribute_aliases' for Conversion::CreateProjectForm:Class

          name = attribute_aliases[name] || name
                 ^^^^^^^^^^^^^^^^^
  Did you mean?  attribute
```

I don't know if it's so important to figure this out now or not?

jkempster34: It should just need the mixins:

  include ActiveModel::Model
  include ActiveModel::Attributes
  
Other than that, I'm not sure!

lozette: FTR this was me being a new year, first day back muppet and typing `include ActiveRecord::Attributes` 🙄

PR #416, opened 2022-12-22
Use locales and ViewComponents to render parts of the TaskList

lozette:
## Changes

This PR has two main parts:

1. Use locales to render the text for tasks, instead of the workflow YAML files.
2. Use ViewComponents to render the TaskLists, instead of partials.

Using ViewComponents should hopefully result in a less complex way of building the new TaskLists, instead of nested partials which can get very complicated quickly. 

In order to use ViewComponents, we needed the Tasks to be aware of their locales, and therefore tie a task to a particular locale file for its content. We can then use the locale files to output the required text for the Task.

This PR also adds templates for the ViewComponents, and relevant styling; as well as unit tests for the ViewComponents.

## Checklist

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -8,6 +8,7 @@ $govuk-global-styles: true;
 @import "accessible-autocomplete/src/autocomplete";
 
 @import "components/task-list";
+@import "components/task_list/check-box-action-component";
 
 // The accessible autocomplete component does not use
 // GOV.UK fonts by default for suggestions.
--- a/app/components/task_list/check_box_action_component.rb
+++ b/app/components/task_list/check_box_action_component.rb
@@ -1,0 +1,17 @@
+class TaskList::CheckBoxActionComponent < ViewComponent::Base
+  def initialize(task:, attribute:)
+    locales_path = task.locales_path
+
+    @task = task
+    @attribute = attribute
+    @label = t("#{locales_path}.#{attribute}.title")
+    @hint = t("#{locales_path}.#{attribute}.hint.html", default: nil)
+    @guidance_link = t("#{locales_path}.#{attribute}.guidance_link", default: nil)
+    @guidance = t("#{locales_path}.#{attribute}.guidance.html", default: nil)
+    @action_id = "#{task.class.name.parameterize(separator: "_")}[#{attribute}]"
+  end
+
+  def check_box_value
+    @task.send(@attribute)
+  end
+end
--- a/app/helpers/task_list_helper.rb
+++ b/app/helpers/task_list_helper.rb
@@ -3,7 +3,12 @@ module TaskListHelper
     "#{task.title.parameterize}-status"
   end
 
-  def task_status_tag(task)
+  def task_id(task)
+    task_title = t("#{task.locales_path}.title").parameterize
+    "#{task_title}-status"
+  end
+
+  def task_status_tag(task, task_status_id)
     # Colours as per https://govuk-components.netlify.app/components/tag/
     tags_for_states = {
       not_started: {
@@ -34,7 +39,7 @@ module TaskListHelper
       text: task_state[:text],
       colour: task_state[:colour],
       classes: "app-task-list__tag",
-      html_attributes: {id: task_status_id(task)}
+      html_attributes: {id: task_status_id}
     )
   end
 end
--- a/app/models/task_list/section.rb
+++ b/app/models/task_list/section.rb
@@ -7,10 +7,6 @@ class TaskList::Section
     @tasks = tasks
   end
 
-  def title
-    I18n.t("task_list.sections.#{identifier}.title")
-  end
-
   def locales_path
     "#{@locales_path}.#{identifier}"
   end
--- a/app/models/task_list/section.rb
+++ b/app/models/task_list/section.rb
@@ -1,12 +1,17 @@
 class TaskList::Section
   attr_accessor :identifier, :tasks
 
-  def initialize(identifier:, tasks:)
+  def initialize(identifier:, tasks:, locales_path:)
+    @locales_path = locales_path
     @identifier = identifier
     @tasks = tasks
   end
 
   def title
     I18n.t("task_list.sections.#{identifier}.title")
+  end
+
+  def locales_path
+    "#{@locales_path}.#{identifier}"
   end
 end
--- a/app/assets/stylesheets/components/task_list/check-box-action-component.scss
+++ b/app/assets/stylesheets/components/task_list/check-box-action-component.scss
@@ -1,0 +1,21 @@
+.govuk-checkboxes__item {
+  label {
+    @extend .govuk-\!-font-weight-bold;
+  }
+
+  .hint {
+    display: block;
+    padding-right: govuk-spacing(3);
+    padding-left: govuk-spacing(3);
+
+    > p {
+      @extend .govuk-hint;
+    }
+  }
+
+  .guidance {
+    display: block;
+    padding-right: govuk-spacing(3);
+    padding-left: govuk-spacing(3);
+  }
+}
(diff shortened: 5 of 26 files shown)


lozette:
[Figure out a way to store the section title and task title in a translations file per task list](https://trello.com/c/TSTBMCDM/969-figure-out-a-way-to-store-the-section-title-and-task-title-in-a-translations-file)

app/assets/stylesheets/application.scss
```diff
@@ -42,6 +42,21 @@ $govuk-global-styles: true;
     label {
       @extend .govuk-\!-font-weight-bold;
     }
+    div.hint {
```
jkempster34, line 45: Do we need the `div` selectors here, and can we nest the `p` selector? Not tested, but something like:

```scss
  .govuk-checkboxes__item {
    label {
      @extend .govuk-\!-font-weight-bold;
    }

    .guidance {
       display: block;
       padding-right: govuk-spacing(3);
       padding-left: govuk-spacing(3);
    }
    
    .hint {
      display: block;
      padding-right: govuk-spacing(3);
      padding-left: govuk-spacing(3);

      > p {
        @extend .govuk-hint;
      }
    }
```

app/models/task_list/section.rb
```diff
@@ -1,12 +1,17 @@
...
   attr_accessor :identifier, :tasks
 
-  def initialize(identifier:, tasks:)
+  def initialize(identifier:, tasks:, locales_path:)
+    @locales_path = locales_path
     @identifier = identifier
     @tasks = tasks
   end
 
   def title
```
jkempster34, line 10: I think this `title` method can go now, along the `title` method in `TaskList::Task` and most of the `task_list.en.yml`, they were part of the previous implementation.

app/components/task_list/task_header.rb
```diff
@@ -0,0 +1,34 @@
...
+  def initialize(project:, task:)
+    @locales_path = task.locales_path
+    @title = title
+    @establishment_name = project.establishment.name
+    @hint = hint
+    @guidance_link = guidance_link
+    @guidance = guidance
+  end
+
+  private def title
```
jkempster34, line 11: Can these just be public methods rather than assigning their values to instance variables?

app/components/task_list/task_header.rb
```diff
@@ -0,0 +1,34 @@
...
+    t("#{@locales_path}.title")
+  end
+
+  private def hint
+    hint = t("#{@locales_path}.hint.html", default: nil)
+    return if hint.nil?
```
jkempster34, line 17: I don't think this is needed, won't it just return nil anyway?

app/helpers/task_list_helper.rb
```diff
@@ -3,6 +3,11 @@ def task_status_id(task)
     "#{task.title.parameterize}-status"
   end
 
+  def task_id(task)
```
jkempster34, line 6: There is a `task_status_id` method in this helper, will that go when the old task list goes?

app/components/task_list/task_header.rb
jkempster34: Same for the rest of the methods in this class

jkempster34 (commented):
Just left some quick initial thoughts! I haven't pulled this down or anything, so not a proper review.

app/helpers/task_list_helper.rb
mec: Yep, the why should be in the commit message.

jkempster34: Ah, yes. So it is!

app/components/task_list/task_header.rb
mec: I don't think so? View components work like views, they need to be instance variables or helper methods?

app/models/task_list/section.rb
mec: I thought I remembered trying that... I'll have  a look.

app/assets/stylesheets/application.scss
mec: We don't need the divs - but one could argue it makes them more specific?

Happy to change it to the suggestion.

app/components/task_list/task_header.rb
jkempster34: I'll pull this down and check, I'm half sure they can

app/assets/stylesheets/application.scss
jkempster34: I really like how the nesting in Sass reflects the HTML structure

app/components/task_list/task_header.rb
mec: The section one looks like it can.

This helper `task_status_tag(task)` looks like it is used for both new and old as well so needs work.

mec: SImplest thing to do right now is to keep `title` on `TaskList::Task` and remove it later?

app/models/task_list/section.rb
mec: The section one looks like it can.

This helper task_status_tag(task) looks like it is used for both new and old as well so needs work.

mec: SImplest thing to do right now is to keep title on TaskList::Task and remove it later?

app/components/task_list/task_header.rb
jkempster34: I think we should probably rework the `task_status_tag(task)` method. The new implementation is relying on the old one being there.

`task_status_tag` calls `task_status_id` which is the one that requires a name

spec/components/check_box_action_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    let(:attribute) { "notes" }
+
+    before do
+      allow(task).to receive_message_chain("class.name").and_return("Conversion::Voluntary::Tasks::Handover")
+    end
+
+    it "renders the correct `name` for the form input" do
+      render_inline(TaskList::CheckBoxAction.new(task: task, attribute: attribute))
+
+      expect(page.find("//input[1]/@name").text).to eq("conversion_voluntary_tasks_handover[notes]")
```
jkempster34, line 53: This xpath could go in place of

`expect(page.find_field("conversion_voluntary_tasks_handover[notes]")).to be_present`

spec/components/check_box_action_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  before do
+    allow(task).to receive(attribute.to_sym).and_return(true)
+  end
+
+  context "when the checkbox only has a title" do
+    let(:attribute) { "notes" }
+
+    it "renders the title" do
+      render_inline(TaskList::CheckBoxAction.new(task: task, attribute: attribute))
```
jkempster34, line 14: This `render_inline` could be placed in the before block and removed from each test, but that's probably just personal preference.

app/components/task_list/check_box_action.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class="govuk-checkboxes__item">
+  <%= hidden_field_tag(@name, "0", {id: nil}) %>
+  <%= check_box_tag(@name, 1, @task.send(@attribute), {class: "govuk-checkboxes__input"}) %>
```
jkempster34, line 3: It's a shame that we can no longer use the Gov form builder!

mec: We can still use the builder on the forms, the intention here is that the component makes composing a common task view simple. 

I never liked the component calling the builder. I guess the longer solution is to extned the builder for our version of a check box action, but that felt like a bigger piece of work.

app/models/task_list/section.rb
mec: As you proposed, we've changed the `task_status_tag` method to take the appropriate id as a temporary solution.

app/components/task_list/check_box_action_component.rb
```diff
@@ -0,0 +1,19 @@
+class TaskList::CheckBoxActionComponent < ViewComponent::Base
+  attr_reader :attribute, :locales_path, :task
```
jkempster34, line 2: I don't think these are used, I think you might have copied them over when checking if methods are exposed to the view when using view components

jkempster34 (approved):
LGTM! Pulled it down and ran through everything, and all works as expected

Really cool that we can keep the same pattern of having the content in the locales!

Left a comment about some `attr_readers` that can be dropped, and also the section and task titles can be dropped from `config/locales/task_list.en.yml` as well!

PR #419, opened 2022-12-23
Fix the CheckBoxActionComponent name creation

mec:
## Changes

Previously we used `parameterize` which would concatonate a multi-word
class name like `StakeholderKickOff` to `stakeholderkickoff` which is
not what our naming conventions expect.

Switching to `underscore` and swapping the `/` for `_` solves this
issue, resulting in `stakeholder_kick_off`.

This issue hightlights:

- how important it is to get the naming right
- that we should use a custom form builder to make this less error prone


jkempster34 (approved):
✔️

PR #425, opened 2023-01-03
(985) Content Security Policy

jkempster34:
> I've had a good browse around. Everything seems to work as expected, and I don't see any violations.

### Enable Content Security Policy
Our recent IT Health Check raised the lack of a CSP as a medium warning for the
application.

This enables the Rails CSP with minimal rules.

For now, we won't report any CSP violations via the `report-uri`, however, this
may be revisited later.

### Add nonce to inline scripts so that are not blocked by the CSP
We have added a CSP to the application. 

This is blocking our inline scripts for:
- Enabling the GOV.UK JavaScript
- Enabling the Accessible Autocomplete component.

This adds nonces to the inline scripts. Because a nonce generator is configured
in `content_security_policy.rb`, we can simply call
`content_security_policy_nonce`.

Note: It's nicer to add nonces to a `javascript_tag` (for example `<%=
javascript_tag nonce: true do %>`). However, the ERB linter suggests avoiding
`javascript_tag` altogether.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 #### Added
 
 - Send notification to caseworker when assigned to a project
+- Add a Content Security Policy
 
 #### Changed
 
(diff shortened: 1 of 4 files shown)


jkempster34:
[Enable Content security policy](https://trello.com/c/RI5yiUL8/985-enable-content-security-policy)

lozette (approved):

PR #431, opened 2023-01-04
(1002) Not applicable task

jkempster34:
### Add an optional task model
We have a bunch of tasks which can be optional/not applicable.

Add a derived class of TaskList::Task which computes the status of a optional
task.

### Add subleases task to involuntary task list
This adds an "optional" Sublease task for the involuntary conversion task
list.

The checkboxes are wrapped in a `govuk_check_boxes_fieldset` from the form
builder so that we can display errors correctly.

### Add validation when both `not_applicable` and other actions are selected
Users should not be able to select both "not applicable" and any other
actions.

Add validation to the optional task model, derived models must define an error
key and message.

It's important to note that the attributes of an optional task must be cast
(via the attribute method on the task model) for the validation and subsequent
re-rendering of the page, else they get assigned as "0" and "1" (which are both
true), and we have to handle the cast manually.

## Screenshots
image

NOTE: Error message should sit above main heading

## Checklist

### DIFF
--- a/app/models/conversion/involuntary/tasks/subleases.rb
+++ b/app/models/conversion/involuntary/tasks/subleases.rb
@@ -1,0 +1,10 @@
+class Conversion::Involuntary::Tasks::Subleases < TaskList::OptionalTask
+  attribute :received, :boolean
+  attribute :cleared, :boolean
+  attribute :signed, :boolean
+  attribute :saved, :boolean
+
+  attribute :email_signed, :boolean
+  attribute :receive_signed, :boolean
+  attribute :save_signed, :boolean
+end
--- a/app/models/conversion/involuntary/tasks/subleases.rb
+++ b/app/models/conversion/involuntary/tasks/subleases.rb
@@ -7,4 +7,12 @@ class Conversion::Involuntary::Tasks::Subleases < TaskList::OptionalTask
   attribute :email_signed, :boolean
   attribute :receive_signed, :boolean
   attribute :save_signed, :boolean
+
+  private def not_applicable_only_key
+    :actions
+  end
+
+  private def not_applicable_only_message
+    I18n.t("errors.conversion_involuntary_tasks_subleases.actions.invalid")
+  end
 end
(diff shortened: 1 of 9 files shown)


app/models/task_list/optional_task.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  private def completed?
+    actions_when_applicable.values.all?(&:present?)
+  end
+
+  private def actions_when_applicable
+    attributes.except("not_applicable")
+  end
+
+  private def not_applicable_only
```
jkempster34, line 18: not sure how i feel about defining this in the base class 🤔

jkempster34:
[Tasks can be not applicable](https://trello.com/c/6Z0n8dmo/1002-tasks-can-be-not-applicable)

app/views/conversion/involuntary/task_lists/tasks/subleases.html.erb
```diff
@@ -0,0 +1,34 @@
+<% content_for :pre_content_nav do %>
+  <% render partial: "shared/back_link", locals: {href: conversion_involuntary_tasks_path(@project)} %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @task, url: conversion_involuntary_update_task_path(@project, @task.class.identifier), method: :put do |form| %>
+      <%= form.govuk_error_summary %>
```
jkempster34, line 8: I've moved this above the title.

We might want to either:
- Update the other tasks to match?
- Update this task to match?

app/models/conversion/involuntary/tasks/subleases.rb
```diff
@@ -0,0 +1,18 @@
+class Conversion::Involuntary::Tasks::Subleases < TaskList::OptionalTask
+  attribute :received, :boolean
```
jkempster34, line 2: These have to be cast for the validation and reloading of the page, else they get assigned as "0" and "1" (which are both true), and we have to do a bunch of extra work.

jkempster34:
Closing as we don't want to handle validation yet.

PR #432, opened 2023-01-05
Fix new task layouts

mec:
## Changes

The new tasklist has slightly different markup to the old.

Here we tweak the markup and styles to bring them closer visually, it's not
perfect but I am keen not to labour over this too much right now - the value is
us shipping the new task list.

image

image

https://trello.com/c/CfJsfWoo

### DIFF
--- a/app/assets/stylesheets/components/task_list/task-header-component.scss
+++ b/app/assets/stylesheets/components/task_list/task-header-component.scss
@@ -1,0 +1,7 @@
+.app-task-hint {
+  @extend .govuk-hint;
+
+  p {
+    @extend .govuk-hint;
+  }
+}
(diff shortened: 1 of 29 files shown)


app/assets/stylesheets/components/task_list/task-header-component.scss
```diff
@@ -0,0 +1,7 @@
+.app-task-hint {
+  @extend .govuk-hint;
```
jkempster34, line 2: Just ran this and it looks like this line isn't needed. Otherwise, everything looks great

jkempster34 (changes requested):
Just one tiny thing

app/assets/stylesheets/components/task_list/task-header-component.scss
mec: I put this in to cover if you forgot to add the `<p>` in the locales. - I should have mentioned this in the commit - I know it's ugly, but I am confident we will be back to sort out other visual issues once we have shipped public beta!

jkempster34 (approved):

PR #433, opened 2023-01-05
Force GOVUK styles in hints and guidance

mec:
## Changes
Content for hints and guidance is in the locales files and we want to
minimise the markup developers have to remember to add there.

We can force any content to have GOVUK styles by targeting elements,
this means no styles are required in the locales.

### DIFF
--- a/app/assets/stylesheets/components/task_list/check-box-action-component.scss
+++ b/app/assets/stylesheets/components/task_list/check-box-action-component.scss
@@ -8,8 +8,12 @@
     padding-right: govuk-spacing(3);
     padding-left: govuk-spacing(3);
 
-    > p {
+    p {
       @extend .govuk-hint;
+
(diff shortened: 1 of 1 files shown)


jkempster34 (approved):
This works great ✅ 

There is a lot to be removed with the old task list, but this not-so-nice `hint` argument (in application_helper) can go now I think 🙌   

```ruby
def render_markdown(markdown, hint: false)
    additional_attributes = ["target"]
    default_attributes = Loofah::HTML5::WhiteList::ALLOWED_ATTRIBUTES

    rendered_markdown = GovukMarkdown.render(markdown)
    rendered_markdown.gsub!("govuk-body-m", "govuk-hint") if hint

    sanitize(rendered_markdown, attributes: default_attributes.merge(additional_attributes))
  end
```

PR #435, opened 2023-01-05
Add optional voluntary conversion tasks

mec:
## Changes
The new `TaskList::OptionalTask` can now be used to make some tasks 'not
applicable' to a project.


jkempster34 (approved):

jkempster34 (changes requested):
We need the checkboxes adding to the views too

jkempster34 (approved):

PR #447, opened 2023-01-10
Move to final url scheme

mec:
## Changes

This work prepares for the swich to the new task list.

The end goal is to have paths:

`/conversions/voluntary/projects/id`
`/conversions/involuntary/projects/id`

We are shooting for Rails resource based paths but with the project types - this felt like a good enough approach.
This meant 

- refactoring the project views to be more modular
- creating new index views for conversions, voiluntary conversions and involuntary conversions
- showing 404 when the wrong type of project is requested on a path e.g involuntary project on the voluntary path
- moving the create new project paths

There is a lot here mostly becuase the Rails path helpers change.

https://trello.com/c/mt5NYc91


app/views/projects/shared/_signed_in_as.html.erb
```diff
@@ -0,0 +1 @@
+<p>You are signed in with the email address <%= email_address %></p>
```
beeLorna, line 1: I would prefer that this would be internationalised

beeLorna (approved):
Approved, but with just with a few small text changes. Not blocking

The last commit needs an edit on the first line, changing `voluntary` to `involuntary`.

There was a suggestion on changing some text to use internalisation, but other than that all looks good 👍

PR #450, opened 2023-01-11
Test all locale strings exist for Conversions

lozette:
## Changes

Add a feature test to check all locale strings are present for tasks.

This is more long-winded than I intended. I had hoped to do this more automatically - to visit the task list pages, iterate over any link on the page, check the content and then go back to the task list page. E.g. 

```
page.find_all("ol.app-task-list ul li a").each do |link|
  click_link link.text
  expect(page).to_not have_css(".translation_missing")
  click_link "Back"
end
```

But for some reason Rspec doesn't like this, it only goes into the first link and exits.

I also looked into using shared examples but I think these make it harder to debug and pinpoint where the missing strings are. 

In this implementation, every time a new task is added to a task list, it will have to be added to the list of tasks, plus the overall number of tasks will need to be amended. 

## Checklist

### DIFF
--- a/app/views/conversions/involuntary/task_lists/tasks/church_supplemental_agreement.html.erb
+++ b/app/views/conversions/involuntary/task_lists/tasks/church_supplemental_agreement.html.erb
@@ -6,7 +6,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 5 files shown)


lozette:
[Should we raise if the task has no locales as we are 100% reliant on them?](https://trello.com/c/S3B28wLG/1001-should-we-raise-if-the-task-has-no-locales-as-we-are-100-reliant-on-them)

PR #454, opened 2023-01-11
Update cookies page with essential cookies info

beeLorna:
This work encapsulates users being able to manage the optional cookies and implimenting Google Analytics based on that decision.

This was far more complicated than it should be and there are many improvements we could make, but this covers enough to deliver GA in the application.

Some decisions we made:

- don't rely on JS
- the cookie that stores the users choice for accepting optional cookies is itself an essential cookie
- the Rails flash message upon updating cookie preferences replaces the need for a 'hide cookie banner, banner'

The user flow is:

- any user (does not need to be signed in)
- show banner
- user accepts optional cookies
  - store their preference
  - show the flash
  - hide the banner
  - include GA
- user rejects optional cookies
  - store their preference
  - show the flash
  - hide the banner
  - do not inlcude GA

Users can always change their cookie preferences from the cookies page.

Google analytics will only be enabled in the production live environment.

### The final cookie page:
Screenshot 2023-01-30 at 15-54-18 Complete conversions transfers and changes

### The cookie banner:
Screenshot 2023-01-30 at 15-54-30 Complete conversions transfers and changes

### The cookie message:
Screenshot 2023-01-30 at 15-54-36 Complete conversions transfers and changes

## Checklist


lozette (approved):
Approved! Only thing I'd say is that there's a lot on not-internationalised text in this PR, and I don't know if you want to address this now, later or at all.

PR #464, opened 2023-01-13
Update contacts based on user feedback

mec:
## Changes

Users told us that the order and presentation of the contact category options could be improved. They felt having 'other' first and the default felt unhelpful.

They also felt a way to store 'the name of an organistation the contact represents' would be helpful.

We also missed adding a category for 'solicitor'

This PR updates the Contact model to meet these needs.

We re-ordered the contact category select box and added a 'choose category' option as the default that cannot be submitted.

We decieded not to use the 'choose category' option on the edit view as the category will have a value at this point.

We add a scope to order the contacts and groups on the index view and added test coverage for that.

The `organisation_name` attribute  gets added to Contact.

And fianlly we take the opportunity to switch to more specific back and redirect paths based on the project type and route.

https://trello.com/c/qhaJaaAh

image

### DIFF
--- a/app/views/contacts/_contacts.html.erb
+++ b/app/views/contacts/_contacts.html.erb
@@ -1,6 +1,6 @@
 <% @contacts.each do |category, contacts| %>
 
-  <h3 class="govuk-heading-m"><%= t("contact.index.category_heading", category_name: format_category_name(category)) %></h3>
+  <h3 class="govuk-heading-m"><%= t("contact.index.category_heading", category_name: category.humanize) %></h3>
 
   <% contacts.each_with_index do |contact, index| %>
 
--- a/app/views/contacts/_contacts.html.erb
+++ b/app/views/contacts/_contacts.html.erb
@@ -22,6 +22,10 @@
           row.key { t("contact.details.name") }
           row.value { contact.name }
         end
+        summary_list.row do |row|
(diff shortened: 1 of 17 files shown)


app/views/contacts/_contacts.html.erb
```diff
@@ -1,6 +1,6 @@
 <% @contacts.each do |category, contacts| %>
 
-  <h3 class="govuk-heading-m"><%= t("contact.index.category_heading", category_name: format_category_name(category)) %></h3>
+  <h3 class="govuk-heading-m"><%= t("contact.index.category_heading", category_name: category.humanize) %></h3>
```
lozette, line 3: Nice!

app/views/contacts/new.html.erb
```diff
@@ -9,10 +9,12 @@
     <%= form_for [@project, @contact], url: project_contacts_path(@project) do |form| %>
       <%= form.govuk_error_summary %>
 
-      <%= form.govuk_collection_select :category,
-            Contact.categories.keys,
-            :to_s,
-            ->(category_name) { format_category_name(category_name) } %>
+      <%= form.govuk_select(:category) do %>
+        <option selected="selected">Choose category</option>
```
lozette: Can this be internationalised please!

lozette (approved):
Approved but can you internationalise one string!

PR #468, opened 2023-01-16
Remove the old tasks and task list, sections, actions

lozette:
## Changes

https://trello.com/c/QLfvcVnm

While we implemented the new Task Lists, we kept the old Task list code hanging around. When we switched to using the new task list in the app, the old one was still there. Now it's time to remove it!

This PR removes the Task, Section and Action models, along with their individual specs, factories and database tables & foreign keys.

We have removed the Workflow YAML files which determined a project's Tasks, Sections and Actions. We have also removed the linters & schemas which ensured these Workflow files were in the correct format.

Notes, which can belong to both projects and tasks, are now completely decoupled from the old-style tasks and only belong to the new-style task lists.

Finally, remove some feature tests which only used the old style tasks lists. We already have feature tests for the new task lists.

## Checklist


lozette:
[Remove the old task list code and data](https://trello.com/c/QLfvcVnm/1020-remove-the-old-task-list-code-and-data)

lozette:
I assume the documents in `docs/workflows` can be removed as they relate to the YAML files. If not, lmk and I will drop the commit!

mec (approved):
Soooo good! Nice work - I think there is more, but this is a great start! 🚢

PR #473, opened 2023-01-19
Remove project details

mec:
## Changes

We were previously going to use this Converison::Project::Details to store non-task related attri\butes of a conversion project, this is no longer the case, eveything willl be in the task list tables - so we can remove this.

The only job this performed was supplying the correct 'conversion route' which we refactor into the conversion project itself before removal.


lozette (approved):
I completely missed that these weren't necessary any more!

PR #479, opened 2023-01-19
Suppress `nil` warning when running test

lozette:
## Changes

Running the test at `/spec/features/users_can_manage_contacts_spec.rb:79` was raising this warning:
```
Checking for expected text of nil is confusing and/or pointless since it will
always match. Please specify a string or regexp instead.
/spec/features/users_can_manage_contacts_spec.rb:79
```

Amend the test to only check for the `organisation_name` if it is not `nil`


mec (approved):
I remember seeing this and thought I had taken care of it!! Thank you!

PR #483, opened 2023-01-24
Task Generators

lozette:
## Changes

https://trello.com/c/9cuLFH1Y

The new Task List requires several files to be added to the codebase every time we add a new task to a project.

Add a rails generator to make this process easier. The involuntary/voluntary task generator will create a model, view, locale file and migration for you, given a task name.

The generators are tested by testing their output. Unfortunately this does not satisfy SimpleCov's method of calculating code coverage, so we have had to exclude the custom generators from  SimpleCov's coverage checker (see `spec/spec_helper.rb`) to keep coverage at 100%.

## Checklist

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -29,3 +29,4 @@
 /coverage
 /terraform
 /.idea/
+lib/generators/*/USAGE
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - the order of the contact groups is now as users might expect.
 
+### Added
+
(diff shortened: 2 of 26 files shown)


mec (approved):
Testing these indirectly feels like the right amount of effort.

As this is tooling to make developers lives simpler, it is frustrating that this sets off SonarClouds duplication and code smell alerts - but I am not convinced it is worth the effort to resolve those just now, perhaps one for the backlog.

lozette:
I hadn't noticed the code duplication bit - that's a fairly simple refactor I reckon

lozette:
I'm going to ignore the code smells - it wants me to add a comment to the fixture files, but that would mean they were no good for comparing against the generated example files, as they'd be different 🤦

PR #490, opened 2023-02-02
Users can supply the confirmed conversion date

mec:
This work introduces data collection to our task lists. 

Here we are adding the confirmed conversion date to the stakeholder kick off  task for both involuntary and voluntary conversions.

Collecting this date replaces the need for users to check off the previous action as the completion of the action is implied in providing the value.

Some thinhgs to look out for:

- our task objects loosley follow the form object pattern, forms submit to them and then they take care of saving the values to the database via the task list - this means when submitting a date, our old friend multiparameter date shows up, handling it here is fairly striaght forward but might look a bit unusual - let me know what you think?
- in order to use the GOVUK form builder helpers, we have to inject our guidance, this works, but is not a great pattern to follow - ideally we would subclass the helper and add the guidance support - I didn't do this as I think it can come later with more date fields and a spike into how we could do it!
- Sonar cloud is complaining about duplication, we expect this as we made the decision to have duplicated task_lists for voluntary and involuntary conversions even though they are mostly the same.
- I kept the validation on the date loose - there are many refinements we could make but I have chosen not to at this point in the spirit of shipping small value more often.
- the trello card calls for more changes to later tasks - I haven't done these as I think there is a different approach we can take to the proposed ones and we won't have projects in those later task righrt now so there is no imediate benefit to users.

https://trello.com/c/Bgkj9hex

image

Screenshot 2023-02-02 at 12-14-16 Complete conversions transfers and changes

Screenshot 2023-02-02 at 12-15-06 Complete conversions transfers and changes


CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   task will become in-progress after this change. Users must supply a value for
   this.
 - The conversion date is now shown as either provisional or confirmed.
+- The Stakeholder kick off, confirmed conversion date is now validated for
+  correcct format.
```
lozette, line 22: Typo `correcct`

lozette (commented):
Added comments for now, will come back to this tomorrow for a fresh look!

spec/features/users_can_complete_conversion_voluntary_tasks_spec.rb
```diff
@@ -23,14 +23,31 @@
     one_hundred_and_twenty_five_year_lease subleases tenancy_at_will
     trust_modification_order]
 
+  tasks_with_collected_data = %w[
```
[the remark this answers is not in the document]
mec: Yeah, I have some ideas that might pan out.

spec/helpers/project_helper_spec.rb
```diff
@@ -54,4 +54,13 @@
       expect(subject).to eq "mailto:john.doe@education.gov.uk"
     end
   end
+
+  describe "#converting_on_date" do
```
lozette, line 58: This should have gone into the previous commit!

lozette (approved):
Approved, although one bit has ended up in the wrong commit, so please fix before merge!

PR #498, opened 2023-02-06
Add notification banner component

mec:
We make use of the Rails flashes. Our general approach is to show one
after a successful operation such as:

- user adds a note
- user adds a contact
- user adds a project
- user signed out

We chose to use the Notfication banner component for this [1], did not
use the explicit 'success' version. This version is designed
specifically for

> Reacting to something the user has done

which is a much better fit for our use case.

We do have a couple of instances where the flash is used for
unsuccessful events, such as:

- user is not registered
- user could not be signed in
- user is not authorised

Here we introduce a view component to handle both cases, using a view
component helps us encapsulate all the behaviour and make it easy to
test.

The component takes the two standard Rails flash types notice and alert
and renders the appropriate design.

We also add styles to control the content of the flashes so that
developers do not have to.

[1] https://design-system.service.gov.uk/components/notification-banner/

Screenshot 2023-02-06 at 12-12-24 Complete conversions transfers and changes

### DIFF
--- a/spec/components/notification_banner_component_spec.rb
+++ b/spec/components/notification_banner_component_spec.rb
@@ -1,0 +1,35 @@
+require "rails_helper"
+
+RSpec.describe NotificationBanner, type: :component do
+  context "when there is a notice flash" do
+    it "displays the correct notification banner" do
+      flash = ActionDispatch::Flash::FlashHash.new
+      flash.notice = "This is a notice"
+
+      render_inline(described_class.new(flashes: flash))
+
+      expect(page).to have_selector(".govuk-notification-banner--success")
+      expect(page).to have_text("Success")
+    end
+  end
+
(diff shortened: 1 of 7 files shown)


spec/components/notification_banner_component_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      expect(page).to have_selector(".govuk-notification-banner--success")
+      expect(page).to have_text("Success")
+    end
+  end
+
+  context "when there is an alert flash" do
+    it "displays the correct notification banner" do
+      flash = ActionDispatch::Flash::FlashHash.new
+      flash.alert = "This is a alert"
```
lozette, line 19: Annoying typo: `an alert` 😁

lozette (approved):

PR #500, opened 2023-02-06
Create internal and external contacts

lozette:
## Changes

Move the Caseworker, Team leader and Regional delivery officer to a new "Internal contacts" tab on the project details. These users can be assigned from this tab as usual.

Rename the Contacts tab External contacts, and alias the routes for `contacts`.

Screenshot 2023-02-06 at 14 42 18
Screenshot 2023-02-06 at 14 42 08

## Checklist

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -10,7 +10,7 @@ class AssignmentsController < ApplicationController
   def update_team_leader
     @project.update(team_leader_params)
 
-    redirect_to project_internal_contacts_path(@project), notice: t("project.assign.team_leader.success")
+    redirect_to helpers.path_to_project_internal_contacts(@project), notice: t("project.assign.team_leader.success")
   end
 
   def assign_regional_delivery_officer
@@ -20,7 +20,7 @@ class AssignmentsController < ApplicationController
   def update_regional_delivery_officer
     @project.update(regional_delivery_officer_params)
 
-    redirect_to project_internal_contacts_path(@project), notice: t("project.assign.regional_delivery_officer.success")
+    redirect_to helpers.path_to_project_internal_contacts(@project), notice: t("project.assign.regional_delivery_officer.success")
   end
 
   def assign_caseworker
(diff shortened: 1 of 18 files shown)


lozette:
[Capture the user that is assigned to a project](https://trello.com/c/b0u5qx4F/1106-capture-the-user-that-is-assigned-to-a-project)

app/views/assignments/show.html.erb
```diff
@@ -0,0 +1,43 @@
+<% content_for :pre_content_nav do %>
+  <% render partial: "shared/back_link", locals: {href: root_path} %>
+<% end %>
+
+<%= render partial: "projects/index/project_summary", locals: {project: @project} %>
```
mec, line 5: I don't think this is the correct summary - this is the one used on index pages.

app/controllers/assignments_controller.rb
```diff
@@ -3,14 +3,17 @@ class AssignmentsController < ApplicationController
   before_action :authorize_user
   after_action :verify_authorized
 
+  def show
```
mec, line 6: I was expecting `internal_contacts_controller` what do you think - are there advantages to having this as the assignment show action?

mec (changes requested):

app/views/assignments/show.html.erb
lozette: Oops, too many partials 🤦

app/controllers/assignments_controller.rb
lozette: No advantage, I was just piggybacking on an existing one for convenience!

app/views/assignments/show.html.erb
mec: for sure!

spec/requests/assignments_controller_spec.rb
```diff
@@ -19,6 +19,52 @@
     end
   end
 
+  describe "#show" do
```
mec, line 22: This should be in a different file now as it is to do with the internal contacts controller?

mec (changes requested):
I feel bad for rushing you on this - sorry. 

Couple of things on here.

lozette:
[Move the Caseworker, RDO and Team lead to a new tab](https://trello.com/c/r7ZAmgP4/1124-move-the-caseworker-rdo-and-team-lead-to-a-new-tab)

mec (approved):
Looks great - thank you! 🚢

PR #512, opened 2023-02-13
Capture if a project is being handed over to Regional Caseworker Team 

lozette:
## Changes

When creating a new voluntary project, allow the user to indicate if the project will be handed over to Regional Caseworker Services or not.

Screenshot 2023-02-13 at 16 30 51

If yes, the project will be created with `assigned_to` left blank, and an email will be sent to the team leaders.

If no, the creating user will be `assigned_to` the project, and no email will be sent to team leaders.

There is no visible change for involuntary projects.

## Checklist

### DIFF
--- a/db/migrate/20230210105741_add_assigned_to_regional_caseworker_team_to_projects.rb
+++ b/db/migrate/20230210105741_add_assigned_to_regional_caseworker_team_to_projects.rb
@@ -1,0 +1,5 @@
+class AddAssignedToRegionalCaseworkerTeamToProjects < ActiveRecord::Migration[7.0]
+  def change
+    add_column :projects, :assigned_to_regional_caseworker_team, :boolean, default: false
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   new "Internal contacts" tab.
 - The notification banner shown to users after a successful operation has a new
   visual style.
+- If a voluntary project is being handed over to Regional Casework Services,
+  send the team leader notification; if it is not, assign the current user to
+  the project and do not send the notification.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Any user can now assign a project to any other user, via the "Assign to" row
   on the project's "Internal contacts" tab.
 - add healthcheck endpoint
+- The form to create a voluntary conversion project allows the user to indicate
+  if the project is being handed over to Regional Casework Services
 
 ## [Release 14][release-14]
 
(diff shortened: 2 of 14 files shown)


lozette:
[Capture if a project is being handed over to the regional case worker team](https://trello.com/c/sj1ipeXD/1108-capture-if-a-project-is-being-handed-over-to-the-regional-case-worker-team)

db/migrate/20230210105741_add_assigned_to_regional_caseworker_team_to_projects.rb
```diff
@@ -0,0 +1,5 @@
+class AddAssignedToRegionalCaseworkerTeamToProjects < ActiveRecord::Migration[7.0]
+  def change
+    add_column :projects, :assigned_to_regional_caseworker_team, :boolean, default: false
```
mec, line 3: Do we need to migrate any data at this point? Happy to do so on the console as the number of project we have is so low.

app/forms/conversion/involuntary/create_project_form.rb
```diff
@@ -9,7 +9,8 @@ def save
       provisional_conversion_date: provisional_conversion_date,
       advisory_board_date: advisory_board_date,
       regional_delivery_officer_id: user.id,
-      task_list: Conversion::Involuntary::TaskList.new
+      task_list: Conversion::Involuntary::TaskList.new,
+      assigned_to_regional_caseworker_team: false
```
mec, line 13: I don't think we need to implement anything for involuntary? We have the two subclasses for just this use case?

app/forms/conversion/create_project_form.rb
```diff
@@ -14,6 +14,7 @@ class NegativeValueError < StandardError; end
   attribute :advisory_board_conditions
   attribute :note_body
   attribute :user
+  attribute :assigned_to_regional_caseworker_team, :boolean
```
mec, line 17: I think we can declare this in the `Conversion::Voluntary::CreateProjectForm` as it is only for that class of project?

app/forms/conversion/create_project_form.rb
```diff
@@ -41,6 +42,12 @@ def initialize(params = {})
     super(params)
   end
 
+  CASEWORKER_TEAM_RESPONSES = [OpenStruct.new(id: true, name: I18n.t("yes")), OpenStruct.new(id: false, name: I18n.t("no"))]
```
mec, line 45: Same with this, as this is only for `voluntary` does it make more sense delcared there?

mec (changes requested):
I think we can make better us of the subclasses in this to keep behaviour in the appropriate class?

db/migrate/20230210105741_add_assigned_to_regional_caseworker_team_to_projects.rb
lozette: We're going to do this manually via the console after deployment.

app/forms/conversion/involuntary/create_project_form.rb
lozette: Good point, I originally did this for both but then backtracked, and didn't backtrack enough 😆

mec (approved):
🚢

PR #513, opened 2023-02-14
Add healthcheck endpoint

beeLorna:
## Changes

Added in `/check` endpoint for TechOps to monitor that the app is up and responding.

## Checklist

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -103,6 +103,8 @@ Rails.application.routes.draw do
   get "cookies", to: "cookies#edit"
   post "cookies", to: "cookies#update"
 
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -99,6 +99,8 @@
   get "cookies", to: "cookies#edit"
   post "cookies", to: "cookies#update"
 
+  get "check" => "healthcheck#check"
```
mec, line 102: Could this be `healthcheck` check is a littel ambighious?

mec (approved):
This looks good as a first step with a tweak to the url - I do wonder if it should check the database connection given how flakey that can be?

PR #519, opened 2023-02-20
Update Container Apps TF module

DrizzlyOwl:
## Changes

Updates the upstream terraform module to v0.13.7 which contains bug fixes to prevent the Worker container from scaling down to zero, and adds a new Slack web hook monitoring feature.


mec (approved):
🚢

PR #520, opened 2023-02-20
Add Assign button to the Unassigned project view

lozette:
## Changes

In a previous commit we added an Unassigned projects tab for Team leaders only. The projects on this tab now have an Assign button, via which the Team leaders can assign a project to another user. Once assigned, the Team leader is redirected to the project's Internal contacts tab. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   kick-off task
 - Add new action Check provisional conversion date to the involuntary
   stakeholder kick-off task
+- Add Assign button to projects on the Unassigned projects tab
 
 ## [Release 14][release-14]
 
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -41,9 +41,10 @@ class AssignmentsController < ApplicationController
   end
 
   def update_assigned_to
-    @project.update(assigned_to_params)
+    @project.update(assigned_to_params.except(:return_to))
+    return_to = assigned_to_params[:return_to] || helpers.path_to_project_internal_contacts(@project)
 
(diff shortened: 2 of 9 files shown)


lozette:
[Add 'assign' button to the unassigned project view](https://trello.com/c/OtIFl807/1111-add-assign-button-to-the-unassigned-project-view)

mec (commented):
My expectation was that the loop should be:

```
unassigend project view > assign form > submit assignment > unassigned project view
```

So that multiple assignment can happen efficiently, instead users are returned to the project view, which is less helpful in my view?

I also think we should show the assign button only on the unassigned project view, currently a team lead also sees it on the 'In progress' projects view.

Lastly, I thought we were not including the 'Completed by' field as it feels really odd to me, was it added back for a reason?

This speaks to a wider issue where a lot of the lucid boards are far from clear on what decisions have been made and I am not sure how to resolve that!?

lozette:
@mec 

`So that multiple assignment can happen efficiently, instead users are returned to the project view, which is less helpful in my view?` - yes this can happen. Because we can reach the Assign form from two places, this means we'll have to redirect back to wherever we came from, but it can be done.

`I also think we should show the assign button only on the unassigned project view, currently a team lead also sees it on the 'In progress' projects view.` - That's a bug!

`Lastly, I thought we were not including the 'Completed by' field as it feels really odd to me, was it added back for a reason?` - it was in the Lucid board so I added it! I will take it out.

mec:
I know - I am sorry - the lucid boards make me grumpy, for me, they are not clear at all!

lozette:
This is ready for a re-review @mec

mec (approved):
Looks good! We have a bunch of places that we do this kind of redirect back to different locations - it would be great to unify our approach - I don't want to block this work though - so I am going to write up a card. 🚢

PR #525, opened 2023-02-22
Increase page width to 1024px

mec:
## Changes

Right now all our users are internal DfE users who have DfE provided
devices. Our analytics show that increasing the width of our viewport,
the space inside the browser window is viable as no one accesses the
service on a screen that cannot accomadte this change.

Horizontal space will place an important role in the application as we
introduce tables and these extra 64px could mean being able to show an
extra column.


beeLorna (approved):
LGTM 👍

PR #530, opened 2023-02-24
Add new `CONTAINER_VNET` value to `config.hosts`

lozette:
Re: https://ukgovernmentdfe.slack.com/archives/C03AANC6UJF/p1677253180105049

We added a new `/healthcheck` endpoint for monitoring.

Unfortunately when implementing the monitoring, techops encountered an error because the health probes' IP address is not in the app's allowed hosts:

```
ERROR -- : [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked host: 172.16.0.126
```

As a solution, techops are going to add a new environment variable to terraform called `CONTAINER_VNET`. This will be an IP range.

https://guides.rubyonrails.org/configuring.html#actiondispatch-hostauthorization

We can add this to the allowed hosts in the application via `IPAddr.new(ENV["CONTAINER_VNET"])`

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -85,6 +85,7 @@ Rails.application.configure do
 
   # configure the host names
   config.hosts << ENV["HOSTNAME"]
+  config.hosts << IPAddr.new(ENV["CONTAINER_VNET"])
   config.hosts.concat ENV.fetch("ALLOWED_HOSTS", "").split(",")
 
   # configure ActionMailer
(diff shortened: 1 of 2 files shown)


mec (changes requested):
Looks good, could we add the variable to the docs though please: https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/main/doc/environment-variables.md

lozette:
[Allow IP address ranges to be added to the allowed hosts list.](https://trello.com/c/yIsnzgK7/1114-allow-ip-address-ranges-to-be-added-to-the-allowed-hosts-list)

mec (approved):

PR #534, opened 2023-02-28
Add Project Openers table

lozette:
## Changes

Add a route to a new view at `projects/openers/:month/:year`. The route is constrained by month & year being present, and within the scope of 1-12 (month) and 2000-2499 (year).

This page shows a tabular view of all projects. The list of projects is filtered to projects which have a confirmed opening date of the supplied year & month.

If a year or month are not supplied in the URL, the user is redirected to a 404. 

If the year is not in the scope of 2000-2499, or the month is not in the scope of 1-12, the user is redirected to a 404. 

If no projects fall within the scope of a correctly supplied month and year, an appropriate informational message will be shown.

Screenshot 2023-02-28 at 16 25 45

Screenshot 2023-03-06 at 15 42 46

As a side effect of doing this work, we now have URL constraints on the `project/:id` URLs (for "plain" projects, and voluntary/involuntary conversion projects). This is to prevent an error when a user tries to go to `projects/openers` and the application thinks the user is trying to access a project with the ID of "openers". The constraint checks that a supplied project ID is a valid UUID, and redirects the user to a 404 page if it is not.

## Checklist

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -32,8 +32,7 @@ class ProjectsController < ApplicationController
     year = params[:year]
     month = params[:month]
 
-    # TODO: add a scope for projects with a conversion date of this month & year
-    @projects = Project.all
+    @projects = Project.opening_by_month_year(month, year)
     @date = "#{Date::MONTHNAMES[month.to_i]} #{year}"
   end
 end
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -25,4 +25,15 @@ class ProjectsController < ApplicationController
     @project = Project.find(params[:id])
     authorize @project
   end
+
+  def openers
+    authorize Project
+
+    year = params[:year]
+    month = params[:month]
+
+    # TODO: add a scope for projects with a conversion date of this month & year
+    @projects = Project.all
+    @date = "#{Date::MONTHNAMES[month.to_i]} #{year}"
+  end
 end
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -42,6 +42,8 @@ class Project < ApplicationRecord
   scope :unassigned_to_user, -> { where assigned_to: nil }
   scope :assigned_to_regional_caseworker_team, -> { where(assigned_to_regional_caseworker_team: true) }
 
+  scope :opening_by_month_year, ->(month, year) { where.not(conversion_date: nil).and(where("YEAR(conversion_date) = ?", year)).and(where("MONTH(conversion_date) = ?", month)) }
+
   def establishment
     @establishment ||= fetch_establishment(urn)
   end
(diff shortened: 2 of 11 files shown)


lozette:
@mec my complete inability to read 🤦  It should be `/:month/:year` according to the ticket 🤦

lozette:
@mec month & year swapped

app/controllers/projects_controller.rb
```diff
@@ -1,6 +1,7 @@
 class ProjectsController < ApplicationController
   after_action :verify_authorized
   after_action :verify_policy_scoped, only: :index
+  before_action :validate_year_month, only: :openers
```
mec, line 4: I would like to see a seperate controller for this to stop projects controller getting bloated, might just be a preference thing though, so I will keep going, take this as non-blocking!

mec (changes requested):

config/routes.rb
```diff
@@ -81,6 +81,8 @@
     collection do
       get "completed"
       get "unassigned"
+
+      get "openers(/:month)(/:year)", to: "projects#openers"
```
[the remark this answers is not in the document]
lozette: Good call, did not know those were a thing!

config/routes.rb
```diff
@@ -109,4 +109,6 @@
   # High voltage configuration for static pages. Matches routes from the root of the domain. Uses
   # HighVoltage::Constraints::RootRoute to validate that the view exists.
   get "/*id" => "pages#show", :as => :page, :format => false, :constraints => HighVoltage::Constraints::RootRoute.new
+
+  match "*unmatched", to: "application#not_found_error", via: :all
```
mec, line 121: I've never seen `match` or `*unmatched` before?! How did you find those?

config/routes.rb
```diff
@@ -65,39 +65,46 @@
     get "/", to: "/conversions/projects#index"
     namespace :voluntary do
       get "/", to: "/conversions/voluntary/projects#index"
+      get "projects/:id", to: "/conversions/voluntary/projects#show", constraints: {id: /[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}/i}, as: :project
```
mec, line 68: Is it possible to extract the regex into a mehtod to give it some context, something lke `valid_uuid`, maybe?

app/models/project.rb
```diff
@@ -42,6 +42,8 @@ class Project < ApplicationRecord
   scope :unassigned_to_user, -> { where assigned_to: nil }
   scope :assigned_to_regional_caseworker_team, -> { where(assigned_to_regional_caseworker_team: true) }
 
+  scope :opening_by_month_year, ->(month, year) { where.not(conversion_date: nil).and(where("YEAR(conversion_date) = ?", year)).and(where("MONTH(conversion_date) = ?", month)) }
```
mec, line 45: Clever use of SQL here!

mec (approved):
This all seems reasonable, I can't shake the feeling that we shouldn't have to work this hard on the routes - but this approach works and I can't offer an alternative!

config/routes.rb
beeLorna: Second that!

beeLorna (approved):
LGTM! 🥳

config/routes.rb
lozette: I've updated the commit message to add this info.

app/models/project.rb
lozette: Had to do a bit of googling to get the MSSQL syntax!! `YEAR()` and `MONTH()` are apparently MSSQL specific.

PR #537, opened 2023-03-01
Store the primary converison date on Project

mec:
Right! So I have been back and forth on this problem for far too long and became completely stuck!

Every approach I tried, led down a path that ended with an unacceptble 'something' that meant I wasn't happy.

I think this speaks to a wider model issue we have with the whole task list vs project, but we need to move forwards and cannot afford to linger on that for long.

The goal here is:

- we want the conversion date on the project to make it super easy to sort and scope projects by the date
- we want to know when a projects conversion date is 'provisional' i.e not yet agreed with all parties
- we want to change the date later, but only via a specific journey where we also capture why the date chagned and the history of it changing - we think this will help improve how projects get managed later

To facilitate this and not get stuck, this work is what I propose!

We add a `conversion_date` to project to store the value.

As it is super hard to get at this value from our super duper task list models, we side step that issue by syncronising the task list conversion date with the project conversion date when the task list gets save.

We have guards on this to prevent it happening:

- when the is no project i.e the project and task list are being saved for the fist time ever
- when the project already has a conversion date i.e it has been set once or set by the to be slipping behaviour

I know what you are thinking and I would argue this is okay for lots of reasons but mostly becuase the process is that that first sync is the only one - once the date is confirmed on that external stakeholder kick off task, the only way users should be able to change it is by 'slipping' - which is it's own thing.

Later we will allow the user to update the `Project.converison_date` directly, storing a history of changes and the why it changed as a note.

I've ended this work here as having `conversion_date` on project means we can populate the openers list with confirmed projects, which is one  of the goals we want to meet.

Deploying this work will prevent users from changing the conversion date more than once and I think that is okay for now becuase of the whole 'slipping' conversion date i.e we don't want user changing the date without telling us why.


app/helpers/project_helper.rb
```diff
@@ -16,11 +16,9 @@ def mail_to_path(email)
...
-    conversion_date = project.conversion_date
+    return project.conversion_date.to_formatted_s(:govuk) unless project.conversion_date.nil?
 
-    return conversion_date.date.to_formatted_s(:govuk) unless conversion_date.provisional?
-
-    date = conversion_date.date.to_formatted_s(:govuk)
+    date = project.provisional_conversion_date.to_formatted_s(:govuk)
     tag = govuk_tag(text: "provisional", colour: "grey")
 
     "#{date} #{tag}".html_safe
```
lozette, line 24: This hasn't been introduced in this commit but I'm wondering why this method either returns a formatted date, or an HTML tag containing the date. Line 19 returns a formatted date if the `project.conversion_date` is not nil, and otherwise returns `"#{date} #{tag}".html_safe` which is an HTML snippet. Seems odd that it returns two different things?

lozette (approved):
Approved, just one comment which is from earlier work and is non-blocking!

app/helpers/project_helper.rb
mec: Yeah, it sucks! I'll make a note to come back to it as I work through more conversion date and slipping stuff.

mec: Second thought! This is a helper and it returns a string that gets rendered into a view - either jsut a date, or a date with the 'provisional' tag next to it, you could argue it does return the same thing - but I will try and make it a bit better if I can!?

PR #545, opened 2023-03-08
Fetch and show the school address in school details on project information tab

mec:
We know that Academy operational pracitce unit users rely on the address
of the school as part of sending the fundaing agreement letters.

As we fetch most school information from Get information about schools
via the academies API, we want to validate we can fetch the address.

Once we have the address and show it, users can gain confidence that it
is the 'correct' one.

Later we can then export the address in ways that are useful to users,
but that comes later.

Some decisions here:

- went for the address over multiple lines [1]
- wrapped the address in the <address> tag, will worry about this not
  working in older browsers if it is an issue
- added a basic style as the address is render in italics otherwise;
  might be a browser default?
- refactored the project information feature specs into their own
  directory and split them up to aid readability

https://trello.com/c/QijLj2Ez

[1] https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#addresses-in-the-uk

image


app/helpers/project_helper.rb
```diff
@@ -35,4 +35,8 @@ def link_to_trust_on_gias(ukprn)
 
     link_to(t("project_information.show.trust_details.rows.view_in_gias"), "https://get-information-schools.service.gov.uk/Groups/Search?GroupSearchModel.Text=#{ukprn}", target: :_blank)
   end
+
+  def address_markup(address)
+    tag.address address.compact.join("</br>").html_safe, class: %w[govuk-address]
```
lozette, line 40: The tag should be `<br/>`

lozette (changes requested):
One tiny change to the markup

lozette (approved):

PR #546, opened 2023-03-08
Update Terraform and Terraform provider versions

DrizzlyOwl:
## Changes

- Updated Terraform version to v1.3.9 and bumped Terraform provider dependencies to their latest supported versions.


mec (approved):
🚢

PR #550, opened 2023-03-10
Refactor provisional conversion date

mec:
This work is the refactoring of how we model the conversion date.

The conversion date moves through three main states:

provisional > confirmed > changed

It can change multiple times, but it is only ever provisional once. A chagned date is still considered confirmed as it would have to be agreed by all parties.

This work is prompted by the difficulty we face building views that rely on the converison date being one column in the database when it may be represented by two, `provisional_converison_date` or `conversion_date`.

Keeping the test suite green through this means a lot has to change in single commits - apologies for that.

With the refactor done, we have one database column that can be used to sort and order projects - which makes working with them much simpler, it also resolves the bug we have where projects are always grouped by the provisional date, even after the date is no longer considered provisional.

MIgrating the projects in production has proven to be a real head scratcher (for me at least!)- I ended up writhing a small 'service' object so that I could test the result and calling that from the migration. In theory we can remove the service and migration once it is done, but I am not sure if that would have other impacts?

As far as I understand it, in production the migrations are only ever run once, unlike locally where we can rollback and forwards - so there shouldn't be a risk there.

A lot of the work here focuses on changing how the inital provisional to confirmed change happens, the approach is that all project are created with a provisional date (by defaulting `conversion_date_provisional` to true) - we can be confident the first change must be to confirmed - a service object handles all the changes so it can be called from the various places that need to trigger a change: the task list save callbacks and the change conversion date form.

We action the change using the same mechanisms we use when a user changes the date, by creating a `Conversion::DateHisotry`. Becuase a date history requires a note and a note requires a user, we have to inject a user into the task list to allow this to happen.

This work does not include removing the `provisional_conversion_date` column as I would like to see the data migration thorugh before deleting data.

Thanks for reviewing this work, it's been a really complicated one to unravel.

### DIFF
--- a/spec/services/project_conversion_date_migrator_spec.rb
+++ b/spec/services/project_conversion_date_migrator_spec.rb
@@ -1,0 +1,76 @@
+require "rails_helper"
+
+RSpec.describe ProjectConversionDateMigrator do
+  before do
+    mock_successful_api_calls(establishment: any_args, trust: any_args)
+  end
+
+  describe "#migrate_up!" do
+    context "when the project is provisional" do
+      it "sets the conversion date to the provisional date" do
+        user = create(:user)
+        project = create(:conversion_project, provisional_conversion_date: Date.today.at_beginning_of_month)
+        project.conversion_date = nil
+        project.save(validate: false)
+
+        project.task_list.user = user
+        project.task_list.stakeholder_kick_off_confirmed_conversion_date = nil
+        project.task_list.save
+
+        described_class.new(project).migrate_up!
+        project.reload
+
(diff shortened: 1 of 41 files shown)


spec/services/project_conversion_date_migrator_spec.rb
```diff
@@ -0,0 +1,71 @@
+require "rails_helper"
+
+RSpec.describe ProjectConversionDateMigrator do
```
lozette, line 3: This is in the wrong commit, should be in the next one!

lozette (approved):

PR #567, opened 2023-03-20
Rationalise project views

mec:
Ahead of work to update our permissions I wanted to rationalise the project views.

We use policy scopes to show the user their 'own' projects i.e the ones to which they are `assigned_to` - this is not a great use of policy scopes as this is not really a policy. User are free to view any projects, but only modified their own.

To move the scopes out the policy and into model scopes, I first wanted to make views that reflect the way we know users group projects:

- all projects that are in-progress
- all projects that are completed
- all projects in Regional casework services that are in-progress
- all projects in Regional casework sercices that are completed
- all projects in Regional casework services that are unassigned i.e those that are not yet assigned to a user

There is one more scope that will come later as it is more complex: projects that are in a specific region; This can now follow the structure set out here.

With this work done, we can strip out the policy scopes and switch the matching views to these new ones.

I also believe there is value in rationalising the views and paths to better speak to the behaviour. 

There is a lot here, mostly boiler plate views - sorry about the size of the PR.


config/locales/project.en.yml
```diff
@@ -90,12 +92,16 @@ en:
...
+        assign: Assign project
       body:
         view_html: View <span class="govuk-visually-hidden">%{school_name}</span> project
+        assign_html: Assign <span class="govuk-visually-hidden">%{school_name}</span> project
       in_progress:
         empty: There are no projects in progress
       completed:
         empty: There are no completed projects
+      unassigned:
+        empty: There are no unassigend projects
```
lozette, line 104: Typo `unassigned`

lozette (changes requested):
*Phew*
I kind of hated the URLs at first because I don't think they're very "discoverable", but then none of the URLs in the service are 🤷 and I can't suggest anything else.

You have a typo which needs fixing tho.

mec:
@lozette yeah I am not sure what discoverable would look like in this context. I know you said you have no suggestions, but I would be keen to hear thoughts on the paths - I am seeing these getting wrapped in navigation at some point soon, perhaps that makes this feel better?

lozette (approved):

PR #569, opened 2023-03-20
Feature/1283 conversion project stores directive academy order issued

lozette:
## Changes

## Checklist


db/migrate/20230317114940_add_directive_academy_order_to_projects.rb
```diff
@@ -0,0 +1,14 @@
+class AddDirectiveAcademyOrderToProjects < ActiveRecord::Migration[7.0]
+  def up
+    add_column :projects, :directive_academy_order, :boolean, default: false
+
+    Project.all.each do |project|
+      project.directive_academy_order = !(project.task_list_type == "Conversion::Voluntary::TaskList")
```
lozette, line 6: I actually hate this syntax but the linter is demanding it. I think it's quite obfuscated.

lozette:
[Conversion project stores 'directive academy order issued?'](https://trello.com/c/RRw19HkF/1283-conversion-project-stores-directive-academy-order-issued)

app/forms/conversion/create_project_form.rb
```diff
@@ -36,11 +38,17 @@ class NegativeValueError < StandardError; end
   validate :establishment_exists, if: -> { urn.present? }
   validate :trust_exists, if: -> { incoming_trust_ukprn.present? }
 
+  validates :directive_academy_order, inclusion: {in: ["true", "false"]}
```
lozette, line 41: These are strings not booleans, as they're coming from a form I think? Bools didn't work here. Happy to be corrected if I've done it wrong!

app/forms/conversion/create_project_form.rb
```diff
@@ -36,11 +38,17 @@ class NegativeValueError < StandardError; end
   validate :establishment_exists, if: -> { urn.present? }
   validate :trust_exists, if: -> { incoming_trust_ukprn.present? }
 
+  validates :directive_academy_order, inclusion: {in: [true, "true", false, "false"]}
```
lozette, line 41: This is awful but it seems like these are bools or strings depending of whether they're coming from a test or the UI?!

db/migrate/20230317114940_add_directive_academy_order_to_projects.rb
mec: Yeah, it's pretty hard to parse for me too! The only thing I could propose would be:

```
def directive_academy_required(project)
return false if project.task_list_type == "Conversion::Voluntary::TaskList"
return true if project.task_list_type == "Conversion::Inoluntary::TaskList"
end
```

app/forms/conversion/create_project_form.rb
mec: I feel like these should be strings in the tests as that is what gets submitted from the form - but it is not a deal breaker - we have the same thing for `assigned_to_regional_caseworker_team`

mec (approved):
I left a couple of comments - nothing that would stop this going in though. 👍

PR #594, opened 2023-03-24
Validate new projects URN

mec:
The same school should never been converting more that once, we've seen
some issues in production where different users add the same project
resulting in duplicates.

This works prevents the issue by checking that a project that is in
progress does not already exist and if it does, showing a simple error
message.

https://trello.com/c/TBOAIBeo

Screenshot 2023-03-24 at 13-14-22 Complete conversions transfers and changes

### DIFF
--- a/spec/support/shared_examples/create_conversion_project_form.rb
+++ b/spec/support/shared_examples/create_conversion_project_form.rb
@@ -204,6 +204,17 @@ RSpec.shared_examples "a conversion project FormObject" do
         expect(form).to be_invalid
       end
     end
+
+    context "when there is another in-progress project with the same urn" do
+      it "is invalid" do
+        _project_with_urn = create(:conversion_project, urn: 121813)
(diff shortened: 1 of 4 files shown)


spec/support/shared_examples/create_conversion_project_form.rb
```diff
@@ -204,6 +204,16 @@
         expect(form).to be_invalid
       end
     end
+
+    context "when there is another in-progress project with the same urn" do
+      it "is invalid" do
```
lozette, line 209: Can we add another test just to check for the presence of the expected error message?

lozette (changes requested):

spec/support/shared_examples/create_conversion_project_form.rb
mec: @lozette added that coverage and a tweak. 👍

PR #597, opened 2023-03-27
Set the `assigned_at` DateTime when a user is keeping a project they created

lozette:
## Changes

A small bug in project creation meant that when a user creates a project and is *not* handing it over, the `assigned_at` DateTime was not being set on the project. The `assigned_at` time should reflect when someone was added to the `assigned_to` field on the project.

## Checklist


mec (approved):
I would prefer to see these specs on the form object rather than a feature, two reasons:

- they could be faster
- it would gather all the 'documentation' around creating a project into one place, a lot happens when we create a project and it could be valuable to have it all in one spec

It's not a deal breaker though, nice work spotting this! 👍

PR #601, opened 2023-03-27
Add tflint test for terraform

DrizzlyOwl:
- Add a new GitHub Action that runs `tflint` against the Terraform code to check for any issues and encourage best practices.


lozette (approved):

lozette:
Would it be useful (and possible) to also add this to our local linters?

PR #616, opened 2023-03-29
User can view statistics page for all project statistics

beeLorna:
## Changes

We now have a statistics page which gives an overview of all projects split into different categories:

- a general overview of all projects
- Regional casework services projects
- Regional projects
- Projects per region
- a 6 month view of all project openers

page can be found at `projects/all/statistics`
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - New views at `/projects/regional/:region/in-progress` and
   `/projects/regional/:region/completed` to show projects NOT assigned to
   Regional caseworker services, filtered by region
+- User can view statistics page for all project statistics
 
 ## [Release 19][release-19]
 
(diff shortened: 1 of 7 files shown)


beeLorna:
Any feedback / suggestions regarding the wording and naming used within the UI would be very much appreciated. Thanks!

lozette:
I think you can refactor this to be smaller by using loops on the Project.regions in a similar way to using loops for the month. E.g. To get the total Projects for each region in a hash:
```
  def total_regional_projects
    result = {}
    Project.regions.each do |key, _value|
      result[key] = @projects.by_region(key).count
    end
    result
  end
```

This will give you a result like this:
`{"london"=>0, "south_east"=>4, "yorkshire_and_the_humber"=>0, "north_west"=>0, "east_of_england"=>3, "west_midlands"=>0, "north_east"=>0, "south_west"=>0, "east_midlands"=>0}`

Then in the view you can iterate over this hash, using the keys ("london" etc) as the IDs for the table rows, and the values for the contents of the table cells. You can also get the region strings `I18n.t("project.region.london")` from the keys.

You could create a hash like this for each permutation of statistic, e.g. Inside the loop, call the scopes you want statistics for, `@projects.by_region(key).voluntary.count` etc

PR #618, opened 2023-03-30
Add region to Projects as a first-class attribute

lozette:
## Changes

Store the region on a project as a first-class attribute (as opposed to something we access via the project's establishment).

When a project is created, fill its region value using the formobject. The value is not manually updated by the user.

Add a migration to migrate all existing projects to have a region, by calling the Academies API, getting the project's establishment and accessing its region value.

## Checklist


mec (commented):
What do you think about setting the region in the create project form object?

The validation already fetches the establishment so a small change would get us access to it.

My reasoning is that this keeps the model about 'storing' only and the form object about setting the right values to store.

I am hoping we can extract the 'create a project' code out of the form object and into a service like object soon whcih would fit this approach well. The hope is by seperating the create a new project from the form, we are setup for creating the project based on some trigger from prepare.

This way definately works and is valid though, happy to chat about this.

app/services/project_region_migrator.rb
```diff
@@ -0,0 +1,12 @@
+class ProjectRegionMigrator
+  def initialize(project)
+    @project = project
+  end
+
+  def migrate_up!
+    if @project.region.nil?
+      establishment = @project.establishment
+      @project.update!(region: establishment.region_code)
```
mec, line 9: I would be tempted to use `save(validate: false)` on this, we've been stung before where a project is not valid so the data migration fails, I feel like we know this 'region' is correct so we can skip validation as part of this - be intereseted in your thoughts?

app/controllers/conversions/voluntary/projects_controller.rb
```diff
@@ -50,7 +50,8 @@ def create
       :note_body,
       :assigned_to_regional_caseworker_team,
       :directive_academy_order,
-      :sponsor_trust_required
+      :sponsor_trust_required,
+      :region
```
mec, line 54: Do we need to permit this if it never comes in via the form or is not massed assigned?

app/forms/conversion/involuntary/create_project_form.rb
```diff
@@ -12,7 +12,8 @@ def save
       regional_delivery_officer_id: user.id,
       task_list: Conversion::Involuntary::TaskList.new,
       directive_academy_order: directive_academy_order,
-      sponsor_trust_required: sponsor_trust_required
+      sponsor_trust_required: sponsor_trust_required,
```
mec, line 15: This is fine, but I would consider all involuntary code frozen from here on as we will strip it out soon.

spec/factories/academies_api/establishment.rb
```diff
@@ -9,6 +9,7 @@
     phase { "Secondary" }
     diocese_name { "Diocese of West Placefield" }
     region_name { "West Midlands" }
+    region_code { "F" }
```
mec, line 12: I wish we could also use the friendly names here as well? I wonder if there is a way to map them?

mec (changes requested):
I am hoping we can make a couple of tweaks to this to make dealing witht he regions as simple as we can for the future - let me know what you think?

spec/factories/academies_api/establishment.rb
lozette: This is a mock up of how the codes are output by the API, so I don't think it's appropriate to use the friendly name here?

app/controllers/conversions/voluntary/projects_controller.rb
lozette: Good point, don't think we need to

config/locales/project.en.yml
```diff
@@ -134,6 +134,16 @@ en:
     notifications:
       completed_html: <h3>Project completed<h3><p>This project was completed on %{date} and cannot be changed.</p>
       not_assigned_html: <h3>Not assigned to project</h3><p>This project is not assigned to you and cannot be changed, you can add notes or contacts if required.</p>
+    region:
```
mec, line 137: ❤️

spec/factories/conversion/project_factory.rb
```diff
@@ -12,6 +12,7 @@
     assigned_to { association :user, :caseworker, email: "user.#{SecureRandom.uuid}@education.gov.uk" }
     directive_academy_order { false }
     sponsor_trust_required { false }
+    region { Project.regions["london"] }
```
mec, line 15: Sneaky! You went with something then?!

mec (approved):
🚢

config/locales/project.en.yml
lozette: Thought I'd add them now as we will definitely need them later!

PR #624, opened 2023-04-03
View a list of projects where the conversion date has changed

mec:
This work covers the concept of 'slipped' projects. Projects that were going to convert in a given month but have since changed their conversion date.

The complexity here is in the scope for these projects, there is an extensive commit message there, but in summary:

- for all in progress projects
- for a given month and year
- get the conversion date histories ordered by creation date
- get the last conversion date history item
- if the previous_date in that item is the same month and year as the given ones, included it

Much of the work is already done with existing scopes, the main work comes from the join to the converison date histories.

With the scope in place, we add a simple view to the application to render matching projects, we'll use this view to see how many projects match the criteria and test this with real world data.

Screenshot 2023-04-03 at 10-56-38 Complete conversions transfers and changes

### DIFF
--- a/spec/features/projects/conversion_date_changed/viewing_projects_with_a_revised_converison_date_spec.rb
+++ b/spec/features/projects/conversion_date_changed/viewing_projects_with_a_revised_converison_date_spec.rb
@@ -1,0 +1,26 @@
+require "rails_helper"
+
+RSpec.feature "Viewing projects with a revised conversion date" do
+  scenario "Users can view the projects that were due to convert for a given month and year" do
+    user = create(:user)
+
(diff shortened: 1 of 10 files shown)


spec/features/projects/conversion_date_changed/viewing_projects_with_a_revised_converison_date_spec.rb
```diff
@@ -0,0 +1,26 @@
+require "rails_helper"
+
+RSpec.feature "Viewing projects with a revised converison date" do
```
lozette, line 3: Small typo `conversion`

lozette (approved):

PR #625, opened 2023-04-03
Handover task is optional

mec:
As we work to make the voluntary project route work for both voluntary
and sponsored converisons we will need the 'Handover with regional
delivery officer' task optional so that when a regional delivery officer
is completed the a sponsored conversion (which is currently always)
they can skip this task.

https://trello.com/c/vcdEuEY0

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_03_30_125747) do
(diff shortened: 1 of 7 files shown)


db/schema.rb
```diff
@@ -284,6 +285,7 @@
     t.date "provisional_conversion_date"
     t.boolean "directive_academy_order", default: false
     t.boolean "sponsor_trust_required", default: false
+    t.string "region"
```
lozette, line 288: Oop this doesn't belong here

lozette (changes requested):

lozette (approved):

PR #643, opened 2023-04-12
Users can view a project's local authority details from the project information tab

beeLorna:
## Changes

This creates a Local Authority model and importer which imports the local authority data into the database.
We then display the project's local authority address within the project information tab from this imported data, along with the local authority name which we still get from the Establishmennt via the Academies API.

I have included the `local_authorities.csv`, to allow easy access to the local authorities data in the future.

To run the importer: `bundle exec rake local_authorities:import["db/seeds/local_authorities.csv"]`
Those using zsh will need to escape the square brackets, like so: 
`bundle exec rake local_authorities:import\["db/seeds/local_authorities.csv"\]`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - new MembersApi Client to call the Parliamentary Members API and retrieve MP
   details
 - show the Member of Parliament details for a school on a new page
+- Users can view a project's local authority details from the project
+  information tab
 
 ### Changed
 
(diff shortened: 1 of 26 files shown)


lozette (approved):
Approved. But can we add instructions on how to run the importer somewhere? Even if it's just in the commit message. I figured out it was `bundle exec rake local_authorities:import["fixtures/local_authorities.csv"]` but it would be good to have that clearly written down.

mec:
@beeLorna it's fine in this case as local authority addresses do not contain personally identifiable information, but let's be super careful when commiting data like in this work! 👍

mec (approved):
Great work @beeLorna! 🚢

PR #649, opened 2023-04-18
(Fix) Remove nil warning in tests

lozette:
This test was testing for the presence of nil, which generated a warning each time the tests were run. Swap the expectation for something better.

```
Checking for expected text of nil is confusing and/or pointless since it will
always match. Please specify a string or regexp instead.
```


mec (approved):
🚢

PR #658, opened 2023-04-20
Raise error if the Academies API returns 401 Unauthorised

lozette:
## Changes

https://sdd-n7.sentry.io/issues/3912539267/?project=6684508&referrer=slack

We have had a few instances where our access to the Academies Api has been
revoked. This has resulted in the application returning a generic 500
error.

If our access has been revoked, raise a specific error with an informative page.
This will alert us to the situation more quickly, so we can
resolve it with the Academies Api maintainers.

## Checklist


mec (approved):
:ship:

PR #661, opened 2023-04-21
(Bug) User can add duplicate projects

lozette:
## Changes

A very subtle bug occurred with the CreateProject form, allowing two projects
with the same URN & UKPRN combination to be created.

This happened because the duplicate project had a nil `assigned_to` value, which
meant it did not appear in the `Project.in_progress` scope.

Adding a new `Project.not_completed` scope to use instead for this validation
has fixed the issue. A project which has no `assigned_to` value is not considered
`in_progress` because no-one will actively be working on it.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased][unreleased]
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


lozette:
[Bug: User can add duplicate projects under certain conditions](https://trello.com/c/ryjQrAdS/1496-bug-user-can-add-duplicate-projects-under-certain-conditions)

mec (approved):
:ship:

PR #670, opened 2023-04-25
Fix wrong notification banner when user not found

mec:
A helpful user pointed out that when their email is not registered in
the applcation and they attempt to sign it, the notification banner
title is 'success' even though the action clearly fails.

Here we switch this notificaiton to the 'important' notification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Address subtle bug in Project creation, which meant two projects with the same
   URN were created in error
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Address subtle bug in Project creation, which meant two projects with the same
   URN were created in error
+- Notificaition banner when users email not recognised is no longer a success
```
lozette, line 13: Typo `Notification`

lozette (approved):

PR #679, opened 2023-04-27
Add AZURE_REDIRECT_URI env var

lozette:
## Changes

In 901c1d4587d56c64c4027b7f03052d80fcd55439 we added a new AZURE_REDIRECT_URI var to support that the app is now behind a proxy on Azure.

This env var is required for the app to run locally. Add it to the Dotenv required keys, and dev & test env example files.

## Checklist


mec (approved):
I really want to explain whtat this is - but that is likely on me so let's add this and I will do some words soonish! 🚢

PR #693, opened 2023-05-03
Add all Local Authority actions

beeLorna:
## Changes

All Local Authority records can be listed and viewed individually.

We have added a new user role `service_support`, this user can add, edit and delete local authority records in the application.

Other users can only list and view the records.

Screenshot 2023-05-10 at 10 45 57

Screenshot 2023-05-10 at 10 46 05

Screenshot 2023-05-10 at 10 46 15

Screenshot 2023-05-10 at 10 46 24

Screenshot 2023-05-10 at 10 46 39

## Checklist


app/models/local_authority.rb
```diff
@@ -1,6 +1,7 @@
 class LocalAuthority < ApplicationRecord
   validates :name, presence: true
   validates :code, presence: true
+  validates :code, uniqueness: true
```
mec, line 4: 👍

mec (approved):
🚢 I am assuming we are adding a role and policy to control this next?

lozette:
@mec Yes, I was going to add it all in one PR, I didn't realise it wasn't a draft PR 😮‍💨 

I'll request a re-review when I've added it

db/schema.rb
```diff
@@ -315,6 +315,7 @@
     t.string "last_name"
     t.string "active_directory_user_id"
     t.boolean "caseworker", default: false
+    t.boolean "service_support"
```
mec, line 318: Does it makes sense for this to be false by default?

mec (changes requested):

mec (approved):
🚢

PR #695, opened 2023-05-04
Users can provide the new Academy URN

lozette:
### Users can edit the Academy URN

Users can now edit a project to add an Academy URN. 

From the "All conversions without an academy URN" page (`/projects/all/new`) they can click on a link to add the Academy URN. This brings them to a form where they can enter the Academy URN.

Screenshot 2023-05-04 at 14 42 06

If they do not enter anything, they see an error message:

Screenshot 2023-05-04 at 14 43 35

If the Academy URN does not exist in GIAS, they see an error message:

Screenshot 2023-05-04 at 14 43 49

If the Academy URN is found in GIAS, they see a confirmation message:

Screenshot 2023-05-04 at 14 43 15

Upon confirmation, they are returned to the list page, with a success message:

Screenshot 2023-05-04 at 14 44 08

### New listing page

The "All conversions without an academy URN" now has a sister page featuring all projects with an academy URN. The two pages are linked by tabs.

Screenshot 2023-05-04 at 14 41 46

Screenshot 2023-05-04 at 14 41 55

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - a new view that shows regional delivery officers the projects they have
   created
+- User can provide the new academy urn once it has be created in Get information
+  about schools
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   created
 - User can provide the new academy urn once it has be created in Get information
   about schools
+- Added a new list of all projects with Academy URNs
 
 ### Changed
 
--- a/app/controllers/conversions/academy_urn_controller.rb
+++ b/app/controllers/conversions/academy_urn_controller.rb
@@ -1,0 +1,58 @@
+class Conversions::AcademyUrnController < ApplicationController
+  rescue_from Api::AcademiesApi::Client::NotFoundError, with: :not_found_error
+
+  def edit
+    @project = Project.find(params[:project_id])
+    authorize @project
+  end
+
+  def check
+    @project = Project.find(params[:project_id])
+    authorize @project
+
+    if academy_urn.blank?
+      @project.errors.add(:academy_urn, :blank)
+      return render :edit
+    end
+
+    @project.assign_attributes(academy_urn: academy_urn)
+    unless @project.valid?
+      @project.errors.add(:academy_urn, :invalid_urn)
+      return render :edit
+    end
+
+    result = fetch_academy_details(academy_urn)
+    @academy_urn = academy_urn
+    @establishment = result
+
+    render "confirm"
+  end
+
+  def update_academy_urn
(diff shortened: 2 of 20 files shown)


lozette:
[Users can provide the new Academy URN](https://trello.com/c/APpvqMcC/1508-users-can-provide-the-new-academy-urn)

mec:
This is looking really nice! Couple of things I think from spinning it up:

- all views should be in a two-thirds column layout expect tables
- we should use the `urn_validator` on the urn entry form as you can add anything
- we need to add a permission for updating only the urn that the service support user can do without being assigned to the project - I don't think we have this `User` role yet (it is coming in the work Lorna has) so for now let any user update the `academy_urn`
- I am struggling a bit with the content, but I think perhaps we come back to that later?

lozette:
@mec Done the required amendments and fixed up the relevant commits 👍

app/controllers/conversions/academy_urn_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    result = fetch_academy_details(academy_urn)
+    @academy_urn = academy_urn
+    @establishment = result
+
+    render "confirm"
+  end
+
+  def update
```
mec, line 31: I think we need to give this a specific name so we can allow this update to happen when the user is not `assigned_to`, right now I can get stuck trying to update the academy urn for a project I am not assigned to - if we call this `update_academy_urn` or something, we can add a specific permission for it?

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Users can edit the Academy URN on a project
```
mec, line 11: As these are primarily aimed at end users could we make this a bit more specific? Something like:

> User can provide the new academy urn once it has be created in Get information about schools

app/views/conversions/academy_urn/confirm.html.erb
```diff
@@ -0,0 +1,39 @@
+<%= render partial: "projects/shared/signed_in_as", locals: {email_address: current_user.email} %>
```
mec, line 1: We should really just delete this and stop propagating it!! Not something you should do here though!

app/views/projects/shared/_with_academy_urn_table.html.erb
```diff
@@ -0,0 +1,38 @@
+<% if projects.empty? %>
+  <%= govuk_inset_text(text: t("project.table.with_academy_urn.empty")) %>
+<% else %>
+  <table class="govuk-table" name="projects_table" aria-label="Projects table">
```
mec, line 4: Please don't change this, but we do have the GOVUK view components, which includes tables: https://govuk-components.netlify.app/components/table/#input-erb-building-a-table-row-by-row

mec (changes requested):
Couple of small things. 

The content doesn't work well for me and I don't think for you either. Because this is a big feature that won't go infront of users just yet, I am inclined to ship the behaviour and get the content fix later - what do you think?

mec (approved):
🚢

PR #697, opened 2023-05-09
Remove involuntary conversion code

mec:
This is a first pass at removing the now redundant involuntary conversion code.

We want to do this ahead of introducing the new tasks list models to stop that becoming too arduous.

It was pretty hard to achieve this in reviewable commits without having the test suite go red at some point so I've gone for that over a green suite all the way through.

I am confident there is plenty more to do to get rid of all traces of involuntary, but we also should:

- be looking to refactor 'voluntary conversion' to simply 'conversion'
- focus on getting the new task list deployed

https://trello.com/c/IZAYbB7R


spec/features/users_can_view_a_voluntary_conversion_task_list_spec.rb
```diff
@@ -1,19 +0,0 @@
-require "rails_helper"
```
lozette, line 1: This is a voluntary task list spec - did you mean to remove it?

lozette (changes requested):
One query!

spec/features/users_can_view_a_voluntary_conversion_task_list_spec.rb
mec: Nope! Thank you!

lozette (approved):

PR #701, opened 2023-05-10
Check for missing i18n strings when running tests

lozette:
We have had several instances of i18n strings being missing from locale files. This initializer will cause the test suite to report failures if any i18n locales are missing. It will only be run in test, and not cause any issues in dev or production modes.

## Changes

## Checklist


mec (approved):
Not seen this? Seems reasonable though. 🚢

lozette:
I think it will massively help the content people!

PR #714, opened 2023-05-12
Refactore voluntary converison to conversion

mec:
- Remove old task list task views
- Move all conversions voluntary code to conversions
- Update routes for removing voluntary converisons
- Remove unused project new view
- Remove unsued task list example


lozette (approved):
Approved. However there are some locales which can be removed too, in another PR!

mec:
> Approved. However there are some locales which can be removed too, in another PR!

Yeah, there is still more to get rid of - could you make a card and add the things you've spotted please, they are getting harder to find! :)

PR #716, opened 2023-05-15
1584 the error shown on the academy order question on the new project form still references a yes no response

SteveMilnes:
## Changes

This work changes the error message content when a user does not select a response for the question about the type of academy order used for a conversion when creating a project in the application.

Screenshot 2023-05-15 at 11 41 27 am

Screenshot 2023-05-15 at 11 41 04 am

## Checklist


SteveMilnes:
I committed without adding the changes first time round and thought I had lost all the changes when I realised, so did not push and re-did the work. 

lt looks like there are 2 commits in this PR, so I don't know if that needs tidying up at all. But the second commit is the one with changes added, then committed.

PR #720, opened 2023-05-16
Fix the stakeholder kick off task

mec:
Now that the new task list is in place this task no longer needs to support the old way of managing the conversion date changing from proivisional to confirmed.

I also noticed a bug where saving the task created a new `Conversion::DateHistory` everytime rather than only once.

This work removes the old behaviour and fixes the new.


beeLorna (approved):
LGTM 👍

PR #723, opened 2023-05-16
Set default CDN health probe state to disabled

DrizzlyOwl:
When only 1 backend origin is configured on the Azure Front Door CDN, there is a significant number of requests to the health-check path which cause unnecessary load and bandwidth. 

Given there is only 1 node, it is redundant to perform health checks as there are no other origins to load balance to


mec (approved):
🎉 🚢

PR #730, opened 2023-05-18
Add index & edit pages for Directors of child services

lozette:
## Changes

(I would add screenshots but they all contain sensitive information!)

Update the views for Directors of Child Services to match the designs in [this board](https://lucid.app/lucidspark/5f12e637-989c-4ee8-b5af-2225f0fdb417/edit?invitationId=inv_d3f413c4-ff92-4dd2-ac9f-6de2ac03cfcd&page=0_0#)

Add routes at `/directors-of-child-services` to expose a list of Directors, with their local authority information. The directors can then be edited. A director cannot be added or deleted from this page - to add or delete a director, you need to do it from the local authorities page `/local-authorities`, to keep the association between Directors and Local authorities.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Add importer to import Director of Child Services data
+- Show the Directors of child services on their own index page. Allow the
+  director records to be edited from this page (but not added or deleted).
 
 ## [Release 25][release-25]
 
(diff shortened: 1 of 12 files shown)


lozette:
[Director of Child services - UI snags](https://trello.com/c/IfsxShYs/1622-director-of-child-services-ui-snags)

mec:
> to add or delete a director, you need to do it from the local authorities page /local-authorities, to keep the association between Directors and Local authorities.

@lozette because of this - is it worthwhile having the index and edit pages at all?

mec:
For me, because of the 1 to 1 link I think modifying the record from the local authority is enough. Does it just add confusion having this view - can the user achieve the goal without adding this - I think so.

I won't push beyond this, also happy to let the team know, if you are not comfortble doing so.

If you still want to go ahead with this - request my review and I'll do it.

lozette:
Yeah I agree that adding the DCS from the Local authority is fine, but I was going off the Lucidspark board 🤦 

We do need the commit updating the Role text, however, but I can pull that off into a separate commit. I will aos keep the inflection as I think it could be useful.

PR #749, opened 2023-05-24
(Fix) Route to project openers page should be `opening`

lozette:
The route to the project openers page should be `/projects/opening`.

This was missed in an earlier commit.


beeLorna (approved):
🚀

PR #755, opened 2023-05-30
First pass at data export for funding agreement letters

beeLorna:
## Changes

This isn't the finished feature.

This work is to introduce the process of exporting projects from the openers list into a csv file.
This covers just a few basic attributes to allow us to test this against a larger number of projects within the `production environment`.

The are at least 44 attributes that will end up in the csv, so we want to make sure we have a solid foundation to work from.

### Next steps
To add in the remaining attributes once we know this is the sensible route.

## Checklist

### DIFF
--- a/app/controllers/all/opening/projects_controller.rb
+++ b/app/controllers/all/opening/projects_controller.rb
@@ -13,6 +13,19 @@ class All::Opening::ProjectsController < ApplicationController
     @date = "#{Date::MONTHNAMES[month.to_i]} #{year}"
   end
 
+  def download_csv
+    authorize Project, :index?
+
+    @projects = ProjectsFetcher.new.sorted_openers(month, year)
+    @date = "#{Date::MONTHNAMES[month.to_i]} #{year}"
+    @month = month
+    @year = year
+
+    @csv = OpeningProjectsCsvExporter.new(@projects).call
+
+    send_data @csv, filename: "opener.csv", type: :csv, disposition: "attachment"
+  end
+
   private def month
     params[:month]
   end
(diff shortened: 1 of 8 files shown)


app/models/api/members_api/client.rb
```diff
@@ -23,6 +23,13 @@ def constituency(search_term)
     end
   end
 
+  def member_for_constituency(constituency)
+    member_id = member_id(constituency).object
+    member_name = member_name(member_id).object
+    contact_details = member_contact_details(member_id).object.find { |details| details.type_id == 1 }
+    {name: member_name.name_display_as, email: contact_details.email, address_line1: contact_details.line1, address_line2: contact_details.line2, address_line3: contact_details.line3, address_postcode: contact_details.postcode}
```
lozette, line 30: We need to include lines 4 & 5 in the address as I have seen entries from the API where the address is spread over all 5 lines. Yes even when it's all the same address 😆

app/controllers/all/opening/projects_controller.rb
```diff
@@ -13,6 +13,19 @@ def revised
...
+    authorize Project, :index?
+
+    @projects = ProjectsFetcher.new.sorted_openers(month, year)
+    @date = "#{Date::MONTHNAMES[month.to_i]} #{year}"
+    @month = month
+    @year = year
+
+    @csv = OpeningProjectsCsvExporter.new(@projects).call
+
+    send_data @csv, filename: "opener.csv", type: :csv, disposition: "attachment"
```
lozette, line 26: I'd change the filename to have the month & year in it, so that the user can download multiple CSVs and not have to manually amend the filename.

lozette (dismissed):

mec (approved):
🚢

PR #759, opened 2023-05-31
Funding agreement letters contact task

lozette:
## Changes

Add the "Confirm who will get the funding agreement letters" task.

This task was begun previously but this is a reworking & completion of it. 

If a project has external contacts already entered, then they will appear on the task page. The user will then be able to select a contact and assign them as the "funding agreement letters" contact. 

On the external contacts page, the "funding agreement letters" contact will be indicated as per the second screenshot.

Screenshot 2023-06-01 at 11 32 14

Screenshot 2023-06-01 at 11 31 52

If a project has no external contacts already, the user will be invited to add more contacts by some text on the task page, and a link leading to the external contacts page. They can then come back to the task later and assign contacts as above.

Previously, we had anticipated that the user would be adding a contact directly on the task page. This reworking means that only the contact ID is stored in the task data, and all contact details are stored in the contacts table.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Display generated DfE number in project information view for the establishment
 - Service support users now see their own sub-navigation in the application
 - Add the Funding Agreement letters task
+- Indicate the Funding Agreement Letters contact on external contacts page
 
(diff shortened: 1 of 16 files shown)


lozette:
[Refine the contact for funding agreement letters task](https://trello.com/c/dfYtl5B6/1559-refine-the-contact-for-funding-agreement-letters-task)

PR #760, opened 2023-06-01
Add some missing locale strings

lozette:
When looking into https://trello.com/c/6toiEBHZ we found some other missing locale strings, which were being overlooked by the i18n default behaviour (of presenting the local key in Sentence Case).


mec (approved):
🚢

PR #762, opened 2023-06-02
Service support user views

mec:
This work brings the three service support user views togther and redirects them to it upon login or visiting the root page.

We haven't consolidate the paths just yet, that may come later, for now we want to have the option of sharing the behaviour of the new app/service with users.

https://trello.com/c/692dTdzf

image


lozette (approved):
Approved. 

One non-blocking question, do we need to add policies to prevent "other" users viewing these pages or is the info not sensitive enough to warrant that?

PR #763, opened 2023-06-02
New project logic

mec:
When a new project is created we can help users by pre-marking some tasks as not applicable:

- when the school is not considered 'religious' the Church supplemental funding agreement task is likely to be not applicable
- when the project is considered 'voluntary' the Process the sponsored support grant task is likely to be not applicable

Here we add this logic into the create conversion project form object.

Along the way we looked for a reliable marker for a school being considered 'religious' - there wasn't a clear one so we have gone with if the school has a diocese - this may turn out to be not the best marker - but this feature is not critical so it is good enough for now.

https://trello.com/c/YvCZ2meO

### DIFF
--- a/app/forms/conversion/create_project_form.rb
+++ b/app/forms/conversion/create_project_form.rb
@@ -146,7 +146,7 @@ class Conversion::CreateProjectForm
       assigned_at: assigned_at,
       directive_academy_order: directive_academy_order,
       region: region,
-      tasks_data: Conversion::TasksData.new
+      tasks_data: new_tasks_data
     )
 
     return nil unless valid?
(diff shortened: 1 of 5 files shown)


lozette (approved):

app/forms/conversion/create_project_form.rb
```diff
@@ -162,12 +162,18 @@ def save
...
     )
   end
 
   private def church_supplemental_agreement_not_applicable?
     return true unless establishment.has_diocese?
     false
   end
+
+  private def sponsored_support_grant_not_applicable?
+    return true if directive_academy_order == false
```
lozette, line 176: Could probably shorten this to `!directive_academy_order` but it might be less clear

PR #769, opened 2023-06-06
Basic header navigation

mec:
Whilst we have not fully fleshed out all the navigation for the site, we really need to be getting users access to all the features of the application.

We are confident in certain groupings of projects and that we will start things off by using the MOJ design system ahead of the DfE one, which is not in a fit state right now.

The header navigation is the top level and we think a good start will be the concept of:

- your projects
- team projects
- all projects
- service support

These loosley relate to user roles we have already. Not all users see all items, you can see this in the navigation policy.

The visual style is a mix of the MOJ and GOVUK header nav, details in the commit.

To bring this together, we also rationalise the service support views so they can be identified and included under the 'service support' header nav.

The main motivation here is:

- introduce the header navigation
- allow service support users to move between their specific views and more generic 'project' focused tasks

image


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- the urls that some service support views were found have changed.
```
lozette, line 16: This is in the wrong commit!

mec: Thank you. 👍

lozette (approved):

PR #770, opened 2023-06-06
(Fix) Redirect the user if they visit `/projects`

lozette:
## Changes

Going to `/projects` was resulting in a 500 error because the app did not know what to do with that route. The route isn't part of any navigation but it is not inconceivable for a user to visit the path.

Add in redirection (identical to the home controller) to take a lost user to the correct place if they visit that path.

## Checklist


lozette:
[Bug: `/projects` throws an error](https://trello.com/c/NeIinfHG/1626-bug-projects-throws-an-error)

mec (changes requested):

PR #775, opened 2023-06-07
Create a bare bones Transfer::Project and Transfer::TasksData

lozette:
## Changes

Add a new Project type, Transfer::Project. It uses the ProjectPolicy.

Add a new Transfer::TasksData model, which uses a new table `transfer_tasks_data` and the TaskListPolicy.

Add factories for the project and tasks data, and prove that new models can be created using them.

## Checklist

### DIFF
--- a/db/migrate/20230607103526_create_transfer_tasks_data.rb
+++ b/db/migrate/20230607103526_create_transfer_tasks_data.rb
@@ -1,0 +1,7 @@
+class CreateTransferTasksData < ActiveRecord::Migration[7.0]
+  def change
+    create_table :transfer_tasks_data, id: :uuid do |t|
+      t.timestamps
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


lozette:
[Add basic model for a Transfer](https://trello.com/c/xjOt9y6L/1568-add-basic-model-for-a-transfer)

db/migrate/20230607103526_create_transfer_tasks_data.rb
```diff
@@ -0,0 +1,7 @@
+class CreateTransferTasksData < ActiveRecord::Migration[7.0]
+  def change
+    create_table :transfer_task_lists, id: :uuid do |t|
```
mec, line 3: Can the table be `transfer_tasks_data` then we don't need the table_name? We need to update the conversion one as well.

mec (changes requested):

mec (approved):
🚢

PR #790, opened 2023-06-12
Move three conversion-specific attributes from Project to Conversion::Project

lozette:
## Changes

Now that Transfer::Project exists, we need to start moving the Conversion-specific attributes and scopes (etc) from Project to Conversion::Project.

The first three are:

- `academy_urn`
- `conversion_date_provisional`
- `conversion_date`

All three have been moved to Conversion::Project, and any scopes which rely on these attributes have also moved to Conversion::Project.

The scope `.in_progress` has stayed in Project, but its ordering by `conversion_date` has been removed. We imagine the concept of "in progress" will apply to all Project types, hence retaining this scope. 

A version of the `.in_progress` scope has been added to Conversion::Project, and retains the ordering by `conversion_date`. 

## Checklist

### DIFF
--- a/app/controllers/all/opening/projects_controller.rb
+++ b/app/controllers/all/opening/projects_controller.rb
@@ -2,7 +2,7 @@ class All::Opening::ProjectsController < ApplicationController
   def confirmed
     authorize Project, :index?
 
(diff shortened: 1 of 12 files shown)


mec (approved):
🚢

PR #791, opened 2023-06-12
Store active directory group ids on user

mec:
We don't want to have to manage user authrorisation ourselves.

When a user is authenticated against the DfE Active Directory, it returns an array of the Group IDs the user is part of. We can leverage this authorise users.

The first step is to store the IDs with the User, we do so here.

We store all the Group IDs in the same column and serialize the result, later we can check which groups we care about.

https://trello.com/c/W1w259AY/1641-store-the-active-directory-group-ids-with-the-user

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -16,7 +16,7 @@ class SessionsController < ApplicationController
   end
 
   def delete
-    session.destroy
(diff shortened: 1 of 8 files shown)


lozette (approved):
Approved but one non-blocking suggestion: update the user factory to reflect this change to the user model?

PR #794, opened 2023-06-13
Fix: Different provisional/confirmed conversion dates is not a change

mec:
Previously we stored the provvisional covnersion date in a seperate
column. We then removed that and simply mark the project as 'with
provisional conversion date'.

However, we missed the impact this has on the
`self.conversion_date_revised_from` scope. Now the first Converison Date
History  is always the change from provisional to confirmed and, if the
dates are diffrent, the test is triggered and the project included when
it should not be.

The fix is to exclude any project that only has one
Conversion::DateHistory when grouping. A project in this state must only
have been confirmed and all projects must be confirmed.

(https://trello.com/c/UpLORrCE)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - A user who has no role is now redirected to the all projects section of the
   application.
+- The list of projects with a revised conversion date for the month and year no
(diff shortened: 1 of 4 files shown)


lozette (approved):
I'm glad you tackled this one as I looked at the ticket and was like 🤯

mec:
> I'm glad you tackled this one as I looked at the ticket and was like 🤯

You would have smashed it - don't be too hard on yourself!

PR #796, opened 2023-06-14
ADR: Model groups in Active Directory

mec:


thill-dxw (approved):
Looks fine. Succinct and to the point.

Given the number of groups in the T1 AD, this adds further urgency to the task of reviewing groups upon migration to a new tenant. But I think that's beyond the scope of this ADR proper.

thill-dxw (approved):

mec:
It is taking too long to get a decisio on which Acitve Directory group to use for this, we are changing approach on this for the momennt. Closing this for now.

PR #806, opened 2023-06-22
Add a BOM to the CSV file and enforce UTF-8

lozette:
## Changes

The CSV was not displaying correctly in MS Excel, due to UTF-8 encoding not being enforced. Enforce UTF-8 via a BOM prepended to the CSV data (byte order mark) and encoding the CSV file as it is generated.

Screenshot below to show the apostrophe in "Children's services" is being rendered correctly in Excel.

Screenshot 2023-06-22 at 11 33 08

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Enforce UTF-8 encoding in the CSV download via a BOM (byte order mark) and by
+  generating the CSV file in UTF-8
+
 ## [Release 29][release-29]
 
 ### Added
--- a/app/services/opening_projects_csv_exporter.rb
+++ b/app/services/opening_projects_csv_exporter.rb
@@ -8,7 +8,7 @@ class OpeningProjectsCsvExporter
   end
 
   def call
-    @csv = CSV.generate(headers: true) do |csv|
+    @csv = CSV.generate("\uFEFF", headers: true, encoding: "UTF-8") do |csv|
       csv << headers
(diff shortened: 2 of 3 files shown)


lozette:
[Help Excel handle special characters in funding agreement letter export: The BOM](https://trello.com/c/wunKigRJ/1761-help-excel-handle-special-characters-in-funding-agreement-letter-export-the-bom)

app/controllers/all/opening/projects_controller.rb
```diff
@@ -23,7 +23,7 @@ def download_csv
 
     @csv = OpeningProjectsCsvExporter.new(@projects).call
 
-    send_data @csv, filename: "opening_#{month}_#{year}.csv", type: :csv, disposition: "attachment"
+    send_data "\uFEFF" + @csv, filename: "opening_#{month}_#{year}.csv", type: :csv, disposition: "attachment"
```
mec, line 26: Are we able to but this in a method in the `OpeningProjectsCsvExporter`? It feels a little icky here, it's so unclear what it is doing, wrapping it in a helpfully named method would help a lot.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Enforce UTF-8 encoding in the CSV download via a BOM (byte order mark)
```
mec, line 18: I don't think we are enforcing UTF-8 - but we should! It probably happens inside `OpeningProjectsCsvExporter`: `CSV.generate`

mec (changes requested):

mec (approved):
🚢

PR #808, opened 2023-06-23
All projects by trust

mec:
Trust relationship managers need a view of projects across the product for the trust that they manage, this work delivers it.

We build a list of trusts that has not compelete projects, we use not complete becuase in-progress excludes unassigned projects and we expect these users to want to see those projects.

The service that builds the data for the table doesn't test the `not_complete` scope, as this is covered seperately.

The table of trusts is paged as we expect the number of trusts to be around the same number of projects, which is too high for a single view.

I was torn about testing for the pager, it would involve a lot of fake trusts to make the numbers and there does not seemt to be a way to overide the number of items per page to make the testing simpler?

Whilst we generally have to use the UKPRN for the code, we believe the Group ID is a more useful identifier of trusts for users.

https://trello.com/c/trdI9uMh

image


app/services/by_trust_project_fetcher_service.rb
```diff
@@ -0,0 +1,42 @@
...
+      trusts = get_trusts(conversion_counts.keys)
+
+      sort_trust_objects_by_name(build_trust_objects(trusts, conversion_counts))
+    else
+      []
+    end
+  end
+
+  private def conversion_count_by_trust
+    projects = Project.not_completed
```
lozette, line 15: I'd make this `Conversion::Project`, just so we don't have to circle back later when we start using Transfer projects.

lozette (changes requested):

app/services/by_trust_project_fetcher_service.rb
mec: Nice!! 👍

lozette (approved):

PR #816, opened 2023-06-26
Opening export handles no academy

mec:
We are still ironing out how this export will work, but as lots of
projects have no academy we should guard against nil.

## Changes

## Checklist


lozette (approved):
🤦

PR #818, opened 2023-06-26
Add `team` to User

lozette:
## Changes

Add a team attribute to User. Possible values are taken from an enum of regions + three other non-region teams.

## Checklist

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -16,6 +16,22 @@ class User < ApplicationRecord
 
   validates :email, format: {with: URI::MailTo::EMAIL_REGEXP}
 
+  enum :team, {
+    london: "london",
+    south_east: "south_east",
+    yorkshire_and_the_humber: "yorkshire_and_the_humber",
+    north_west: "north_west",
+    east_of_england: "east_of_england",
+    west_midlands: "west_midlands",
+    north_east: "north_east",
+    south_west: "south_west",
+    east_midlands: "east_midlands",
+    regional_casework_services: "regional_casework_services",
+    service_support: "service_support",
+    academies_operational_practice_unit: "academies_operational_practice_unit",
+    education_and_skills_funding_agency: "education_and_skills_funding_agency"
+  }, suffix: true
+
   def full_name
     "#{first_name} #{last_name}"
   end
--- a/db/migrate/20230626112559_add_team_to_user.rb
+++ b/db/migrate/20230626112559_add_team_to_user.rb
@@ -1,0 +1,5 @@
+class AddTeamToUser < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :team, :string
+  end
+end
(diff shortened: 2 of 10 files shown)


lozette:
[Add team to User](https://trello.com/c/c9dJXj3a/1765-add-team-to-user)

app/models/user.rb
```diff
@@ -16,6 +16,22 @@ class User < ApplicationRecord
 
   validates :email, format: {with: URI::MailTo::EMAIL_REGEXP}
 
+  enum :team, [
+    :london,
```
mec, line 20: How would you feel about declaring and storin the enum as strings? I prefer it as wherever you are you know what the value means, rather than having to convert an integer to the right value?

It would looke like this:
```
{
  london: "london",
   ...
}
```

and the column would need to be a string.

app/models/user.rb
```diff
@@ -16,6 +16,22 @@ class User < ApplicationRecord
...
+    :east_of_england,
+    :west_midlands,
+    :north_east,
+    :south_west,
+    :east_midlands,
+    :regional_casework_services,
+    :service_support,
+    :academies_operational_practice_unit,
+    :education_and_skills_funding_agency
+  ], suffix: true
```
mec, line 33: ❤️ the suffix here.

mec (changes requested):

app/models/user.rb
lozette: Yep can do, agree it would be more easily readable. I didn't think to do it because the values would be the same.

db/migrate/20230626112559_add_team_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddTeamToUser < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :team, :string
```
mec, line 3: I should have raised this earlier - sorry!

We can populate the `user.team` for some users and should do so, I wrote it for `project.team` but missed it for this work! 😞 

- When `user.caseworker?` is true, `user.team == "regional_casework_services"`
- When `user.team_leader?` is true `user.team == "regional_casework_services"`
- When `user.service_support?` is true `user.team == "service_support"`

We can do this once this gets deployed if you prefer that option?

mec (changes requested):
Sorry @lozette - I missed the data migration for this work.

db/migrate/20230626112559_add_team_to_user.rb
lozette: I can add it here! I actually forgot to add it to the Projects one as well 😆

mec: Projects one is on another card so you can leave it: https://trello.com/c/NB8mmAFA

mec (approved):
I would have called the service from the migration - but this works! 🚢

lozette:
Doh why didn't I think of that 😆

PR #823, opened 2023-06-27
Update Project.team

lozette:
## Changes

Add a service to populate the `team` attribute on Projects. If the project has `assigned_to_regional_casework_team` set to true, set its team to be `regional_casework_services`. 

Otherwise set its team to the the same as its region.

## Checklist


lozette:
[Update Project.team values](https://trello.com/c/NB8mmAFA/1769-update-projectteam-values)

mec (approved):
I feel like the database transaction should be on the enumerator rather than the individual update object - that said, I am not worried as this is a one shot data migration, so let's go for it. 🚢

PR #827, opened 2023-06-28
Deployment workflow

DrizzlyOwl:
## Changes

Consolidates multiple GitHub deployment workflows into a single streamlined one.

_NB. This will remain a draft until all of the repository secrets have been migrated from their legacy names to the new names to ensure no blocks are introduced to the deployment process_


mec:
This looks right enough. Can we hold off on deleting the old actions until the new one is validated?

When do we think the secrets will be migrated?

You need to run Prettier: `yarn run lint:format:fix`

DrizzlyOwl:
- **Can we hold off on deleting the old actions until the new one is validated?**
  - I have dropped the commit that removes the old workflows from this PR. We should be mindful to disable the old ones before merging this PR in otherwise there will 2x deployments to Dev which could collide.
- **When do we think the secrets will be migrated?**
  - This is waiting for a ServiceNow request, so will take a short amount of time. Should be ready before the end of the week.

- ✅ You need to run Prettier: `yarn run lint:format:fix`
  - Prettier wants to rewrite the GitHub action yml files but the expected changes would break the syntax. Can we include an ignore pattern for `.github/` folder?

DrizzlyOwl:
@mec I have now finished getting the secrets set up. Quick question:
>Prettier wants to rewrite the GitHub action yml files but the expected changes would break the syntax. Can we include an ignore pattern for .github/ folder?

mec (approved):
Let's try this. 🚢

PR #835, opened 2023-07-03
Allow CSV export to go ahead when there are data issues with the members API

lozette:
## Changes

Because of data inconsistencies between GIAS and the Parliamentary Members API,
we have had issues getting the CSV export to work properly when a school's
parliamentary constituency is unclear (e.g. "Middlesbrough" is not a single
constituency, but GIAS insists some schools are in "Middlesbrough")

To get round this, if we query the Members API for a constituency but get
multiple results back, still allow the CSV to be downloaded but with empty
values for the MP details. The CSV requester will then have to fill in the
blanks themselves.

We also send a warning to Application Insights to let us know this has happened. This way we
can track which constituencies are returning multiple results.

The solution for this is a little complex in the CSV exporter, due to the way
the Members details objects are set up, but this should capture everything.
## Checklist

### DIFF
--- a/app/models/api/members_api/client.rb
+++ b/app/models/api/members_api/client.rb
@@ -24,7 +24,14 @@ class Api::MembersApi::Client
   end
 
   def member_for_constituency(constituency)
-    member_id = member_id(constituency).object
+    member = member_id(constituency)
+
+    if member.error.present?
+      track_error(member.error.message)
+      return nil
+    end
+
+    member_id = member.object
     member_name = member_name(member_id).object
     contact_details = member_contact_details(member_id).object.find { |details| details.type_id == 1 }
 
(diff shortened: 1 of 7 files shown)


lozette:
[When the members API returns multiple constienies in the search we should make it into a nil](https://trello.com/c/eQRrj61t/1803-when-the-members-api-returns-multiple-constienies-in-the-search-we-should-make-it-into-a-nil)

app/models/api/members_api/client.rb
```diff
@@ -24,7 +24,14 @@ def constituency(search_term)
   end
 
   def member_for_constituency(constituency)
-    member_id = member_id(constituency).object
+    member = member_id(constituency)
+
+    if member.error.present?
+      Sentry.capture_message(member.error.message, level: :warning)
```
mec, line 30: We should probably try and use Application Insights instead of Sentry here: https://github.com/Microsoft/ApplicationInsights-Ruby#sending-a-simple-event-telemetry-item

mec (changes requested):

mec (approved):
🚢

PR #836, opened 2023-07-03
Basic in application user management

mec:
This is a first pass at adding basic user management to the application itself. We are doing this becuase the long term plan is the the Service Support team take over responsibility of adding new users to the applcation.

It is worth noting, we tried to leverage the Active Directory for this purpose, but the decision was made not to pursue that path for a number of reasons.

This work updates the User model which breaks a bunch of specs, we chose to fix those in seperate commits to keep reviewing more paletable at the cost of the suite going red briefly.

With the User model validating `team` as required, we won't need the data migration tasks and services (which have been run in production anyway) so we delete them as making the specs pass on these would not have been possilble, as they work on the basis of `team` being nil.

Most of the work here is standard Rails. We decided not to put the routes at `/service-support` for now.

https://trello.com/c/UGB3Qo3w

image

image

image


config/locales/user.en.yml
```diff
@@ -1,8 +1,63 @@
...
+      yorkshire_and_the_humber: Yorkshire and the Humber
+      north_west: North West
+      east_of_england: East of England
+      west_midlands: West Midlands
+      north_east: North East
+      south_west: South West
+      east_midlands: East Midlands
+      regional_casework_services: Regional Casework Services (RCS)
+      service_support: Service Support
+      academies_operational_practice_unit: Acadmeies Operational Practice Unit (AOPU)
```
lozette, line 48: Typo `Acadmeies`

lozette (changes requested):

lozette (approved):

PR #837, opened 2023-07-04
Fix the default opening date routes

mec:
Previously we used a default set of parameters in the routes file, but
this was only called once during Rails initialisation (we think) and so
was locked to a single month the application was deplyed.

Here we switch to a redirect based on the current month and year, me
moved the path helpers as we don't actually use the specific month and
year paths excpet from the redirect.

https://trello.com/c/sCpYdjeR


beeLorna (approved):
🚀 LGTM!

PR #838, opened 2023-07-04
Enforce date ordering in Project index pages

lozette:
## Changes

We recently moved the `conversion_date` attribute onto Conversion::Project, and as a result any Project index pages using plain Project objects lost the expected date ordering.

Make all the project index pages use Conversion::Project, and add a test to make sure the date ordering is working as expected.

TODO: Figure out how these controllers will work when we have more than one project type.

## Checklist


lozette:
[Project views should be sorted by conversion date](https://trello.com/c/kAopLrEH/1802-project-views-should-be-sorted-by-conversion-date)

mec (approved):
🚢

PR #839, opened 2023-07-04
Replace branch name with SHA reference

DrizzlyOwl:
- Previously this was set to the branch name which meant the az cli command to restart the container did not do anything because the image tag had not changed. Changing this to be the 'SHA' means that az cli will consider it a changed image for every deployment


mec (approved):
🚢

PR #841, opened 2023-07-04
Always use 'development' build flag

DrizzlyOwl:
the `RAILS_ENV` needs to be set to `development` during the build. This is then overridden at runtime by the appropriate value based on the environment it is deployed into.


mec (approved):
I don't understand why this is the fix, but happy to give it a try.

PR #850, opened 2023-07-06
DfE frontend

mec:
This work addes the DfE Frontend to the application - we add it from the dist soruce as the npm package does not work with the version of Node we run.

We've tried to use overrides on GOVUK Frontend to keep things as modular as possible.

We updated the footer - although there is no guidance, it seems logical that the contents of the app is not under the Open Government Licence (OGL).

https://design.education.gov.uk/design-system/dfe-frontend

image


lozette:
@mec The bugs that SonarCloud are reporting here are legitimate - they are duplicated CSS declarations, I think they should be addressed.

mec:
> @mec The bugs that SonarCloud are reporting here are legitimate - they are duplicated CSS declarations, I think they should be addressed.

My problem is that they are in source of the DfE frontend, and I want to stay modular (like we do with the GOVUK frontend) i.e. that should be fixed upstream we just use it without modification.

lozette:
@mec Aaah fair enough, it's not great that they have duplicated CSS rules in that case! 🤦

lozette (approved):

PR #853, opened 2023-07-07
Dynamic team projects view

lozette:
## Changes

Scope the Team projects views to reflect the current user's team.

If the user's team is `regional_casework_services`, then they see all projects
there the Project.team is `regional_casework_services`

If the user's team is a region, then they see all projects where Project.region
is the same as the user's team (note that it is the project's REGION).

Change the navigation within the Team projects view so that only team_leader
users can view the Unassigned projects view.

All users with a role and a team can now view the Team projects views.

## Checklist

### DIFF
--- a/app/controllers/team/projects_controller.rb
+++ b/app/controllers/team/projects_controller.rb
@@ -4,33 +4,17 @@ class Team::ProjectsController < ApplicationController
 
   def in_progress
     authorize Project, :index?
-    @pager, @projects = pagy(Project.assigned_to_regional_caseworker_team.in_progress.includes(:assigned_to))
-
-    pre_fetch_establishments(@projects)
-    pre_fetch_incoming_trusts(@projects)
+    @pager, @projects = pagy_array(ByTeamProjectFetcherService.new(current_user.team).in_progress)
   end
 
   def completed
     authorize Project, :index?
-    @pager, @projects = pagy(Project.assigned_to_regional_caseworker_team.completed)
-
-    pre_fetch_establishments(@projects)
-    pre_fetch_incoming_trusts(@projects)
+    @pager, @projects = pagy_array(ByTeamProjectFetcherService.new(current_user.team).completed)
   end
 
   def unassigned
-    authorize Project, :index?
-    @pagy, @projects = pagy(Project.assigned_to_regional_caseworker_team.unassigned_to_user)
+    authorize Project, :unassigned?
 
-    pre_fetch_establishments(@projects)
-    pre_fetch_incoming_trusts(@projects)
-  end
-
-  private def pre_fetch_establishments(projects)
-    EstablishmentsFetcherService.new(projects).call!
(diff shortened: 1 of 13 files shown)


lozette:
[Dynamic team projects view](https://trello.com/c/MRizvpW2/1789-dynamic-team-projects-view)

app/services/by_team_project_fetcher_service.rb
```diff
@@ -0,0 +1,35 @@
+class ByTeamProjectFetcherService
+  def initialize(user)
+    @team = user.team
+  end
+
+  def in_progress
+    return [] if @team.nil?
+
+    if @team.eql?("regional_casework_services")
+      Project.assigned_to_regional_caseworker_team.in_progress.includes(:assigned_to)
```
mec, line 10: We are going to need to sort all these by `conversion_date` which means they all need to be `Conversion::Project`.

app/services/by_team_project_fetcher_service.rb
```diff
@@ -0,0 +1,35 @@
+class ByTeamProjectFetcherService
+  def initialize(user)
```
mec, line 2: I wonder if it should just initialize with a team?

app/controllers/team/projects_controller.rb
```diff
@@ -4,33 +4,17 @@ class Team::ProjectsController < ApplicationController
 
   def in_progress
     authorize Project, :index?
-    @pager, @projects = pagy(Project.assigned_to_regional_caseworker_team.in_progress.includes(:assigned_to))
-
-    pre_fetch_establishments(@projects)
```
mec, line 9: Loosing the pre fetching will make these views really slow. I would add the pre fetchers to the service, you could have method that fetches and returns the projects, something like:

```
private def pre_fetch_establishments_and_trusts(projects)
   EstablishmentsFetcherService.new(projects).call!
  TrustsFetcherService.new(projects).call!
  projects
end
```

mec (changes requested):

app/controllers/team/projects_controller.rb
lozette: Ah yes I saw you'd done that on the ConversionProjectsFetcherService 👍

app/services/by_team_project_fetcher_service.rb
```diff
@@ -0,0 +1,46 @@
+class ByTeamProjectFetcherService
+  def initialize(team)
+    @team = team
+  end
+
+  def in_progress
+    return [] if @team.nil?
+
+    projects = if @team.eql?("regional_casework_services")
+      Conversion::Project.assigned_to_regional_caseworker_team.in_progress.includes(:assigned_to)
```
mec, line 10: These just need the `by_conversion_date` scope and we should be good!

mec (changes requested):

PR #859, opened 2023-07-11
Users can be deactivated and reactivated

mec:
This works allows users to be deactivated. A deactivated user:

- cannot access the application
- does not appear in list of users when assgining projects
- may have done these things in the past

We store the `deactivated_at` date and time and wrap it with a `active` method so we can treat it like a boolean in the interface.

https://trello.com/c/g6tlmXfG

image


app/models/user.rb
```diff
@@ -33,6 +33,18 @@ def has_role?
     false
   end
 
+  def active
+    deactivated_at.nil?
+  end
+
+  def active=(value)
```
lozette, line 40: Oh this is nice

lozette (approved):

app/models/user.rb
mec: Oh cool! I wasn't sure! 👍

PR #863, opened 2023-07-12
Add team projects by user view

lozette:
## Changes

Add a "Users" page to the Team projects area, showing a list of all users on the current_user's team, with a list of their conversion project counts (ordered by user name).

Screenshot 2023-07-12 at 10 13 24

The current_user can then click through to view a list of user's projects. This view reuses the table view from the Your projects area, as it is essentially the same data (projects assigned to a certain user). 

Screenshot 2023-07-12 at 10 13 33

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - User accounts can now be deactivated which prevents the user from accessing
   the application or from showing up in user selections, inactive accounts can
   be reactivated.
+- Add a "Users" view to the Team projects area, to view other users on your
+  team, with their conversion project counts.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   be reactivated.
 - Add a "Users" view to the Team projects area, to view other users on your
   team, with their conversion project counts.
+- Add a "by user" view to the Team projects area, where a user can view projects
+  assigned to another user on the same team.
 
 ### Changed
 
--- a/app/controllers/team/projects_controller.rb
+++ b/app/controllers/team/projects_controller.rb
@@ -17,4 +17,10 @@ class Team::ProjectsController < ApplicationController
 
     @pager, @projects = pagy_array(ByTeamProjectFetcherService.new(current_user.team).unassigned)
   end
+
+  def users
+    authorize Project, :index?
+
+    @pager, @users = pagy_array(ByTeamProjectFetcherService.new(current_user.team).users)
+  end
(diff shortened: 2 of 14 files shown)


lozette:
[Add Team projects > by user view](https://trello.com/c/yoIPh9zf/1771-add-team-projects-by-user-view)

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   view.
 - Add a "Users" view to the Team projects area, to view other users on your
   team, with their conversion project counts.
+- Add a "by user" view to the Team projects area, where a user can view projects
```
mec, line 17: Honestly I wouldn't worry about this - I see no value in preventing this until we find a reason?

app/views/shared/navigation/_team_projects_primary_navigation.html.erb
```diff
@@ -13,6 +13,8 @@
 
           <%= render partial: "shared/navigation/primary_navigation_item", locals: {name: t("navigation.primary.team_projects.completed"), path: completed_team_projects_path, namespace: "/projects/team/completed"} %>
 
+          <%= render partial: "shared/navigation/primary_navigation_item", locals: {name: t("navigation.primary.team_projects.by_user"), path: users_team_projects_path, namespace: "/projects/team/user"} %>
```
mec, line 16: This should go before 'completed' to be consistent with the All projects nav.

mec (changes requested):
I don't think we need the last commit on here - unless I am missing something? I would rather keep access open and flexible until we know there is a reason not to?

CHANGELOG.md
lozette: Do you want me to drop this whole commit or just the permissions part?

app/controllers/team/projects_controller.rb
```diff
@@ -23,4 +23,25 @@ def users
...
   end
+
+  def by_user
+    authorize Project, :index?
+
+    user_id = params[:user_id]
+    @user = User.find(user_id)
+    @pager, @projects = pagy(Conversion::Project.assigned_to(@user)
+                                                .in_progress
+                                                .by_conversion_date, items: 10)
```
mec, line 34: The pager doesn't show up on the view we are using, which is not part of this work, but as a general rule we don't include `items: 10` on tabular views, it defaults to 20.

I'll make a not to add a pager to this view later. If we drop items here, we can crack on!

mec (changes requested):
Noticed two things - one to fix, another for later.

app/controllers/team/projects_controller.rb
lozette: Oh that's odd, I copied this `page: 10` from another view (can't remember which). I'll remove it.

mec (approved):
:ship

PR #877, opened 2023-07-14
Add confirmed and revised project openers views, filtered by team

lozette:
## Changes

Add a tab for Project openers to the Teams view. 

This tab is split into two pages - confirmed openers for a given month by team, and revised openers for a given month by team.

These pages closely replicate the existing Openers pages, with the only difference that they are filtered by the current_user's team.

## Checklist

### DIFF
--- a/app/controllers/all/local_authorities/projects_controller.rb
+++ b/app/controllers/all/local_authorities/projects_controller.rb
@@ -14,8 +14,8 @@ class All::LocalAuthorities::ProjectsController < ApplicationController
   end
 
   private def projects_for_local_authority(local_authority_code)
+    EstablishmentsFetcherService.new(Project.not_completed).batched!
     projects = Project.not_completed.includes(:assigned_to)
-    EstablishmentsFetcherService.new(projects).batched!
     projects.to_a.select { |p| p.establishment.local_authority_code == local_authority_code }
   end
 
--- a/app/services/by_month_project_fetcher_service.rb
+++ b/app/services/by_month_project_fetcher_service.rb
@@ -14,6 +14,21 @@ class ByMonthProjectFetcherService
     sort_by_conditions_met_and_name(projects)
   end
 
+  def confirmed_openers_by_team(month, year, team)
+    projects = if team.eql?("regional_casework_services")
+      Conversion::Project.assigned_to_regional_caseworker_team.opening_by_month_year(month, year).includes(:tasks_data)
+    else
+      Conversion::Project.not_assigned_to_regional_caseworker_team.by_region(team).opening_by_month_year(month, year).includes(:tasks_data)
+    end
+
+    if @prefetch
+      pre_fetch_establishments(projects)
+      pre_fetch_trusts(projects)
+    end
+
+    sort_by_conditions_met_and_name(projects)
+  end
+
   private def sort_by_conditions_met_and_name(projects)
     projects.sort_by { |p| [p.all_conditions_met? ? 0 : 1, p.establishment.name] }
   end
--- a/app/services/by_month_project_fetcher_service.rb
+++ b/app/services/by_month_project_fetcher_service.rb
@@ -1,4 +1,4 @@
-class ConversionProjectsFetcherService
+class ByMonthProjectFetcherService
   def initialize(prefetch: true)
     @prefetch = prefetch
   end
--- a/app/services/by_month_project_fetcher_service.rb
+++ b/app/services/by_month_project_fetcher_service.rb
@@ -29,6 +29,21 @@ class ByMonthProjectFetcherService
     sort_by_conditions_met_and_name(projects)
   end
 
+  def revised_openers_by_team(month, year, team)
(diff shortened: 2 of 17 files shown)


lozette:
[Add Team projects > Opening view](https://trello.com/c/fZTEyWiU/1823-add-team-projects-opening-view)

app/views/shared/navigation/_team_projects_primary_navigation.html.erb
```diff
@@ -15,6 +15,8 @@
 
           <%= render partial: "shared/navigation/primary_navigation_item", locals: {name: t("navigation.primary.team_projects.completed"), path: completed_team_projects_path, namespace: "/projects/team/completed"} %>
 
+          <%= render partial: "shared/navigation/primary_navigation_item", locals: {name: t("navigation.primary.team_projects.opening"), path: confirmed_team_opening_projects_path, namespace: "/projects/team/opening"} %>
```
mec, line 18: This should go next to "In progress" so it mirrors the All projects view.

app/services/by_month_project_fetcher_service.rb
```diff
@@ -14,6 +14,19 @@ def sorted_openers(month, year)
     sort_by_conditions_met_and_name(projects)
   end
 
+  def confirmed_openers_by_team(month, year, team)
+    users = User.by_team(team)
+
+    projects = Conversion::Project.assigned_to_users(users).opening_by_month_year(month, year).includes(:tasks_data)
```
mec, line 20: If we add a `by_team` scope to Project could we avoid the extra DB query here? It is ultimately the projects that are in the same team as the current user and we have the `current_user`?

It would then be:

`Conversion::Project.by_team(current_user.team).opening_by_month_year(month, year).includes(:tasks_data)`

app/services/by_month_project_fetcher_service.rb
```diff
@@ -27,6 +27,19 @@ def confirmed_openers_by_team(month, year, team)
     sort_by_conditions_met_and_name(projects)
   end
 
+  def revised_openers_by_team(month, year, team)
+    users = User.by_team(team)
+
+    projects = Conversion::Project.assigned_to_users(users).conversion_date_revised_from(month, year).includes(:tasks_data)
```
mec, line 33: I've hit the bullet issue you had straight away:

`ActiveRecord::EagerLoadPolymorphicError in Team::Opening::ProjectsController#confirmed`

It feels like the issue is that `in_batches` cannot cope with our super new use of delegated types? The existing tests use doubles, yours uses ActiveRecord objects, which might explain the different results?

Given how expensive this view is already, I am not worried about eager loading the tasks_data here and would be happy to exlude it via the Bullet config - we can have a chat about this if you like, it's pretty late in the day and I might be missing something!

mec (changes requested):
Happy to chat about this - this is possibly are most complicated view, give me a yell!?!

app/services/by_month_project_fetcher_service.rb
lozette: *Is* a project always in the same team as the user, though? When we add a project and/or assign a user to a project, are we updating the project's team? I'm not convinced.

lozette: Or are you talking about a project's *region* matching the user's team?

mec: Great point!! Let me think it over in the morning when my brain might be working!

mec: Right, brain freshley caffinated and working as well as can be expected! ☕ 

I think as ever, the answer to this is: it depends!

The logic is something like this:

```
if user.regional_casework_services_team?
  Projects.by_team(team)
else
  Projects.by_region(team)
end
```

The RCS team want to see:

> all projects that are assigned to them across all regions

A region team wants to see:

> all projects in their region across all teams

lozette: I think I've fixed this 🤞

app/controllers/all/local_authorities/projects_controller.rb
```diff
@@ -14,8 +14,8 @@ def show
   end
 
   private def projects_for_local_authority(local_authority_code)
+    EstablishmentsFetcherService.new(Project.not_completed).batched!
```
mec, line 17: I feel like this is just fetching the establishments for no reason becuse the next line gets another set of projects from the DB and passes that along - I am not 100% on this though.

Regardless, I think it's good enough for now - this whole prefetching thing is fast becoming a real issue that we should take a step back and try to solve.

mec (approved):
As a workaround, I think I can live with this - but we need to resolve it long term, let's get this out and we can write something about rationalising all this prefecthing spaghetti or may be loose it altogether.

app/controllers/all/local_authorities/projects_controller.rb
lozette: That's pretty much what it's doing 😆

PR #883, opened 2023-07-18
Limit line lengths to 75 characters in 1200px app

anthonyomalley:
## Changes

This change adjusts line lengths within the application to approximately 75 characters. It comes in response to the expansion of the app's width from 1024px to 1200px. By constraining the line lengths, the update ensures optimal readability and consistency across the wider layout.

The class is adapted from the Design System documentation app. Adaptations include limitations to checkbox label and detail component texts to ensure consistency with the 1200px app layout.

### Before

Screenshot 2023-07-18 at 16 05 21

### After

Screenshot 2023-07-18 at 16 05 33

### To test

Go to tasks views from within the task list.

## Checklist

- [X] Attach this pull request to the appropriate card in Trello.
- [X] Update the `CHANGELOG.md` if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Limit line lengths to 75 characters after increasing app width to 1200px
 - Only show counts of in-progress projects on the Team "By user" projects page
 - Send the new user email when an account is added.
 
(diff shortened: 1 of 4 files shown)


.bundle/config
```diff
@@ -1,2 +1,3 @@
 ---
 BUNDLE_BUILD__TINY_TDS: "--with-freetds-include=/opt/homebrew/include --with-freetds-lib=/opt/homebrew/lib"
+BUNDLE_PATH: "vendor/bundle"
```
mec, line 3: I don't think this should be here?

mec: If you need or want your bundle in the app directory, you can add ~/.bundle/config and put it there I think.

mec (changes requested):

.bundle/config
anthonyomalley: Cool, just removed that 👍

lozette (approved):

mec (approved):

PR #884, opened 2023-07-18
Add handed over view to team projects

beeLorna:
## Changes

This work adds in a `Handed over` view to the Team projects area which displays the projects that have been handed over to the Regional casework services team.
Each regional team can view this `Handed over` view and will see all their handed over regional projects.
The view is restricted to the Regional casework services team who won't see the `Handed over` tab.

## Checklist

### DIFF
--- a/app/controllers/team/projects_controller.rb
+++ b/app/controllers/team/projects_controller.rb
@@ -19,7 +19,11 @@ class Team::ProjectsController < ApplicationController
   end
 
   def handed_over
-    authorize Project, :handed_over?
+    begin
+      authorize Project, :handed_over?
+    rescue Pundit::NotAuthorizedError
+      return head(:not_found)
+    end
 
     @current_users_team = current_user.team
     @pager, @projects = pagy(ByRegionProjectFetcherService.new.regional_casework_services_projects(@current_users_team))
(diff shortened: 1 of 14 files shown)


beeLorna:
[Add handed over view to Team projects](https://trello.com/c/2VfBfLp8/1905-add-handed-over-view-to-team-projects)

app/controllers/team/projects_controller.rb
```diff
@@ -17,4 +17,11 @@ def unassigned
 
     @pager, @projects = pagy_array(ByTeamProjectFetcherService.new(current_user.team).unassigned)
   end
+
+  def handed_over
+    authorize Project, :handed_over?
+
+    @current_users_team = current_user.team
+    @pager, @projects = pagy_array(ByRegionProjectFetcherService.new.regional_casework_services_projects(@current_users_team))
```
mec, line 29: Now that it is available, could you call the new `AcademiesApiPreFetcherService` here to make this view feel responsive:

`AcademiesApiPreFetcherService.new.call!(@projects)`

This is the pattern we use now anytime we have a paged view of projects.

app/controllers/team/projects_controller.rb
```diff
@@ -19,7 +19,11 @@ def unassigned
   end
 
   def handed_over
-    authorize Project, :handed_over?
+    begin
+      authorize Project, :handed_over?
+    rescue Pundit::NotAuthorizedError
+      return head(:not_found)
```
mec, line 25: I've not seen this `head` before, something about this makes me wonder if there was another approach, but this works - so let's use it.

mec (changes requested):
Not you fault as it came later, but could you add the prefetcher and then this is good to go!

mec (approved):

PR #888, opened 2023-07-20
Academies api pre fetcher

mec:
Our approach to batch fetching from the Academies API to speed up the application has grown organically, this has caused us some problems with various things including eager loading and paging.

This work sets out to resolve this by offering a single object that takes care of it all with a simple method call.

Because this object can be passed ActiveRecord results with eager loading, we have to disbale Bullet inside it to prevent false positives caused by bullet thinking we are not using the the eager loaded association, when we are just passing them on to what does use them.

We continue with the idea of 'dependancy injection' and allow the pre fetching to be diisabled to allow testing of any class that utilising pre fetching without having to deal with the pre fetching as it can make mocking really difficult.

This object can also handle any number of projects and splits them up into batches of 10 it also works when passed to Pagy, this resolves the two main issues we have faced previously.

Here we only switch the currently broken 'Team projects' - with the intention of verifying the behaviour and then switching over to this new object everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- The Team projects views now work when there are more than twenty projects.
+
 ## [Release 34][release-34]
 
 ### Added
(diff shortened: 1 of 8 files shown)


lozette (approved):
Do itttt

PR #889, opened 2023-07-20
Add DfE Frontend javascript

mec:
This will make the mobile menu work.

I can't see value in a test for this as we would have to launch a whole
browser for this one thing - if you think this is a bad call, please go
ahead and add the spec! We've been running with it broken for weeks now.

## Changes

## Checklist


lozette (approved):
Add a changelog entry so we remember to test it in product review!

PR #892, opened 2023-07-20
Add "2 Requires improvement" (2RI) to Conversion Projects

lozette:
## Changes

Add a new question to the create Conversion project form to indicate if a school is converting because it has received two "requires improvement" ratings. This question is mandatory for conversion projects.

Screenshot 2023-07-20 at 16 45 03

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Add "2RI" question to the new Conversion project form
 - Remove hint from the "DAO" question on the create Conversion project form
+- Show the "2RI" value on the Project information page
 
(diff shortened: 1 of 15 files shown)


db/migrate/20230721101132_update2_ri_values_in_projects.rb
```diff
@@ -0,0 +1,8 @@
+class Update2RiValuesInProjects < ActiveRecord::Migration[7.0]
+  def up
+    Project.all.each do |project|
```
mec, line 3: If the database defaults to false - do we need this migration?

mec (changes requested):
Possibly my bad, but I don't think we need the migration with a default of false!

mec (approved):
🚢

PR #896, opened 2023-07-24
Refactor converison date

mec:
This work refactors the current `conversion_date` into `significant_date` moving it up into Project so that any subclass can use it, i.e. Transfer.

We move all the bechaviour along with it so that users can change the significant date and the reason for the change is recorded.

We use aliases to keep things working at this point, we might decide to always refer to significant date or keep the aliases and let each subclass refer to it's own date as needed - we'll have to see.

Right now, almost all views specifically load `Conversion::Project` but that is set to change as we start to bring transfer projects into the application.

https://trello.com/c/Yn5iRx87


app/controllers/all/opening/projects_controller.rb
```diff
@@ -15,7 +15,7 @@ def confirmed_next_month
   def revised
     authorize Project, :index?
 
-    @projects = ByMonthProjectFetcherService.new.revised(month, year)
+    @projects = Conversion::Project.significant_date_revised_from(month, year)
```
lozette, line 18: Why change this line, the action of `ByMonthProjectFetcherService.new.revised` has changed so it's not necessary

lozette (changes requested):

spec/services/by_month_project_fetcher_service_spec.rb
```diff
@@ -32,14 +32,32 @@
...
+
   describe "#confirmed_openers_by_team" do
     context "with pre fetching disabled for this test" do
       before do
         mock_all_academies_api_responses
       end
 
       let(:user_1) { create(:user, :caseworker, team: "regional_casework_services") }
-      let(:user_2) { create(:user, :caseworker, team: "london") }
+      let(:user_2) { create(:user, :regional_delivery_officer, team: "london") }
```
lozette, line 60: I feel like this change doesn't belong in this commit, it's to do with user factories and not scopes. Did it creep in accidentally?

lozette (approved):
Approved pending fixing the failed test

PR #905, opened 2023-07-27
Frontend updates

mec:
Some small frontend tweaks - mostly making sure eveything is laid out properly.

This does put the task list in a two-thirds column and fixes an issue with some of the project sub views not being rendered at the correct width.

image


beeLorna (approved):
LGTM!  🚀

PR #907, opened 2023-07-28
Users can confirm transfer date

mec:
The goal of this work is to allow the transfer date to be changed in exactly the same way as the conversion date.

To do this we first have to model the provisional date and get it confirmed, which means adding the first bones of the stakeholder kickoff task. We've added a generator, just like the conversion one to automate as much of the creating of the files to create the task.

With that done we show the date on the project summary.

Most of this is moving things around so they can be used across both conversion and transfers.


spec/features/transfers/users_can_create_a_transfer_project_spec.rb
```diff
@@ -13,6 +13,7 @@
     let(:incoming_ukprn) { 10061021 }
     let(:outgoing_ukprn) { 10090252 }
     let(:two_weeks_ago) { Date.today - 2.weeks }
+    let(:two_moths_time) { Date.today + 2.months }
```
lozette, line 16: typo: `two_months_time`

lozette (approved):
Just one typo to fix before merging

beeLorna:
Looks like CI is failing due to SimpleCov coverage and an accessibility check

mec:
> Looks like CI is failing due to SimpleCov coverage and an accessibility check

Yep, it's the rebase onto the frontend changes - I am working on it.

PR #932, opened 2023-08-01
Update your projects

mec:
The main motivation for this work is to be able to show a transfer project on the your in-progress projects views.

We took some time to update the routing and namespacing to align it to 'yours' rather than 'user'.

We also improved the layout of the view as a step towards using the GOVUK summary card pattern.

You'll see we had to disable Bullet for a spec because we load both Conversion and Transfer projects, but only the conversions use the eager loading right now.

https://trello.com/c/lkSWzi4I

image


app/views/your/projects/_converison_card.html.erb
```diff
@@ -0,0 +1,50 @@
+<li>
+  <span class="govuk-caption-m">URN <%= project.urn %></span>
+  <h3 class="govuk-heading-m govuk-heading-m--school-name">
+    <%= link_to project.establishment.name, project_path(project), class: "govuk-link govuk-link--no-visited-state" %>
+  </h3>
+  <dl class="govuk-summary-list">
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">
+        Type
```
lozette, line 9: I'd love for this to be internationalized!

app/views/your/projects/in_progress.html.erb
```diff
@@ -10,14 +10,28 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <% if @projects.empty? %>
+      <%= govuk_inset_text(text: t("project.index.empty")) %>
+    <% else %>
+      <ul class="projects-list list-style-none govuk-!-padding-0">
+        <% grouped_project = @projects.group_by { |project| project.significant_date.strftime("%B %Y") } %>
+        <% grouped_project.each do |opening_date, projects| %>
+          <h2 class="govuk-heading-l"><%= opening_date %></h2>
+          <% projects.each do |project| %>
+            <%= render partial: "converison_card", locals: {project: project} if project.type.eql?("Conversion::Project") %>
```
lozette, line 29: Typo in file name (need to fix file name)

lozette (approved):

PR #947, opened 2023-08-02
Update transfer create form

mec:
We went ahead and added transfer behaviour before we had fully designed some aspects of it (and that is okay!!)

This work brings the new transfer project form up to scratch, using the supplied design for layout and content.

We tweaked a few things in the spririt of 'leave it better'.

https://trello.com/c/9xjVW2bS

image

### DIFF
--- a/app/views/transfers/projects/new.html.erb
+++ b/app/views/transfers/projects/new.html.erb
@@ -7,22 +7,38 @@
     <%= form_with model: @project, url: transfers_path do |form| %>
       <%= form.govuk_error_summary %>
 
-      <h1 class="govuk-heading-l"><%= t("transfer_project.new.title") %></h1>
+      <h1 class="govuk-heading-xl"><%= t("transfer_project.new.title") %></h1>
       <%= t("transfer_project.new.hint_html") %>
 
-      <%= form.govuk_text_field :urn, label: {size: "m", text: t("helpers.label.transfer_project.urn")}, hint: {text: t("helpers.hint.transfer_project.urn").html_safe}, width: 10 %>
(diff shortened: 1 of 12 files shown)


.erblint-rubocop20230802-43724-kcow2j
```diff
@@ -0,0 +1,7 @@
+---
```
lozette, line 1: Don't think this should have been added!

lozette (changes requested):

app/views/transfers/projects/new.html.erb
```diff
@@ -19,6 +19,7 @@
       <%= form.govuk_text_field :incoming_trust_sharepoint_link, label: {size: "m", text: t("helpers.label.transfer_project.incoming_trust_sharepoint_link")} %>
       <%= form.govuk_text_field :outgoing_trust_sharepoint_link, label: {size: "m", text: t("helpers.label.transfer_project.outgoing_trust_sharepoint_link")} %>
 
+      <%= form.govuk_collection_radio_buttons :assigned_to_regional_caseworker_team, @project.assigned_to_regional_caseworker_team_responses, :id, :name, legend: {text: t("helpers.hint.conversion_project.assigned_to_regional_caseworker_team")}, form_group: {id: "assigned-to-regional-caseworker-team"} %>
```
lozette: The locale key on this is for `conversion_project`, for consistency we should add the key to the transfer project locale.

lozette (changes requested):
Sorry one more tiny thing!

lozette (approved):

PR #946, opened 2023-08-02
Add Transfer project counts

lozette:
## Changes

All "By" index views (By user, By trust etc) now include counts for Transfer projects as well as Conversion projects.

The associated "show" views (show projects by user or by trust etc) now include a mix of Trust and Conversion projects.

The table headers have been updated to reflect that "Conversion date" could be Conversion or transfer date, and "School" could be School or academy.

## Checklist

- [x] Attach this pull request to the appropriate card in Trello.
- [x] Update the `CHANGELOG.md` if needed.

PR #948, opened 2023-08-02
Update all table views

beeLorna:
## Changes

On the back of adding `transfer` projects to the service, we need our table views to be able to reflect the two project types they will be displaying.

There was a slight overlap with PR https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/946, so the work around making the `conversion_date` and associated controllers step away from being solely conversion projects now happens within that work. 

This PR builds on that by adding in a new table column `Type` to display whether it is a conversion project or transfer prioject and formats the `conversion or transfer date` using the `significant_date` format.
This PR also updates the table view by removing the `View project` column and adds in the project link to the school name

## Checklist

### DIFF
--- a/app/views/all/local_authorities/projects/show.html.erb
+++ b/app/views/all/local_authorities/projects/show.html.erb
@@ -15,7 +15,8 @@
             <tr class="govuk-table__row">
               <th class="govuk-table__header" scope="col"><%= t("project.table.headers.school_name") %></th>
               <th class="govuk-table__header" scope="col"><%= t("project.table.headers.school_urn") %></th>
-              <th class="govuk-table__header" scope="col"><%= t("project.table.headers.significant_date") %></th>
+              <th class="govuk-table__header" scope="col"><%= t("project.table.headers.conversion_or_transfer_date") %></th>
+              <th class="govuk-table__header" scope="col"><%= t("project.summary.type.title") %></th>
               <th class="govuk-table__header" scope="col"><%= t("project.table.headers.assigned_to") %></th>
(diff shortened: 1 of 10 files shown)


beeLorna:
[Update all table views](https://trello.com/c/nkMoFd04/1996-update-all-table-views)

app/views/all/local_authorities/projects/show.html.erb
```diff
@@ -25,7 +26,8 @@
             <tr class="govuk-table__row">
               <td class="govuk-table__header govuk-table__cell"><%= project.establishment.name %></td>
               <td class="govuk-table__cell"><%= project.urn %></td>
-              <td class="govuk-table__cell"><%= project.significant_date.to_formatted_s(:govuk) %></td>
+              <td class="govuk-table__cell"><%= project.significant_date.to_formatted_s(:significant_date) %></td>
+              <td class="govuk-table__cell"><%= project.type %></td>
```
lozette, line 31: This will output Conversion::Project or Transfer::Project, we want to show just Conversion or Transfer.

Have a look at Project#type_locale - this will return "conversion_project" or "transfer_project" and you can use that to build up a locale key for the correct value.

lozette (changes requested):

config/locales/project.en.yml
```diff
@@ -165,7 +165,8 @@ en:
         assign: Assign project
         added_by: Added by
         conversion_date: Conversion date
-        significant_date: Conversion or transfer date
+        conversion_or_transfer_date: Conversion or transfer date
+        type: Type
```
lozette, line 169: There is already a key `type` in this file, 8 lines up.

PR #959, opened 2023-08-04
Show the Director of child services in the External contacts area

lozette:
## Changes

Move the Director of Child Services contact information into the External contacts area.

Screenshot 2023-08-04 at 09 04 15

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   task to improve readability
 - removed extra the from notifcation of change task and add a transfer form
 - add simple side navigation to the statistics page
+- the Director of Child Services contact information is now in the External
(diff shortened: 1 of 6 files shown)


lozette:
[Show the director of child services in the external contacts](https://trello.com/c/Sr6CHz4G/1977-show-the-director-of-child-services-in-the-external-contacts)

mec (approved):
Cripes! This definately works! For me, this creates a special case of the DCS, which I am not a fan of, I would rather load `Contacts` and render them than see this special case of DCS.

That said, I can't really come up with a good reason other than the system shouldn't encourage this kind of special case, especially when it uses inheritance and it makes me feel icky - so I am deffering to you - I'll approve this and you can merge it or you can side with me and change it! 

P.S. Thanks for fixing the link on the project info!

mec:
We chatted about this - I didn't see all the association complexity - this is good enough! 🚢

PR #960, opened 2023-08-04
Do not send deactivated users any emails

lozette:
## Changes

Do not send Team Leader mails or Assigned To mails if the user is not active.

Theoretically you cannot select an inactive user on the Assigned to page, but
this is "belt & braces".

We don't add this restriction to the new user mailer because in theory a new
user shouldn't be deactivated (maybe??)

## Checklist


mec (commented):
I feel like this make us have to remember to check everytime we send anything? I wonder if there is a way to leverage a call back in ApplicationMailer so that each Mailer doesn't do anything if the user is inactive?

https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-callbacks

mec (approved):
We talked about this in Slack, this seems like the best way! 🚢

PR #964, opened 2023-08-07
Store when a user last created a session

lozette:
## Changes

When a user logs in (creates a session), store a DateTime in the `latest_session` column on the user. This value will update each time a user creates a session.

## Checklist

### DIFF
--- a/spec/helpers/user_helper_spec.rb
+++ b/spec/helpers/user_helper_spec.rb
@@ -1,0 +1,26 @@
+require "rails_helper"
+
+RSpec.describe UserHelper, type: :helper do
+  describe "#last_seen_datetime" do
+    context "with an active user" do
+      it "shows the formatted latest_session date" do
+        user = create(:user, latest_session: DateTime.new(2023, 1, 1, 10, 30, 0, 0))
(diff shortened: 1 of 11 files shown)


lozette:
[Record the last sign in datetime for each session.](https://trello.com/c/JA7zkZKh/1890-record-the-last-sign-in-datetime-for-each-session)

mec (changes requested):
Could we show the last session date and time on the users table at `/users/active` - sorry it wasn't on the story!

spec/helpers/user_helper_spec.rb
```diff
@@ -0,0 +1,26 @@
+require "rails_helper"
+
+RSpec.describe UserHelper, type: :helper do
+  describe "#last_seen_datetime" do
+    context "with an active user" do
+      it "shows the formatted latest_session date" do
+        user = create(:user, latest_session: DateTime.new(2023, 0o1, 0o1, 10, 30, 0o0, 0))
```
mec, line 7: What are those dates (`0o1`?)

mec (changes requested):
Not so much a change, as an explination of those date arguments - no idea what those are?

PR #966, opened 2023-08-07
Add All conditions met? task to Transfers

lozette:
## Changes

This task is identical to the same task for Conversions, including the same content (with references to Conversions changed to Transfers).

There's no need for any migrations here as the Transfer and Conversion projects store their task data in different tables.

## Checklist


mec (commented):
I was expecting the all conditions met attribute to move up into the `Project` model? This works and is fine, I just want to be sure about why we duplicate it across both `TasksData` instead.

I think the rationale for it being on Project is that it is such an important attribute that we want it quickly and simply - I am not wedded to that though as we can (and I think do) delegate it down to TasksData.

PR #974, opened 2023-08-09
Add All conditions met? for Transfers

lozette:
## Changes

Previously, the `all_conditions_met?` value for a Conversion::Project was determined by its
TasksData, on the `conditions_met_confirm_all_conditions_met` task.

Move this value up to be a native attribute on Project. This reduces a lot of
eager loading and makes the Project model simpler, although it does mean that
the "Conditions met" task becomes another "special case" task.

The All conditions met? task for Transfers is identical to the same task for Conversions, including the same
content (with references to Conversions changed to Transfers).

## Checklist

### DIFF
--- a/app/forms/conversion/task/conditions_met_task_form.rb
+++ b/app/forms/conversion/task/conditions_met_task_form.rb
@@ -1,3 +1,15 @@
 class Conversion::Task::ConditionsMetTaskForm < BaseTaskForm
   attribute :confirm_all_conditions_met, :boolean
+
+  def initialize(tasks_data, user)
+    @tasks_data = tasks_data
+    @user = user
+    @project = @tasks_data.project
+    super(@tasks_data, @user)
+    self.confirm_all_conditions_met = @project.all_conditions_met
+  end
+
+  def save
+    @project.update(all_conditions_met: confirm_all_conditions_met)
+  end
 end
--- a/app/forms/transfer/task/conditions_met_task_form.rb
+++ b/app/forms/transfer/task/conditions_met_task_form.rb
@@ -1,0 +1,15 @@
+class Transfer::Task::ConditionsMetTaskForm < BaseTaskForm
+  attribute :confirm_all_conditions_met, :boolean
+
+  def initialize(tasks_data, user)
+    @tasks_data = tasks_data
+    @user = user
+    @project = @tasks_data.project
+    super(@tasks_data, @user)
+    self.confirm_all_conditions_met = @project.all_conditions_met
+  end
+
+  def save
+    @project.update!(all_conditions_met: confirm_all_conditions_met)
+  end
+end
--- a/config/locales/transfer/tasks/conditions_met.en.yml
+++ b/config/locales/transfer/tasks/conditions_met.en.yml
@@ -1,0 +1,13 @@
+en:
+  transfer:
+    task:
+      conditions_met:
+        title: Confirm all conditions have been met
+        hint:
+        guidance_link: How to check all conditions have been met
+        guidance:
+        confirm_all_conditions_met:
+          title: Confirm all conditions are met
+          hint:
+          guidance_link: What to do if conditions are not met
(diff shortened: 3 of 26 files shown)


lozette:
[All conditions met for both project types](https://trello.com/c/UUKN5fqt/2012-all-conditions-met-for-both-project-types)

db/migrate/20230809083044_add_all_conditions_met_to_project.rb
```diff
@@ -0,0 +1,39 @@
...
+      Conversion::Project.all.each do |project|
+        tasks_data = project.tasks_data
+        all_conditions_met = tasks_data.conditions_met_confirm_all_conditions_met
+        project.all_conditions_met = all_conditions_met
+        project.save(validate: false)
+      end
+    end
+  end
+
+  def revert_all_projects
```
mec, line 29: Once we are confident the up works, we might re-think the down.

If the up has worked, and it kind of has to now, we know that project.all_conditions_met will be either true or false, so can we skip the transactions and checks here?

Is it something like (untested - just food for thought):

```
def revert_all_projects
  Conversion::Project.all.each do |project|
        all_conditions_met = project.all_conditions_met || false
        project.update_attribute(:all_conditions_met, all_conditions_met)
      end 
end
```

config/locales/transfer/tasks/conditions_met.en.yml
```diff
@@ -0,0 +1,27 @@
+en:
```
mec, line 1: I would leave off all but the absolute basic of content for this as I don't think it will be the same as conversions, like we did for the stakholder kick off task, coming back and adding the content later, when it's been done.

app/forms/transfer/task/conditions_met_task_form.rb
```diff
@@ -0,0 +1,14 @@
+class Transfer::Task::ConditionsMetTaskForm < BaseTaskForm
+  attribute :confirm_all_conditions_met, :boolean
+
+  def initialize(tasks_data, user)
+    @tasks_data = tasks_data
+    @user = user
+    @project = @tasks_data.project
```
mec, line 7: I don't think `confirm_all_conditions_met` is being initialized here so the ui never updates. The test you have passes, but I suspect if you go back to the task page an assert the box is ticked, it will fail.

This is also the case on Conversions.

Something like:

```
  def initialize(tasks_data, user)
    @tasks_data = tasks_data
    @user = user
    @project = @tasks_data.project
    super(@tasks_data, @user)
    self.confirm_all_conditions_met = @project.all_conditions_met
  end
```

The assignment happens after the super class has been setup.

app/forms/conversion/task/conditions_met_task_form.rb
```diff
@@ -1,3 +1,14 @@
 class Conversion::Task::ConditionsMetTaskForm < BaseTaskForm
   attribute :confirm_all_conditions_met, :boolean
+
+  def initialize(tasks_data, user)
+    @tasks_data = tasks_data
+    @user = user
+    @project = @tasks_data.project
+    super(@tasks_data, @user)
```
mec, line 8: See comment on Transfers version of this.

app/forms/transfer/task/conditions_met_task_form.rb
```diff
@@ -0,0 +1,14 @@
...
+    super(@tasks_data, @user)
+  end
+
+  def save
+    @project.update(all_conditions_met: confirm_all_conditions_met)
```
mec, line 12: I would bang this (`update!`) as a rule always bang the save methods inside form objects, so you raise loudly if something is wrong.

mec (changes requested):
You did say 'fine tooth comb'! 

This is not quite there yet and I went over board on the migration - happy to chat anything through or pair - just let me know. 👍

lozette:
I think I got them all 🤞

PR #988, opened 2023-08-14
(Fix) Enable Transfer projects to be assigned to people & teams

lozette:
## Changes

We found a big with assigning Transfer projects to people and teams: the AssignmentsController was expecting parameters with a `conversion_project` key. Obviously, with transfer project,s this key will be `transfer_project`.

This solution doesn't feel very elegant but does work. We will probably have to think of a more elegant solution as we add more project types.

## Checklist


beeLorna (approved):
👍

PR #989, opened 2023-08-14
Bring the flash messages inline with GOV.uk styles

lozette:
## Changes

The flash banner should be in its own two-thirds layout.

The banner should be styled as per the GOV.uk guidelines. The main message
must be an h3, and any other text in a paragraph.

We have added a helper to parse the flash messages correctly. If a message only
has one line (no HTML), the helper encapsulates the message in an h3 tag. If
a message contains HTML, we output the HTML in its entirety. 

We determine if a message has HTML by parsing it with Nokogiri and counting
the number of node children. A non-zero number means there is HTML present.

We will have to
ensure any HTML added to flash messages via local files has the correct HTML
format.

We only currently have one multi-line flash message, and the HTML for this has
been updated in the locale file.

https://design-system.service.gov.uk/components/notification-banner/

## Checklist

### DIFF
--- a/app/helpers/notification_helper.rb
+++ b/app/helpers/notification_helper.rb
@@ -1,0 +1,12 @@
+require "nokogiri"
+
+module NotificationHelper
+  def notification_banner_html(message)
+    parsed_message = Nokogiri.parse(message)
+    if parsed_message.children.count > 0
+      message.html_safe
+    else
+      content_tag(:h3, message.html_safe, class: "govuk-notification-banner__heading").html_safe
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Every page now has a unique page title
 
+### Changed
+
+- Bring the flash banners in line with GOV.uk styles
+
 ## [Release 38][release-38]
 
 ### Added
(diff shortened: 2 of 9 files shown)


lozette:
[Update the notification banner](https://trello.com/c/CTtOA0uo/1593-update-the-notification-banner)

app/helpers/notification_helper.rb
```diff
@@ -0,0 +1,12 @@
+require "nokogiri"
+
+module NotificationHelper
+  def notification_banner_html(message)
+    parsed_message = Nokogiri.parse(message)
```
beeLorna, line 5: What is nokogiri doing here? And whats the reasoning behinding using nokogiri? 
More for my understanding than anything else as I haven't seen this before

lozette: I needed a way of telling if the flash message contains any HTML or not. If it doesn't, we can assume it's a one-line message and wrap it in an h3. 

If it does, it's a bit more complex - the GOVuk standard expects the first line of a flash message to be in an h3, and any other lines in paragraphs. 

So I'm using Nokogiri to tell if there's any HTML, by parsing the message and counting its children. If there's HTML, we output the flash message as-is, without wrapping it in an h3. The only wrinkle here is that we will then need any HTML in the flash messages to match the GOVuk standard (h3 for the first line, para for subsequent lines).

I couldn't think of a better way of telling if a flash message is single or multi line!

I did consider just seeing if the string `include?` any tags (e.g. `message.include?("<")`, but that would mean a message like "This value is < that value" would be interpreted as HTML.

PR #990, opened 2023-08-14
Add transfer project statistics to statistics page

beeLorna:
## Changes

Now that we have transfer projects within the service, we want to be able to differentiate between Conversion projects and Transfer projects within our statistics page.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Every page now has a unique page title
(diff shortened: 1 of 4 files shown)


beeLorna:
[Add project type counts to statistics page](https://trello.com/c/0vDcCeni/2011-add-project-type-counts-to-statistics-page)

lozette (approved):
Approved. I think this could do with being refactored in the future to make it less "wordy" though. Each table could be represented by an array of values maybe? Much like we've created an array for the `opener_date_and_project_total`, we could build each table on the page using a similar array/hash structure. It would reduce the number of lines in the implementation and test.

PR #992, opened 2023-08-15
Every page has a unique page title

lozette:
## Changes

The new page titles consist of the `h1` content already on the page, plus the
service name. 

We use a new `#page_title` helper to build the page title string, formed from
the passed-in string and the service name.

The page title is passed into a yielding block in the application base HTML
template.

## Checklist


beeLorna (approved):
👍

PR #1000, opened 2023-08-17
(Fix) Fix alignment of notification banner on Sign in page

lozette:
The notification banner was mistakenly not put into its own `govuk-grid-row` div, which meant on some pages the `h1` content was floating next to the notification banner.

Move the grid row and column divs into the Notification banner partial, so that they only appear when a flash is present.

Move the sign-in content into a grid row and column div to preserve the correct alignment.

Screenshot 2023-08-17 at 11 27 52

## Checklist


mec (approved):
🚢

PR #1009, opened 2023-08-21
Validate that a transfers incoming trust is not the same as its outgoing trust

beeLorna:
## Changes

A transfers incoming trust should not be the same as its outgoing trust.
This work sets a validation to check that these UKPRNs do not match, and if they do to throw an error

## Checklist

### DIFF
--- a/app/forms/transfer/create_project_form.rb
+++ b/app/forms/transfer/create_project_form.rb
@@ -11,6 +11,8 @@ class Transfer::CreateProjectForm < CreateProjectForm
 
   validates :outgoing_trust_sharepoint_link, presence: true, sharepoint_url: true
 
+  validate :check_incoming_trust_and_outgoing_trust
+
   def initialize(params = {})
     @attributes_with_invalid_values = []
     super(params)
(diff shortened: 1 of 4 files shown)


beeLorna:
[Validate transfer incoming > outgoing trust](https://trello.com/c/rjKJ4h5Z/2076-validate-transfer-incoming-outgoing-trust)

app/forms/transfer/create_project_form.rb
```diff
@@ -52,4 +54,8 @@ def provisional_transfer_date=(hash)
   rescue TypeError, Date::Error, NegativeValueError
     @attributes_with_invalid_values << :provisional_transfer_date
   end
+
+  def check_incoming_trust_and_outgoing_trust
+    errors.add(:outgoing_trust_ukprn, I18n.t("errors.attributes.incoming_trust_ukprn.ukprns_must_not_match")) if incoming_trust_ukprn == outgoing_trust_ukprn
```
mec, line 59: If you add the error to `incoming_trust_ukprn` you can do:

`errors.add(:incoming_trust_ukprn, :ukprns_must_not_match) if incoming_trust_ukprn == outgoing_trust_ukprn`

Which also put the error on the field where it makes sense:

> Incoming trust must not be the same as the outgoing trust

mec (changes requested):
See my comment.

app/forms/transfer/create_project_form.rb
beeLorna: Yeah, this makes a lot more sense given the language used 👍

mec (approved):
🚢

PR #1013, opened 2023-08-22
(Fix) Move this locale file to the correct place

lozette:
This local file was in the wrong spot. Doesn't affect the app, but it was not where it was meant to be!


mec (approved):
🚢

PR #1021, opened 2023-08-24
Update the developer scripts

mec:
I am not a fan of the scipts, but I know other developers do like them - that said, they are a mess.

This is a very opinionated attempt to improve things. 

We are shooting for simplification and naming along use and the tasks that developers actually want to use the sciprts for.

An example is the `check_all` script, which runs everything that CI does (almost!) and is a good indicator that your PR will pass.

Along the way we are switching everyone to a 'hybrid' approach to running the application, so the two backing services (DB and Redis) run in Docker with the Ruby (and tiny amount of Node) running natively.

I would like @lozette and @beeLorna to have a look at this branch and let me kwow what they think before we go further.

Other work is happening to remove other scripts that CI currently use.

Hopefully by the end, we'll strike a good balance between helping new developers to the code and those already familiar with it.


doc/developer-scripts.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Running the code analysis
+
+Run [Brakeman](https://brakemanscanner.org/).
+
+`script/static_analysis`
+
+## Start the backing services
+
+You want to have the backing services running but not the application
```
lozette, line 66: Make this a bit more explicit (which backing services?)

lozette (approved):

PR #1027, opened 2023-08-25
(Fix) Correct error message for "role" field on contacts

lozette:
This should be "role" not "title"


lozette:
[Contact validation error message](https://trello.com/c/OmVtUkFc/2130-contact-validation-error-message)

mec (approved):
🚢

PR #1028, opened 2023-08-25
(Fix) Supply a default page title if none is set

lozette:
I really wanted to add a test for this but I can't figure out how to mock out `content_for?` in a feature spec!

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -2,7 +2,7 @@
 <html lang="en" class="govuk-template ">
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


lozette:
[Supply a default page title](https://trello.com/c/XExh473S/2126-supply-a-default-page-title)

mec:
@lozette a view spec? 

https://rspec.info/features/6-0/rspec-rails/view-specs/view-spec/

it looks like the 'View specs can stub a helper method' might work?

mec (approved):
We looked at this and agreed it is not worth the amount of effort to stub everything needed to test this - if we have 100% coverage, let's go with it. 🚢

PR #1029, opened 2023-08-25
Show correct value in data export when `Project.all_conditions_met?` is false

lozette:
## Changes

The presenter was assuming that the value of `Project.all_conditions_met?` would be either "yes" or "nil". In fact it can also be "false", and as a result the export was showing "yes" for a "false" value.

## Checklist


lozette:
[All conditions met on ESFA export is yes even when it is not](https://trello.com/c/XTzWIMZS/2145-all-conditions-met-on-esfa-export-is-yes-even-when-it-is-not)

mec (approved):
🚢

PR #1052, opened 2023-09-04
Remove script from workflow

mec:
The workflows depending on the scripts feels a bit off, whilst it keeps
things in one place, the link is not obvious and as the lint and format
script is the only one we can use in the ci environment, it feels better
to keep the workflows consistent by not running scripts.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -55,7 +55,8 @@ jobs:
       -
         name: Run linters and formaters
         run: |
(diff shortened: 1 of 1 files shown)


mec:
> I think you'll need a `\` at the end of `L58` so that the command remains unbroken when split onto multiple lines

I don't think it matters because we are in the quotes? But I also missed the `&&` 👍

DrizzlyOwl (approved):

PR #1055, opened 2023-09-04
Can we not load the image at build?

mec:
## Changes

## Checklist


mec:
Still cannot get the container name to copy out the file - thought this might work, it doesn't appear to!

PR #1057, opened 2023-09-04
Added named volume mount

DrizzlyOwl:
- Using a named volume mount allows us to persist the files between container runs and extract the files with an ephemeral container
- After the /bin/rspec test has finished and the container exits, you can manipulate the coverage report with other containers
- `docker run --rm -v test-ci-coverage-report:/source -v $PWD:/dest -w /source alpine cp -r . /dest/coverage/` can be used to download the coverage report onto the runner using a shim container


mec (approved):
🚢

PR #1063, opened 2023-09-06
Allow a Project Contact to be selected as the main contact for an establishment

lozette:
## Changes

We want to be able to designate a Project Contact as the "main contact" for an establishment (school).

Users can now mark a contact as the "establishment main contact" when they create or edit an external contact.

ONLY "School" contacts can be establishment main contacts, not any other category of contact.

There can only be one establishment main contact per Project.

If a contact is the establishment main contact, this is shown on the External contacts list page.

### Mark contact as establishment main contact

Screenshot 2023-09-07 at 08 40 08

### Validation

Screenshot 2023-09-07 at 08 40 19

### Show that contact is establishment main contact

Screenshot 2023-09-07 at 08 39 53

## Checklist

### DIFF
--- a/app/models/contact/project.rb
+++ b/app/models/contact/project.rb
@@ -4,4 +4,10 @@ class Contact::Project < Contact
   end
 
   belongs_to :project, optional: true
+
+  def establishment_main_contact
+    return false unless project_id
+    project = ::Project.find(project_id)
+    project&.establishment_main_contact_id.eql?(id)
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added the 'Deed of termination for the church supplemental agreement' task to
   transfers.
 - The application can now store the un-published GIAS establishment records.
+- Allow a contact to be marked as the "establishment main contact"
 
 ### Changed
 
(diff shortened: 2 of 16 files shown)


lozette:
[Add main establishment contact](https://trello.com/c/ZMHB4wgT/2132-add-main-establishment-contact)

app/models/contact/project.rb
```diff
@@ -4,4 +4,10 @@ def self.policy_class
   end
 
   belongs_to :project, optional: true
+
+  def establishment_main_contact
+    return false unless project_id
+    project = ::Project.find(project_id)
```
mec, line 10: This is interesting! I was expecting to be able to just call `project` (the association) on the contact, but that doesn't work - not sure that is right - but not something to fix here, I will make a note to look at the Contact associations.

mec (changes requested):
We should definately use update! and maybe even use a transaction?

Everything else is a comment - thanks for sticking with this - I know it was a slog - now we have a great approach for the trust verstions though! :)

app/models/contact/project.rb
lozette: It doesn't work (I think) because the object itself is a `Project` (a Contact::Project)

lozette: Oh sorry, was typing too early without coffee 🤦  I'm not sure why just calling `project` doesn't work - maybe bc it's optional? There is something odd in these associations tbh.

mec: No problem! I've made a note here to have a look: https://dfe-gov-uk.visualstudio.com/Academies-and-Free-Schools-SIP/_backlogs/backlog/Complete%20conversions%20transfers%20and%20changes/Epics/?workitem=139133

mec (approved):
🚢

PR #1067, opened 2023-09-07
Fix missing line space in readme

mec:
Not sure how this got through, we'll look into it.

## Changes

## Checklist


mec:
We are linitng in the container which does not contain ALL the files, we selectively copy over the app files, so here, the Readme is not in the container and so skips linting in CI but not locally.

This is a pain for contatiners, but we can add the main files such as readme for now.

lozette (approved):

PR #1072, opened 2023-09-08
Add basic GIAS Establishment model

mec:
We want to model GIAS establishment so we can:

- have access to the unopen academy data
- allow searching
- improve performance
- break dependancies on the Academies API

This is a very simple start to that work that creates the table,
indexes, model and factory.

We don't validate any attribute other than `urn` because we do not own
the data (we'll be importing it later) so we need to remain flexible for
the moment and validate the data set first.

If a urn was missing, the record would be unusable for us, so we do
validate that attribute.

We try to balannce following the attribute names from the dataset, which
is the GIAS data from a signed in user with all fields selected, whilst
keeping thing approachable.

We store attriubute name and 'code' as these may be useful later, we
already model the 'GOR Region' codes on Project for example.

For the factory, we create a 'with_faith' trait as one of the main
indicators we use (showing diocese and marking some task as applicable).

Right now all of this is speculative, but takes us on a path we have
designed.

https://dfe-gov-uk.visualstudio.com/Academies-and-Free-Schools-SIP/_sprints/taskboard/Complete%20conversions%20transfers%20and%20changes/Academies-and-Free-Schools-SIP/Complete%20sprint%2031

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add deeds of termination for MFA task to Transfer projects
 - Added the 'Deed of termination for the church supplemental agreement' task to
   transfers.
+- The application can now store the un-published GIAS establishment records.
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


lozette (approved):
Approved. One tiny thing is that the commit mentinos `with_faith` as the factory trait, but it's actually `with_diocese`

mec:
> Approved. One tiny thing is that the commit mentions `with_faith` as the factory trait, but it's actually `with_diocese`

ah okay - thanks! I'll change it! 👍

PR #1076, opened 2023-09-11
Basic gias establishment importer

mec:
This is a super early version of the GIAS Establishment importer.

We need to get a feel for the actual data and there is a lot, ~50k rows, so the idea here is to allow developers to import the entire dataset and give it a prod to see what and how we want to move this on.

There is documentation about the process and background as part of this work which should explain things.

https://dfe-gov-uk.visualstudio.com/Academies-and-Free-Schools-SIP/_sprints/taskboard/Complete%20conversions%20transfers%20and%20changes/Academies-and-Free-Schools-SIP/Complete%20sprint%2031?workitem=138938


beeLorna (approved):
🚀 LGTM!

PR #1083, opened 2023-09-13
Update build workflow to use cache

mec:
We've proven the value of using the Github action cache to speed up
builds, here we add it to the deploy workflow.


DrizzlyOwl (approved):
Looks good to me. I'm assuming the cache-from bit will try to load as much as possible from the cache that was generated during the CI build?

mec:
> Looks good to me. I'm assuming the cache-from bit will try to load as much as possible from the cache that was generated during the CI build?

Yep, you can see the build step has dropped to ~23s.

PR #1092, opened 2023-09-18
Update GIAS Establishment import

mec:
We've been running the importer locally and thinking about how we could improve or change it.

We noticed that some of the 'codes' are padded with zeros and had to decided how to handle that - we've taken the approach of storing the values as strings and leave them alone -  this data is for reference only and we don't want to introduce more complexity than is necessary.

We also record the changes as we go and return a simple hash of the details of the import process.


beeLorna (approved):
👍

PR #1094, opened 2023-09-18
Update backing service dev scripts

mec:
The name of the two containers has changed, so these two scripts are not
looking for the right thing - this updates them.


lozette (approved):

beeLorna (approved):
👍

PR #1104, opened 2023-09-20
Basic event log model

mec:
We want offer users an 'audit log' of changes made in the application.

To faclilitate this we have the Event model. Events can be grouped and
assoicated to any ActiveRecord objects, a Project here.

We validate this all works by logging when a project is created, later we can
expose this to users in the UI.

This work lays a foundation for other logging to happen in a consistent way,
there is documentation in the PR.

This work does not include capturing further events or exposing the events.

We already know we want to log the events relating to importing GIAS data and
this also makes a clear path to that.

Interested to hear thoughts on naming here as it was tough to reach a decision -
we do have our old friend 'able' in 'eventable' - I've decided just to follow
convention here as it is at least clear what this might mean.

### DIFF
--- a/doc/event-logging.md
+++ b/doc/event-logging.md
@@ -1,0 +1,47 @@
+# Event logging
+
+Users and the DfE need a log of events that happen to the main entities in the
+service. To meet this need we have the `Event` model.
+
+Events offer a simple interface to create a user facing log of events in the
+application that can then be surfaced in the UI.
+
+An Event is linked to a user, who triggered the event and optionally another
+object that is associated to the event.
+
+Right now we consider all events that are not associated to another object
+'system' events.
+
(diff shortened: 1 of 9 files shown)


doc/event-logging.md
```diff
@@ -0,0 +1,47 @@
...
+To create a new system event:
+
+```ruby
+Event.log(grouping: :system, user: user, message: "A description of the event")
+```
+
+To create a new project event you also supply the project object as `with`:
+
+```ruby
+Event.log(grouping: :project, user: user, message: "A description of the project
```
lozette, line 29: There's no `with` in this example, and the `with: project` is mandatory

lozette (changes requested):

lozette (approved):

PR #1115, opened 2023-09-26
Update Container app module to v1.1.0

DrizzlyOwl:
## Changes

- New: App Insights Exceptions Log query alert rule so we can get notified when they happen
- Fix: CDN Origin hostname is now correctly referencing the FQDN of the Container App environment instead of the individual replica


olivia-work (approved):
LGTM

PR #1116, opened 2023-09-26
Refactor and fix Contacts

mec:
This work started out as 'I wonder why calling the `project` association on a `Contact` results in nil?...' to what you see here!

There is a fair bit so jump in:

The project association was caused by a naming collision between `Contact::Project` and `Project` - ideally we would rename the `Contact::Project` class - but the scope of felt a bit big (at the time!).

Whilst working on this we found a bug on the 'main contact' behaviour that meant any time the main contact for a category was updated, all of the other main contact states were dropped - this would have meant a sort of data loss if we deployed it.  We fixed that and simplified the approach, we don't think this is the final solution to this area, but it does let us deploy the feature. We've changed most of the language around this to 'primary contact for category' to avoid confusion with the 'main contact' for the project - as mentioned, this is be not means perfect -just a step forwards.

We also refactored how the director of child services is fetched, adding a service to do the fetching and so simplifying the way the view works - this fixes another issue where the director of child services is not showing up on transfer project and we no longer have to think about it.

I feel like:

- Contacts are pretty important and complex
- There is a lot more we can do if we have the capacity
- this is enough for now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   order of some of the information, remove some fields, add 50px spacing between
   each block.
 
+### Fixed
+
+- the director of child services is now shown on the external contacts view as
+  long as one is available, it is centrall managed and cannot be edited.
+
 ## [Release 39][release-39]
 
 ### Added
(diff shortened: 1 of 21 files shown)


lozette (approved):
Well done on this, I was baffled as to why the associations didn't just work by magic but now we know 🤦

PR #1118, opened 2023-09-27
Fix the GIAS import stats

mec:
We had these wrong, which got revealed when testing with chagned data
sets.

The import still works successfully, it is only the counts that were
worng.


lozette (approved):
Maths is hard

PR #1122, opened 2023-09-28
Import GIAS establishments data job

lozette:
## Changes

Add a job to run the GIAS Establishment Importer from a provided file. The job then sends an email to the requesting user with the output of the importer. 

The email has not been "contentified" - it just sends the raw output. We can iterate on this.

A requirement of this ticket was to run the job at 4am (or a provided time). However, we need to do this from the controller which calls the job. When we call the job, we should call it with (e.g.)

`Import::GiasEstablishmentImportJob.set(wait_until: <DateTime>).perform_later(file_path, user)`

Finally, when the job has been run & the email has been sent, the uploaded file is deleted.

## Checklist

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,0 +1,7 @@
+class ApplicationJob < ActiveJob::Base
+  # Automatically retry jobs that encountered a deadlock
+  # retry_on ActiveRecord::Deadlocked
+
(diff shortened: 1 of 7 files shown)


mec (approved):
🚢 this seems nice and straight forward - we'll get to prove that once we can get it running in prod.

PR #1125, opened 2023-09-29
Allow RCS and RDO users to add a new transfer project

lozette:
## Changes

We want to temporarily grant RCS users the ability to add new projects
(specifically Transfer projects, but all projects for convenience).

To do this, we have overridden the `add_new_project` column in the db with
a method. The db column is still set to `false` on creation of an RCS user,
but the overriding method returns true.

Once that is done, add the "Add new transfer" button for users who can `add_new_project`.

## Checklist


mec (approved):
🚢

PR #1131, opened 2023-10-04
Setup a maintenance page

beeLorna:
## Changes

This is a temporary static page that will be used when the service is being intentionally switched off to users for maintenance work. 
As the service won't actually be down, we are able to use HighVoltage for this. This reflects the other static pages we have within the service.

There is no navigation to this page from the service, but is accessed via `/maintenance` route.

Screenshot 2023-10-04 at 15 28 22

## Checklist


mec (approved):
Having the header nav isn't ideal - but I can see how much work it will be to drop it, so happy to go with this - clicking the links will take you back to the same page anyway 🚢

PR #1132, opened 2023-10-04
(Fix) Validate outgoing trust exists on Transfers

lozette:
When we put Transfer projects live, we were getting uncaught exceptions from a user trying to create a Transfer. After some investigation we discovered they were trying to add a UKPRN for a trust that doesn't exist, to the outgoing trust UKPRN field.

After some investigation we discovered that the outgoing trust UKPRN field was not being validated in the same way as the incoming trust. We were checking it was an integer, but not that the trust existed. There was a typo in the create transfer form spec which masked this error.

Add proper validations for both the incoming & outgoing trust UKPRNs, for both Conversions and Transfers.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased][unreleased]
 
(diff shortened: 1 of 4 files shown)


mec (approved):
🚢

PR #1135, opened 2023-10-05
Release 43

lozette:
Fixed

- Fix an error that meant outgoing trust UKPRNs were not being properly validated for Transfers. We were not checking the outgoing trust actually existed.

Added

- setup a maintenance page


mec (approved):
🚀

PR #1137, opened 2023-10-05
Import GIAS data upload page

lozette:
## Changes

Ticket [Task 140090](https://dfe-gov-uk.visualstudio.com/Academies-and-Free-Schools-SIP/_workitems/edit/140090): Upload GIAS establishment data via application

Add a very basic upload page to allow service support users to upload GIAS data. This needs content designing!

The page will show validation errors if the uploaded file is not CSV, or is empty, has the wrong headers, or is too large (> 150Mb)

On successful upload, the user will see a success message and the file will be scheduled for ingestion at 4am the next day. This time can be overridden via the `GIAS_IMPORT_TIME` environment variable.

If the ingestion fails later on, the user will receive an email with the errors; they will not be shown on this page.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,6 +127,7 @@ COPY config.ru ${APP_HOME}/config.ru
 COPY Rakefile ${APP_HOME}/Rakefile
 COPY script ${APP_HOME}/script
 COPY public ${APP_HOME}/public
+COPY storage ${APP_HOME}/storage
 COPY vendor ${APP_HOME}/vendor
 COPY bin ${APP_HOME}/bin
 COPY config ${APP_HOME}/config
--- a/app/mailers/gias_establishment_import_mailer.rb
+++ b/app/mailers/gias_establishment_import_mailer.rb
@@ -4,8 +4,18 @@ class GiasEstablishmentImportMailer < ApplicationMailer
       "316ef413-5e53-48e4-8a78-2aeaa9b98114",
       to: user.email,
       personalisation: {
-        result: result
+        result: format_result(result)
       }
     )
+  end
+
+  private def format_result(result)
+    <<~TEXT
+      Total CSV rows: #{result[:result][:total_csv_rows]}
+      New records: #{result[:result][:new_records]}
+      Changed records: #{result[:result][:changed_records]}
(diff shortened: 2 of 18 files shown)


Dockerfile
```diff
@@ -127,6 +127,7 @@ COPY config.ru ${APP_HOME}/config.ru
 COPY Rakefile ${APP_HOME}/Rakefile
 COPY script ${APP_HOME}/script
 COPY public ${APP_HOME}/public
+COPY storage ${APP_HOME}/storage
```
mec, line 130: 👌

app/views/shared/navigation/_service_support_primary_navigation.html.erb
```diff
@@ -14,6 +14,8 @@
 
           <%= render partial: "shared/navigation/primary_navigation_item", locals: {name: t("navigation.primary.service_support.users"), path: users_path, namespace: "/users"} %>
 
+          <%#= render partial: "shared/navigation/primary_navigation_item", locals: {name: t("navigation.primary.service_support.import.gias_establishments"), path: service_support_upload_gias_establishments_new_path, namespace: "/upload"} %>
```
mec, line 17: Shall we just delete this?

app/mailers/gias_establishment_import_mailer.rb
```diff
@@ -4,8 +4,19 @@ def import_notification(user, result)
...
       }
     )
   end
+
+  private def format_result(result)
+    <<~TEXT
+      Total CSV rows: #{result[:result][:total_csv_rows]}
+      New records: #{result[:result][:new_records]}
+      Changed records: #{result[:result][:changed_records]}
+      Changes: #{result[:result][:changes]}
```
mec, line 17: Changes has the potential to be a lot - I think we should avoid including it here for now, we need to decide what to do with all this info, if anything, but I am not sure it should go in an email.

mec (changes requested):
Couple of things then this is great! 👍

mec (approved):
🚢 Thanks for seeing this seemingly small but actually complex bit of work!

PR #1159, opened 2023-10-18
Remove any references to Sentry

lozette:
## Changes

We no longer use Sentry. We now use the `USER_ENV` env variable to determine which environment we're in, and we use Application Insights for error reporting.

## Checklist


mec (approved):
NIce, we should write an ADR to cover reporting in App Insights - I can do that! 👍

PR #1160, opened 2023-10-18
Deploy StatusCake SSL Checks for public endpoints

DrizzlyOwl:
**What is the change?**
Deploy 3 SSL Checks and a Contact Group on StatusCake

**Why do we need the change?**
Azure Monitor is not very reliable for TLS validity checks so we're using StatusCake instead.
We can deploy a SSL Certificate validity check that polls the dev, test, and prod health check endpoints once every 24H and notify us via Slack (integration)/Email at 14, 7 and 3 days before expiry

**What is the impact?**
None


DrizzlyOwl:
This feature is blocked due to Terraform state drift. Once v1.2.0 of the Terraform Container App Module ships, we can unblock this PR

DrizzlyOwl:
https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/1177

Stretch96 (approved):

PR #1161, opened 2023-10-18
Pin Pagy v5 & update govuk-components to v3.3

lozette:
## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -452,7 +452,7 @@ DEPENDENCIES
   erb_lint
   factory_bot_rails
   faraday
-  govuk-components (~> 3.1)
+  govuk-components (~> 3.3)
   govuk_design_system_formbuilder (~> 3.1)
   govuk_markdown
   high_voltage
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -154,10 +154,12 @@ GEM
     ffi (1.15.5)
     globalid (1.1.0)
       activesupport (>= 5.0)
-    govuk-components (3.3.0)
+    govuk-components (3.1.1)
```
mec, line 157: This is going backwards? Is that right?

mec (changes requested):
Is there a reason `govuk-components` goes back to 3.1.1 here?

Gemfile.lock
lozette: Ooh I didn't even notice! No that's not right

lozette: Oh no 
```
Bundler could not find compatible versions for gem "pagy":
  In Gemfile:
    pagy (~> 6.1)

    govuk-components (~> 3.3) was resolved to 3.3.0, which depends on
      pagy (~> 5.10.1)
```
Looks like we have to hold off updating pagy until govuk-components catches up?

mec: Aha! That is what I was worried about when pagy came in via Renovate!

Okay, shall pin Pagy to the compatible version, we can say why in the commit?

PR #1171, opened 2023-10-24
Extract and update Headteacher importer

mec:
When we ran the combined GIAS esatablishment and headteacher contacts import in the development environment, whilst the import suceeded, the data was all there as far as we could tell, the job seemed to die and the container restart at the end. 

We have also now moved our database to an application specific instance in Azure, which means we have a bigger window to run background jobs like these.

For these two reasons we decided to extract the headteacher import from the establishment import and run it later.

One could argue there are downsides to this approach (as ever!) but we think this is a good step in this context, we are  very much still in the testing phase for this data set coming into the application - once production is 'seeded' with the initial import, whch is large, the time taken and resources used will drop dramatically.

This work extracts the headteacher contact import and updates it to do more logging rather than adding errors to the result object that are then difficult for us to extract and use.

We take the same approach when updating the establishment import.

We recognise that there is duplication across the two csv import classes, but optimising that can wait whilst we test with the real data.

With the import service objects in place, we move on to add the background job and mailers to complete the feature.

One final step was to run the background job earlier, as mentioned we are in our own database instance now and so do not have to worry as much about the load on the instance.


lozette (approved):
Def more elegant than before 👍

PR #1172, opened 2023-10-25
Model Contact::Parliament

lozette:
## Changes

Model the Member of Parliament contact like the other contact types. However, this contact doesn't persist in our database - they are fetched from the Members API via the constituency name, and only modeled as a transient object.

If there are any errors from the Members API, raise them. We will have to ensure we handle these errors in the relevant controllers.

## Checklist

### DIFF
--- a/app/models/api/members_api/client.rb
+++ b/app/models/api/members_api/client.rb
@@ -44,7 +44,9 @@ class Api::MembersApi::Client
     constituency_data = constituency(search_term)
     raise constituency_data.error if constituency_data.error.present?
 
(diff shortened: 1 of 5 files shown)


mec (changes requested):
This is good, but I think it highlights that when a match cannot be found the client blows up?

Passing something like `constituency: "Mars West"` results in an unhandled error, when it should probably return some sort of 'No matching MP found for Mars West' error?

Could we expand this work to include that?

mec (approved):
Thank you! 🚢

PR #1178, opened 2023-10-27
Deploy File Shares and mount it to the Container App

DrizzlyOwl:
## Changes

- Deploys a Storage Account and a File Share and mounts it as a volume on the Container App.
- The volume mount path default is `/srv/app/storage`
- The volume is made available to the main container and the worker container so they can both read/write to it
- An IP address allow list can be specified to allow humans to access the file share on the Azure Portal


mec (approved):
Let's go 🚢

PR #1184, opened 2023-11-01
(Fix) Do not attempt to send "Assigned to" mail when assignee is nil

lozette:
Sentry error: https://sdd-n7.sentry.io/issues/4589134654/events/77ffc5f7670946b48667dfe044c8ebbd/

We were seeing an error on production where someone seemed to be trying to send an "assigned to" mail on a project with no assignee. The project's `assigned_to` value was nil, and no-one was being assigned to the project, and as a result the mailer was trying to sent an email to a nil person.


beeLorna (approved):
👍

PR #1193, opened 2023-11-03
Extend search service to search by establishment name and GIAS number

lozette:
## Changes

Note that the search service is still not available to users.

Extend the existing search service to search by words (establishment name) and GIAS number (4-digit string). 

The search by words allows partial matches and is case-insensitive.

The search by GIAS number expects an exact 4-digit number. 6 digit numbers are assumed to be URNs and 8-digit numbers are assumed to be UKPRNs. We cannot do partial number matches at the moment.

## Checklist

### DIFF
--- a/app/services/project_search_service.rb
+++ b/app/services/project_search_service.rb
@@ -8,6 +8,10 @@ class ProjectSearchService
       return search_by_ukprn(query)
     end
 
+    if word_pattern(query)
+      return search_by_words(query)
+    end
+
     []
   end
 
@@ -20,11 +24,27 @@ class ProjectSearchService
       .or(Project.where("outgoing_trust_ukprn = ?", ukprn))
   end
 
+  def search_by_words(query)
+    establishment_results = search_establishments_by_name(query)
+    return [] if establishment_results.empty?
+
+    urns = establishment_results.pluck(:urn)
+    search_by_urns(urns)
+  end
+
+  private def search_establishments_by_name(query)
+    Gias::Establishment.where("LOWER(name) LIKE ?", "%#{query.downcase}%")
+  end
+
   private def urn_pattern(query)
     query.match?(/^\d{6}$/)
   end
 
   private def ukprn_pattern(query)
     query.match?(/^\d{8}$/)
+  end
+
+  private def word_pattern(query)
+    query.match?(/\D/)
   end
 end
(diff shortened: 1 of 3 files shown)


app/services/project_search_service.rb
```diff
@@ -25,6 +25,6 @@ def search_by_ukprn(ukprn)
   end
 
   private def ukprn_pattern(query)
-    query.match?(/\d{8}$/)
+    query.match?(/^\d{8}$/)
```
mec, line 60: I wonder if we should put in more coverage around incorrect query strings like this - clearly originated with me, but might be worthwhile?

app/services/project_search_service.rb
```diff
@@ -11,8 +11,8 @@ def search(query)
     []
   end
 
-  def search_by_urn(urn)
-    Project.where("urn = ?", urn)
+  def search_by_urns(urns)
+    Project.where(urn: urns)
```
mec, line 23: So this has changed: are we confident we are doing enough to prevent injection here? When this is called from line 4 it is user input?

app/services/project_search_service.rb
```diff
@@ -8,6 +8,10 @@ def search(query)
       return search_by_ukprn(query)
     end
 
+    if word_pattern(query)
```
mec, line 11: I wonder if this just becomes a case with a default? Does it look like a urn, do this, Does it look like a UKPRN, do this, otherwise assume it is a name?

app/services/project_search_service.rb
```diff
@@ -12,6 +12,10 @@ def search(query)
       return search_by_words(query)
     end
 
+    if establishment_number_pattern(query)
```
mec, line 15: Now it really feels like a case statement! Perhaps the default is to return no results because we couldn't infer what the users query is?

mec (changes requested):
I love seeing this starting to work! I think we can tighten this up a little in places - let me know what you think?

app/services/project_search_service.rb
lozette: Yes I'm confident this is OK! https://www.fastruby.io/blog/rails/security/dangerous-query-method-deprecation.html

app/services/project_search_service.rb
lozette: Done

app/services/project_search_service.rb
mec: Nope, sorry, must have missed them!

mec (approved):
🚢

PR #1194, opened 2023-11-03
Update TF Container App module to 1.3.1

DrizzlyOwl:
## Changes

- Fixes Redis networking so that connections do not rely on the public internet and instead go via private endpoints.
- Adds the ability to override the default redis configuration with the `redis_config` variable
- Adds the ability to change the SKU of the redis instance with the `redis_sku` variable
  - This will change from the default "Basic" to "Standard" in Production.The specs remain the same, but an SLA is included with Standard so it is more suitable for production workloads.


mec (approved):
Redis SKU!!? 😄 🚢

PR #1201, opened 2023-11-06
Search projects

lozette:
## Changes

Add UI search to the application (!)

Add a search box to the DfE header as per the [standard pattern](https://design.education.gov.uk/design-system/components/header#example3).

Add search results views which differ depending on the number of results.

No search term entered:

Screenshot 2023-11-07 at 11 46 45

No results found: 

Screenshot 2023-11-07 at 11 47 24

Results found: 

Screenshot 2023-11-07 at 11 48 01

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased][unreleased]
 
+### Added
+
+- Enable Search in the UI, add a search box in the header and search results
+  view. We can search by:
+  - school/academy name
+  - school/academy URN
+  - incoming trust by UKPRN
+  - outgoing trust by UKPRN
+
 ## [Release-45][release-45]
 
 ### Added
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,0 +1,15 @@
+class SearchController < ApplicationController
+  def results
+    return unless query.present?
+
+    search_service = ProjectSearchService.new
+    @results = search_service.search(query)
+    @query = query
+    @count = @results.count
+    @pager, @results = pagy_array(@results)
+  end
+
+  private def query
+    params[:query]&.strip
+  end
+end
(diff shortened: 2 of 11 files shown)


app/controllers/search_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    search_service = ProjectSearchService.new
+    @results = search_service.search(query)
+    @query = query
+    @count = @results.count
+    @pager, @results = pagy_array(@results)
+  end
+
+  private def query
+    params[:query]
```
mec, line 13: I don't know if it should be here, but we should `strip` the query string, trim leading and trailing white space.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased][unreleased]
 
+### Added
+
+- Enable Search in the UI, add a search box in the header and search results
+  view
```
mec, line 12: Can we just outline what we can serach for so it is recorded? I think it is:

- school/academy name
- school/academy URN
- incoming trust by UKPRN
- outgoing trust by UKPRN

app/views/shared/navigation/_dfe_header_search.html.erb
```diff
@@ -0,0 +1,26 @@
+<div class="dfe-header__search">
+  <button class="dfe-header__search-toggle" id="toggle-search" aria-controls="search" aria-label="Open search">
+    <svg class="dfe-icon dfe-icon__search" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" focusable="false" width="27" height="27">
+      <path d="M19.71 18.29l-4.11-4.1a7 7 0 1 0-1.41 1.41l4.1 4.11a1 1 0 0 0 1.42 0 1 1 0 0 0 0-1.42zM5 10a5 5 0 1 1 5 5 5 5 0 0 1-5-5z"></path>
+    </svg>
+    <span class="govuk-visually-hidden"><%= t("search.search") %></span>
+  </button>
+  <div class="dfe-header__search-wrap" id="wrap-search">
+    <form class="dfe-header__search-form" id="search" action="/search" method="get" role="search">
```
mec, line 9: Do you think we should use a form builder here so that we get the CSRF protection and stuff like that?

mec (changes requested):
Super nice and clean work! Couple of things I would like to see, let's go!

app/views/shared/navigation/_dfe_header_search.html.erb
lozette: Added - although I had to leave the SVG button stuff as-is because I have no idea how to do those with form helpers 😆

mec: That IS how you do those! The form doesn't add what I thought - but at least we are using it! 👍

mec (approved):
Brilliant! Let's get this on dev and get people to go look!

PR #1202, opened 2023-11-07
Add New project table view for your team projects

beeLorna:
## Changes

Our views of projects are usually sorted by the significant date, which is the most valuable, but there are some instances where it is valuable to know what projects have been added and to view them in order of when they were first created.

Screenshot 2023-11-08 at 12 15 11

## Checklist

### DIFF
--- a/app/controllers/team/projects_controller.rb
+++ b/app/controllers/team/projects_controller.rb
@@ -2,6 +2,11 @@ class Team::ProjectsController < ApplicationController
   after_action :verify_authorized
   rescue_from ActiveRecord::RecordNotFound, with: :not_found_error
 
+  def new
+    authorize Project, :index?
+    @pager, @projects = pagy_array(ByTeamProjectFetcherService.new(current_user.team).new)
+  end
+
   def in_progress
     authorize Project, :index?
     @pager, @projects = pagy_array(ByTeamProjectFetcherService.new(current_user.team).in_progress)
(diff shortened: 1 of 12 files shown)


spec/services/by_team_project_fetcher_service_spec.rb
```diff
@@ -49,6 +49,62 @@
     end
   end
 
+  describe "#new" do
+    before do
+      Timecop.freeze(Time.local(2023, 1, 1, 0, 0))
```
lozette, line 54: We don't need to use Timecop as Rspec has built-in time helpers already. Have a look at other tests which use `travel_to`, e.g. /spec/requests/all/by_month/projects_controller_spec.rb#L20

app/controllers/team/projects_controller.rb
```diff
@@ -2,6 +2,11 @@ class Team::ProjectsController < ApplicationController
   after_action :verify_authorized
   rescue_from ActiveRecord::RecordNotFound, with: :not_found_error
 
+  def new
+    authorize Project, :index?
+    @pager, @projects = pagy_array(ByTeamProjectFetcherService.new(current_user.team).in_progress)
```
lozette, line 7: Should this be calling the `new` method on the Service, not `in_progress`?

beeLorna: yes, this should be `new` - thanks!

lozette (approved):

PR #1204, opened 2023-11-08
Add Gias::Group model

lozette:
A Gias Group is what we currently think of as a "trust", but the data from GIAS contains more than just trusts, so for safety's sake we are calling them Groups.

Note that the Companies House number is being stored as a string, because it can contain leading zeros which are significant. If we stored it as an integer, Rails would "helpfully" strip off leading zeroes.

### DIFF
--- a/db/migrate/20231108110923_add_gias_group.rb
+++ b/db/migrate/20231108110923_add_gias_group.rb
@@ -1,0 +1,18 @@
+class AddGiasGroup < ActiveRecord::Migration[7.0]
+  def change
+    create_table :gias_groups, id: :uuid do |t|
+      t.integer :ukprn, index: true
+      t.integer :unique_group_identifier, index: {unique: true}
+      t.string :group_identifier, index: true
+      t.string :original_name
+      t.string :companies_house_number
+      t.string :address_street
+      t.string :address_locality
+      t.string :address_additional
+      t.string :address_town
+      t.string :address_county
+      t.string :address_postcode
+      t.timestamps
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


db/migrate/20231108110923_add_gias_group.rb
```diff
@@ -0,0 +1,17 @@
+class AddGiasGroup < ActiveRecord::Migration[7.0]
+  def change
+    create_table :gias_groups, id: :uuid do |t|
```
mec, line 3: It's possibly not on the card, but can we add the `Unique group identifier (UID)` as I think this the GIAS ID for the record whereas `group_identifier` is from KIM and `UKPRN` is from an external body so possibly not as robust.

I would possibly label the column `urn` as that makes it equivilent to URN on Establishment - not sure why they chose a different name for the value, but it looks like the value is used in GIAS:

https://get-information-schools.service.gov.uk/Groups/Group/Details/2073

mec (changes requested):
Wasn't expecting to see this so soon, but here it is! 😄

db/migrate/20231108110923_add_gias_group.rb
lozette: I can certainly do that - are we sure about calling it URN? Only I'm thinking of the `urn_validator` which expects 6 digits, while these are 4 digits. I know we don't have to use the validator, but it would be confusing for URNs to be 6 digits (and enforced as 6) in one place but 4 here.

mec: Fair point - can we still include it, I think it is important `unique_group_identifier` is fine! 👍

mec (approved):
🚢

PR #1213, opened 2023-11-16
Users are not directed to update KIM in 'Receive grant payment certificate' task

lozette:
## Changes

We are gradually removing any need for or mention of KIM in our applications, so remove this action in the Receive Grant Payment Certificate task as we do not want to encourage people to still use KIM.

## Checklist


beeLorna (approved):
👍

PR #1283, opened 2024-01-09
(151107) Add date field to Conversions Receive grant payment certificate task, and split "Check and save" action

lozette:
1. We need to be able to enter the date the Grant payment certificate was received to the Conversion task list. This was not a simple piece of work as we do not yet have any tasks that collect dates.

Refactor the Grant payment certificate task form to collect a date, or show an existing date if one was previously entered. We have also modified the `Project#grant_payment_certificate_received?` method to take the date received into account.

2. Then, split the "Check and save" action on this task into "Check" and "Save". This has required a database data migration for existing projects. 

Without an entered date

Screenshot 2024-01-09 at 16 07 08

With an entered date

Screenshot 2024-01-09 at 16 07 36

## Checklist

### DIFF
--- a/app/forms/conversion/task/receive_grant_payment_certificate_task_form.rb
+++ b/app/forms/conversion/task/receive_grant_payment_certificate_task_form.rb
@@ -1,3 +1,77 @@
 class Conversion::Task::ReceiveGrantPaymentCertificateTaskForm < BaseTaskForm
   attribute :check_and_save, :boolean
+  attribute "date_received(1i)"
+  attribute "date_received(2i)"
+  attribute "date_received(3i)"
+  attribute :date_received, :date
+
+  validate :date_format
+
+  def initialize(tasks_data, user)
+    @tasks_data = tasks_data
+    @user = user
+    @project = tasks_data.project
+    super(@tasks_data, @user)
+
+    self.date_received = @project.tasks_data.receive_grant_payment_certificate_date_received
+  end
+
+  def save
+    @tasks_data.assign_attributes(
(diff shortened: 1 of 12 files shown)


app/forms/conversion/task/receive_grant_payment_certificate_task_form.rb
```diff
@@ -1,3 +1,79 @@
...
+    Date.new(year, month, day)
+  rescue TypeError
+    errors.add(:date_received, I18n.t("conversion.task.receive_grant_payment_certificate.date_received.errors.format"))
+  end
+
+  private def day
+    day = attributes["date_received(3i)"]
+    return if day.blank?
+
+    month.to_i
```
Floppy, line 47: This should be `day.to_i`, I think! Might want to change a test to not use Jan 1, to pick up the bug!

spec/forms/conversion/tasks/receive_grant_payment_certificate_task_form_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  describe "#save" do
+    before { mock_successful_api_response_to_create_any_project }
+
+    let(:project) { create(:conversion_project) }
+
+    context "when the date is a valid date" do
+      it "sets date_received to the supplied date" do
+        form = described_class.new(project.tasks_data, user)
+        form.assign_attributes("date_received(1i)": 2024, "date_received(2i)": 1, "date_received(3i)": 1)
```
Floppy, line 14: Change to not Jan 1 to pick up the bug in day parsing.

Floppy (changes requested):

PR #1312, opened 2024-01-24
Enable service support users to be able to update all projects

beeLorna:
## Changes

The service support user does a lot of data quality work in KIM and updates projects within this service on behalf of regional delivery officers when there is an error. This change will allow this process to happen without needing the assigned user to be present and make the process a lot simplier.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Service support users can navigate to the GIAS data upload page
 - Add additional columns to the ESFA report
+- Service support users can now update projects which are assigned to other
+  users
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


lozette (changes requested):
One spec to add.

Also I think we should change /app/helpers/project_helper.rb#L61 which shows the "Not assigned to project" banner. I think we should remove the banner if a service_support user is viewing a project, so they know they can edit a project.

Screenshot 2024-01-24 at 14 02 26

If you change line 61 on that file to `elsif !policy(project).update? || project.assigned_to.nil?` it should do it (i.e. `if the policy does not permit this user to update the project, or the project is assigned to no one`)

Edit: Nope, my mistake, ` !policy(project).update?` does not work here. Lorna devising a different method to get the same result!

lozette (approved):

PR #1347, opened 2024-02-13
(156240) Enable business support users to view exports

lozette:
We added the business support team a while ago, but never actually granted any rights to users in that team. Add them to the "can export" team group via the ExportPolicy. Check they can see the export pages.

## Checklist


mec (approved):
:ship:

beeLorna (approved):

PR #1381, opened 2024-02-28
Feature/157878 show form a mat label in details page

lozette:
## Changes

Screenshot 2024-02-28 at 16 28 02

## Checklist


app/models/project.rb
```diff
@@ -84,9 +84,14 @@ def establishment
   end
 
   def incoming_trust
+    return new_trust_object if form_a_mat?
     @incoming_trust ||= fetch_trust(incoming_trust_ukprn)
   end
 
+  def new_trust_object
```
mec, line 91: Sneaky! I think I like this!

mec (approved):
Yeah, this feels temporary with the non-existent Trusts in a spreadsheet - which is my understanding of the whole approach, so let's go! 🚢

mec:
Oh and the 'You can find the TRN in the [new trusts spreadsheet (opens in new tab)' link on the new MAT from doesn't open in a new tab - no related, but noticed during this review!

app/models/project.rb
lozette: Haha seemed like the easiest thing to do to ensure it all works as expected!

PR #1385, opened 2024-02-29
Fix SonarCloud Issues reported on Dockerfile

DrizzlyOwl:
## Changes

- Variables are properly escaped
- No vendor recommended packages are installed via `apt-get`
- References to version pins are made consistent
- Create a non-root user and assign it permission to manage the app home dir
- Run the Docker container as the non-root user
- Ensure references to `curl` are enforcing TLS connections and HTTPS protocol and limiting redirects where possible to prevent domain highjacking
- Clean up apt cache after installing packages

## Checklist


mec (commented):
Thanks for all the effort on this! 👍 I am happy to go, but I think @lozette should have site of these change as well and then approve.

Dockerfile
```diff
@@ -107,6 +117,7 @@ RUN \
   else \
     yarn install --frozen-lockfile; \
   fi
+USER root
```
lozette, line 121: Question bc I don't understand it: why do we switch back to root here?

mec: I had assumed that eveything after has to run as root, up until the next `USER` - but @DrizzlyOwl should advise?

DrizzlyOwl: We might be able to remove this line actually

lozette (approved):

PR #1392, opened 2024-03-04
(158463) Fall back to next trust contact if a project doesn't have a main trust contact

lozette:
While a user can select a main outgoing/incoming trust contact ID, not all projects have one selected. If this is the case, we can try to fall back onto the next trust contact, instead of presenting an empty column.

## Checklist


mec (approved):
🚢

PR #1400, opened 2024-03-05
(158464) Add School, Incoming trust and "Other" contacts to Funding agreement letters export

lozette:
## Changes

Add more columns to the Funding agreement letters export: School contact details, Incoming trust contact details and "Other" contact details.

## Checklist

### DIFF
--- a/app/presenters/export/csv/project_presenter.rb
+++ b/app/presenters/export/csv/project_presenter.rb
@@ -266,26 +266,18 @@ class Export::Csv::ProjectPresenter
   end
 
   def other_contact_name
-    contacts = ContactsFetcherService.new.all_project_contacts(@project)
-    return if contacts["other"].blank?
-
-    contact = contacts["other"].first
-    contact&.name
+    other_contact&.name
   end
 
   def other_contact_email
-    contacts = ContactsFetcherService.new.all_project_contacts(@project)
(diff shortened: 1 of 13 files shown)


app/presenters/export/csv/school_presenter_module.rb
```diff
@@ -94,4 +94,40 @@ def school_sharepoint_folder
   end
 
   alias_method :school_sharepoint_link_with_academy_label, :school_sharepoint_folder
+
+  def school_main_contact_name
```
mec, line 98: I think these would be way more efficient if we do all the work to get the contact once and the present the name, email and title from that.

app/presenters/export/csv/project_presenter.rb
```diff
@@ -264,4 +264,28 @@ def esfa_notes
 
     notes.map(&:body).join("\n\n")
   end
+
+  def other_contact_name
```
mec, line 268: Same here, makes me feel like perhaps the ContactsFetcherService should be responsible for this effort?

mec (changes requested):
I think it is worth moving the responsibility for getting the contact or next one out of the presenter, at the very least, getting the contact once and then presenting the resulting attribute.

mec (approved):

PR #1421, opened 2024-03-13
Add static favicon to root of public dir

DrizzlyOwl:
## Changes

* This is not dynamically updated from the gov-uk frontend so will need to be manually updated if this ever changes

* Enables older browsers to safely load a fallback/default favicon from the root of the web app

## Checklist


mec (approved):
Meh! If we must, now we have an asset outside of the asset pipeline. I think there is a case to be made that we should not use the crown anyway, as we are not part of GOV.UK. But I can see how this will silence the logging. 🚢

PR #1422, opened 2024-03-13
Refactor ContactsFetcherService for performance

mec:
We do a lot of fetching contacts and this service is pretty happy to hit
the database a lot. In production with ~120 projects, exports are taking
a full minute to run and so the app is timing out.

We want to start by optimising our queries and this is the first
candidate (with more to follow).

The principles here are:

- there can generally be one ContactsFetcherService for one Project as
  it references the project internally
- our exports presenters are actually modules that all get mixed in to
  `project_presenter` so we make one instance of ContactsFetcherService
  available to all rather than instantiating (and so new db calls) new
  ones

This change has a big impact on the presenter specs as we have to stub
extra methods or (as we chose) switch to factories.

The result of this is, for a single conversion:

- 39 database queries per project/row in export before change
- 23 database queries per project/row in export after change

We'll have to see if this makes enough of a difference and we have
identified more optimisations we can make.

### DIFF
--- a/app/controllers/external_contacts_controller.rb
+++ b/app/controllers/external_contacts_controller.rb
@@ -2,7 +2,7 @@ class ExternalContactsController < ApplicationController
   include Projectable
 
   def index
-    @grouped_contacts = ContactsFetcherService.new.all_project_contacts(@project)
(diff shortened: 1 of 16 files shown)


lozette (approved):
Nice 😎

PR #1423, opened 2024-03-13
(159544) (Fix) Exclude Form a MAT projects from By Trust page

lozette:
## Changes

We received a bug report that the By Trusts page was not working after we
released the Form a MAT conversion/transfer functionality. This was an
oversight by us, as the By Trusts page relies on grouping projects by their
incoming trust UKPRN - and Form a MAT projects do not have an incoming trust
UKPRN.

As a quick fix, we have excluded Form a MAT projects from this page.

Going forward, we will have to implement a better fix (may need design work) to
include Form a MAT projects on this page under their "trust to be formed"
trusts (which do not yet exist).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Show the date a conversion happened in the ESFA export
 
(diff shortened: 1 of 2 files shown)


mec (approved):
🚢

PR #1428, opened 2024-03-14
(159128) Create a Schools due to convert export

lozette:
## Changes

We are going to merge the ESFA & Grant management and Finance exports into one
bigger export.

The name of this export is not "schools due to convert" in the UI, but because
we already have an "academies due to transfer" export, the naming of this export
as "schools due to convert" makes sense.

This export is more or less every column of data we collect for conversion
projects.

We haven't had to add any more presenters, it reuses all existing presenters.

## Checklist


mec (approved):
🚢 it's so good how straightforward we made this - will be interesting to see how slow this one is!

PR #1429, opened 2024-03-14
(159129) Add UI for "all schools converting" mega-export

lozette:
## Changes

Build out the UI for the new "mega" Conversions export. This is at `/projects/all/export/by-month/conversions` but not currently linked from the Exports page.

Screenshot 2024-03-15 at 10 59 50

Screenshot 2024-03-15 at 11 00 37

## Checklist

### DIFF
--- a/app/services/projects_for_export_service.rb
+++ b/app/services/projects_for_export_service.rb
@@ -24,6 +24,11 @@ class ProjectsForExportService
     AcademiesApiPreFetcherService.new.call!(projects)
   end
 
+  def conversion_by_month_projects(month:, year:)
+    projects = conversion_projects_by_month_and_year(month, year)
+    AcademiesApiPreFetcherService.new.call!(projects)
+  end
+
   private def transfer_projects_by_month_and_year(month, year)
     Transfer::Project.filtered_by_significant_date(month, year)
   end
--- a/app/services/projects_for_export_service.rb
+++ b/app/services/projects_for_export_service.rb
@@ -1,14 +1,4 @@
 class ProjectsForExportService
-  def risk_protection_arrangement_projects(month:, year:)
-    projects = conversion_projects_by_month_and_year(month, year)
-    AcademiesApiPreFetcherService.new.call!(projects)
-  end
-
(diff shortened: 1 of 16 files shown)


app/views/all/export/by_month/conversions/projects/index.html.erb
```diff
@@ -0,0 +1,37 @@
...
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h3 class="govuk-heading-l"><%= t("export.by_month.conversions.index.title") %></h3>
+    <%= t("export.by_month.conversions.index.body_html") %>
+  </div>
+</div>
+
+<table class="govuk-table" name="projects_table" aria-label="Transfers by month exports">
```
mec, line 16: `aria-label` should be 'Conversions by month exports'?

app/services/projects_for_export_service.rb
```diff
@@ -1,14 +1,4 @@
 class ProjectsForExportService
-  def risk_protection_arrangement_projects(month:, year:)
```
mec, line 2: This came up when I was triaging - I decided I like the verbosity and link to what each export is called by users and told Sonarcloud to ignore this one, for me I'll take the duplication for the more 'obvioius' code - plus machines are silly - happy to keep this, or drop it and tell SC to ignore it as a personal preference

mec (approved):
I prefer the original duplicated service code and telling SonarCloud to ignore it, but it's not a blocker.

There is a tiny copy and paste aria label error to fix and then go 🚢

app/services/projects_for_export_service.rb
lozette: I'm actually going to be removing those methods in a later PR (as we're getting rid of the RPA and funding agreement letters exports) so I'll leave these for now. But thanks for the heads up 👍

app/views/all/export/by_month/conversions/projects/index.html.erb
lozette: Cut & paste 😬

PR #1431, opened 2024-03-14
(159810) Temporarily remove future date validation for confirmed project dates

lozette:
We need to temporarily remove the future date validation for confirmed significant dates. People are backfilling the application with projects already completed, and the confirmed dates will be in the past for some of these projects.

## Checklist


mec (approved):
🚢

PR #1454, opened 2024-03-22
(Chore) Refactors and redundant code removal

lozette:
## Changes

Some refactoring and removal of unneeded code, plus a small bugfix.

- Replace uses of FactoryBot `create` with `build` where possible to speed things up
- Replace multiple uses of mocks where only one is needed
- Remove a redundant migration that was used once only
- Remove an unused task and its skipped tests
- Fix a small bug in the Project presenter, the true value of a Transfer task was being obscured by some inappropriate shared code


mec (approved):
👩‍🍳 👌 🚢

PR #1460, opened 2024-03-26
(Fix) Clear the value of sponsored_support_grant_type if "Not applicable" ticked

lozette:
We had a bug (more a UI anomaly) on the Sponsored Support Grant task: a user had previously selected a support grant type but now wanted the task to be "Not applicable". However because the support grant types are in radio buttons, they were unable to clear the value.

This confused the user, as although the task was now correctly logged as "not applicable" the UI showed the type still selected.

Override the `save` method on these two tasks, so that if the Not applicable box is checked, any value in sponsored_support_grant_type is removed.

image (8)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased][unreleased]
 
+### Fixed
+
+- Clear any radio button values in the Sponsored support grant type task if the
+  task is later saved as "Not applicable"
+
 ## [Release-61][release-61]
 
 ### Added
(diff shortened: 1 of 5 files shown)


mec (approved):
This solution is good, but it makes me wonder if we need to think about a wider issue here? Should this be the same for all 'optional' tasks? Should 'not applicable' be one of the radio options for this task? Should we add cards to the backlog to go and update other tasks?

lozette:
@mec I was pondering that too. Maybe all optional tasks with radio buttons need this. Might be worth discussing with @IainMacDonaldUX

lozette:
Going to merge this for now and add a tech debt item to address this

PR #1463, opened 2024-03-27
Add an Information banner

mec:
We've used this banner before for scheduled maintenance so it makes
sense to make it even more generic and quicker to use.

Here we add it to every page when the appropriate environment variable
is set to true and add the content to the locales to make identifying
and updating quick and painless.

image (1)


README.md
```diff
@@ -40,6 +40,7 @@ code straight forward for those that would like to use them.
 - [Infrastructure and Terraform](/doc/infrastructure-and-terraform.md)
 - [Developer scripts](/doc/developer-scripts.md)
 - [GOV.UK Notify](/doc/govuk-notify.md)
+- [Information banner](/dov/information-banner.md)
```
lozette, line 43: Typo - Should be `doc` not `dov`

lozette (changes requested):

lozette (approved):

PR #1473, opened 2024-04-04
Introduce a concept of "state" to projects

lozette:
We want to be able to mark projects as "deleted", but not actually delete the record from the database. We intend to add a deleted state to projects .

Ahead of this, we have decided to add a `state` to projects, and enumerate it for the various different states a project can be in. For now, this is "active" and "completed"

Add an enumerator to projects - 0 for active (default) and 1 for completed.

As the state is an enumerator, we get the scope Project.completed for free now. Due to Rails magic, we also get Project.not_completed - anything not in the completed state. In the first commit, we keep Project.not_completed as is; in the second, we replace Project.not_completed with the more semantic Project.active.

As the original Project.completed scope had an order (ordering by completed_at date) we have had to split this ordering out and add it as a new scope.

We have added the new state to the project factories, and updated the relevant tests.

Unfortunately this was the smallest single commit we could make for this change.

(No user-facing changes in this commit)

### DIFF
--- a/app/controllers/all/completed/projects_controller.rb
+++ b/app/controllers/all/completed/projects_controller.rb
@@ -4,7 +4,7 @@ class All::Completed::ProjectsController < ApplicationController
 
   def index
     authorize Project, :index?
-    @pager, @projects = pagy(Project.completed)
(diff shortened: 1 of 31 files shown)


mec (approved):
🚢

PR #1474, opened 2024-04-04
(160908) Users can change the incoming trust UKPRN on conversion projects

mec:
This is the first in a series of PRs that we allow certain project attributes to be changed by users.

Here the incoming trust UKPRN can be chagned by going to 'About this project' and using the appropriate 'change' link.

We've decided to keep all the changable attributes on one edit form, rather than splitting them out, we've added anchors so each change link jumps to the relevent section of the edit form.

We carry out the same set of validation for the value as the create form, meaning the value must:

- be present
- be a valid UKPRN
- be for a trust that can be located on the API

This means users cannot use the from to change a conversion from joining a multi-academy trust to forming one, this is by design.

### Screenshot of the edit form
image


lozette (approved):
👍

PR #1478, opened 2024-04-05
(161623) Do not show "deleted" projects in application

lozette:
## Changes

Now that we have the concept of "soft deleted" projects (projects which exist in the database but are in a deleted state) we need to ensure they never appear to users.

To ensure this, we have decided to use a `default_scope` for projects. Default scopes can be problematic as they can conceal issues and cause confusion - but the alternative is to add `active` and `completed` scopes to ALL calls to Project in the application. 

This does mean that we lose access to the `Project.deleted` scope, but if we need to call on deleted projects specifically we can unscope the default scope and call `Project.unscoped.where(state: :deleted)`

We then go on to add some Policy changes and extra specs, to make sure deleted projects do not appear in listing pages, and cannot be shown/edited.

## Checklist

### DIFF
--- a/app/controllers/all/export/grant_management_and_finance_unit/conversions/projects_controller.rb
+++ b/app/controllers/all/export/grant_management_and_finance_unit/conversions/projects_controller.rb
@@ -12,7 +12,7 @@ class All::Export::GrantManagementAndFinanceUnit::Conversions::ProjectsControlle
   end
 
   def show
-    authorize Project, :show?
+    authorize Project, :index?
     @month = Date.new(year, month)
   end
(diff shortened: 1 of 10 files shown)


mec (approved):
Ugh! `default_scope` again and me reviewing it again (we have a default scope on Note)! 😱 

I suspect that, like you, I cannot come up with a good reason not to use it here, 'deleted' projects are a pretty odd concept, it's hard to understand what they are actually for.

For that reason, I am going to go against everything I have read and been told by better developers than me about never using `default_scope` and approve this!

PR #1480, opened 2024-04-08
(160908) Users can change directive academy order and two requires improvement responses

mec:
Based on #1481 

Following the existing pattern for allowing users to edit conversion project attributes, here we allow:

-  academy order type
- result of two requires improvement

to be changed.

### DIFF
--- a/app/helpers/project_helper.rb
+++ b/app/helpers/project_helper.rb
@@ -84,4 +84,11 @@ module ProjectHelper
       OpenStruct.new(id: false, name: I18n.t("helpers.responses.conversion_project.directive_academy_order.no"))
     ]
   end
+
+  def yes_no_responses
+    @yes_no_responses ||= [
+      OpenStruct.new(id: true, name: I18n.t("yes")),
+      OpenStruct.new(id: false, name: I18n.t("no"))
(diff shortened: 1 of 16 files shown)


app/helpers/project_helper.rb
```diff
@@ -77,4 +77,18 @@ def confirmed_date_original_date(project)
     original_date = project.provisional_date
     "#{confirmed_date.to_fs(:govuk_short_month)} (#{original_date.to_fs(:govuk_short_month)})"
   end
+
+  def directive_academy_order_responses
```
lozette, line 81: As these are just yes & no I'm wondering if it'd be easier/less code to drop them and use `yes_no_responses` again?

lozette (approved):
Approved with one comment

app/helpers/project_helper.rb
mec: It's confusing for sure! But the actual responses are 'directive academy order' and 'academy order', we have to keep mentally mapping between the two because the attribute is `directive_academy_order` and boolean, but we must have decided this other way is better and not felt it worth changing the database to something like an enum.

PR #1484, opened 2024-04-09
(161334) Users can change transfer project details

mec:
This works follow the path already laid by conversions, allowing specific attributes on transfer projects to be changed - it's pretty much rinse and repeat.

The only thing to watch out for is that where we update a `project` and `tasks_data` we chose not to use a transaction as it feels better to know something has gone wrong here, for this reason, we just call `save!` which will raise an error which the team will see, this seems better than rolling back and telling the user nothing.


spec/validators/outgoing_incoming_trusts_ukprn_validator_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.describe OutgoingIncomingTrustsUkprnValidator do
+  subject { OutgoingIncomingTrustsUkprnValidatorTest.new }
+
+  context "when the outgoing and incoming trust UKPRNs are the same" do
```
lozette, line 6: The contexts on here are the wrong way round - this one should be "when they are different"

spec/validators/outgoing_incoming_trusts_ukprn_validator_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    it "is valid" do
+      subject.outgoing_trust_ukprn = 12345678
+      subject.incoming_trust_ukprn = 87654321
+
+      expect(subject).to be_valid
+    end
+  end
+
+  context "when the outgoing and incoming trust UKPRNs are different" do
```
lozette, line 15: This context should be "when they are the same"

lozette (changes requested):

lozette (approved):

PR #1492, opened 2024-04-11
Disable the Add notes for ESFA task

mec:
User research has shown that data consumers do not use this information
so we want to disable the task for now but not delete it.

This is achieved by removing the task from the task list definition.

We've also skipped the spec.

This work called for the note to be removed from the exports, but I cannot find it in any exports? It is in the presenter so may have been missed - so that user need is met?!


lozette:
You're right, we added the presenter but never exposed it in the exports 🤦

lozette (approved):

PR #1494, opened 2024-04-12
(161624) Service support users can "soft delete" a project

lozette:
Now that projects can have a "deleted" state (where they do not appear on the UI but are still in the database) we want to allow service support users to "soft delete" a project.

To do this, we have used a "delete" route as opposed to the usual "destroy". As a result, the delete routes are not the same as the destroy routes for notes and contacts. "Destroy" implies the record is removed from the database, while the projects are only hidden, not removed.

We use a "confirm delete" pattern to be absolutely sure the user wants to delete the project, as there is currently no way for a user to restore a "deleted" project.

Once a project is "soft deleted", the user is returned to the All Projects list page, with a confirmation the project was deleted.

Screenshot 2024-04-12 at 14 31 21

Screenshot 2024-04-12 at 11 31 03

Screenshot 2024-04-12 at 14 24 47

## Checklist

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -79,6 +79,10 @@ Rails.application.routes.draw do
         memberable
         academy_urn_updateable
       ]
+    resources :projects, except: :destroy do
+      get "confirm_delete", on: :member, action: :confirm_delete, as: :confirm_delete
+      patch "/", on: :member, action: :delete, as: :delete
+    end
   end
 
   # New projects by type
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased][unreleased]
 
+### Added
+
+- Service support users can "soft delete" a project
+
 ### Changed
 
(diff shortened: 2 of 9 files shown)


config/routes.rb
```diff
@@ -79,6 +79,10 @@
         memberable
         academy_urn_updateable
       ]
+    resources :projects, except: :destroy do
+      get "confirm_delete", on: :member, action: :confirm_delete, as: :confirm_delete
+      delete "/", on: :member, action: :delete, as: :delete
```
mec, line 84: I don't know if having this as `DELETE` is right? It doesn't actually delete anything?

mec (changes requested):
I am not sure about the delete verb in the routes - does it actually delete anything?

config/routes.rb
lozette: True, it neither deletes nor destroys! Should we just have a POST or PUT you reckon?

mec: Maybe `PATCH`? It updates the `state` value of the record with a fixed value?

lozette: Good call! PATCH it is.

mec (approved):
🙏 🚢

PR #1503, opened 2024-04-16
Release 64

lozette:
Added

- Service support users can "soft delete" a project

Changed

- The 'Add notes for ESFA task' has been disabled as there is evidence that the data is never used.
- some tweaks to content in the high needs places tasks to use the correct name for the notification of changes to funded high needs places form
- new form a multi-academy trust projects now require the new trust name to match any existing values with the same new trust reference number
- The 'confirm the academy's risk protection arrangement' task now includes additional guidance about auto enrollment.
- When the 'Confirm the academy's risk protection arrangements' task value is exported it is set to 'standard' when the task is incomplete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased][unreleased]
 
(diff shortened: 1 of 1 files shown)


mec (approved):
🚀

PR #1511, opened 2024-04-17
(162943) Split all projects in progress into tabs

mec:
We want to split the all in progress projects view up into different tabs so users can scope in progress projects to:

- all
- conversions only
- transfers only

Then later this is the home for form a multi academy trust projects.

This work updates the existing view so that it sits in a tab and then adds the conversions and transfers tabs.

Along the way we noticed a style bug on the MOJ sub navigation item, and fixed it.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased][unreleased]
 
+### Added
+
+- All in progress projects now includes at tab to view only conversion projects.
+
 ### Changed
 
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased][unreleased]
 
+### Added
+
+- All in progress projects now includes at tab to view only conversion projects.
```
lozette, line 11: Just a small thing but this should be in the previous commit

lozette (approved):

PR #1512, opened 2024-04-18
(162958) Form a multi academy trust show view

mec:
Schools and academies may come together to form a new Multi academy trust and users want to see these collections of projects reflected in the applicaiton.

We've creeated a simple object as an interface to this need and then added a show view for the object.

The `ProjectGroup` object loads projects with the same trust reference number so we can show them in a table.

The only gotya here is that we felt the same authorisation as projects applies to the project group, so we have to implement the `deleted?` method to use the `ProjectPolicy`.

The  `ProjectGroup` is so close to being an ActiveRecord model, I think it possibly should be, this would also make fetching multiple project groups faster (this is not implemented at this point).

## Screenshot (Fake data)
image

### DIFF
--- a/app/controllers/form_a_multi_academy_trust/project_groups_controller.rb
+++ b/app/controllers/form_a_multi_academy_trust/project_groups_controller.rb
@@ -1,0 +1,12 @@
+class FormAMultiAcademyTrust::ProjectGroupsController < ApplicationController
+  after_action :verify_authorized
+  rescue_from FormAMultiAcademyTrust::ProjectGroup::NoProjectsFoundError, with: :not_found_error
+
+  def show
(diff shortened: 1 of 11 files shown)


app/controllers/form_a_multi_academy_trust/project_groups_controller.rb
```diff
@@ -0,0 +1,12 @@
+class FormAMultiAcademyTrust::ProjectGroupsController < ApplicationController
+  after_action :verify_authorized
+  rescue_from FormAMultiAcademyTrust::ProjectGroup::NoProjectsFoundError, with: :not_found_error
```
lozette, line 3: I like this

lozette (approved):

PR #1519, opened 2024-04-18
(162946) Form a multi academy trust index view

mec:
This is the final step to expose all 'form a multi academy trust' project groups to users.

This work adds a service object to collate the project groups and an index view to show them.

Some of this design is a bit odd:

- showing a single region for a project group assumes all project regions are the same which we have no validation for, so we show the region from the first project
- showing a single assigned_to users for a project groups assumes all projects are assigned the same, which we have no way to enforce, so we show the user from the first project

We also hit a Bullet false positive which is documented in the commit, but we ended up skipping bullet on the spec as our best option - please let us know if there is a better way.

## Screenshot with fake data
image


lozette (approved):
I'm OK with the bullet thing 👍

PR #1523, opened 2024-04-19
(163286) Make the "Date the academy opened" task mandatory on Conversion projects

lozette:
## Changes

We want to make the "Date the academy opened" task mandatory on Conversion projects.

To do this, we have to add the task to the list of mandatory conditions & tasks in the project. Before adding it, we have refactored the `Conversion::Project.completable?` method slightly to make it easier to add more tasks/conditions.

Now when a user goes to complete a project, if the "Date the academy opened" task is not filled out, the user will see an error message telling them this task needs to be completed before the project is completed.

## Checklist

### DIFF
--- a/app/models/conversion/project.rb
+++ b/app/models/conversion/project.rb
@@ -22,6 +22,12 @@ class Conversion::Project < Project
   scope :with_academy_urn, -> { where.not(academy_urn: nil) }
   scope :by_conversion_date, -> { ordered_by_significant_date }
 
+  MANDATORY_CONDITIONS = [
+    :all_conditions_met?,
+    :confirmed_date_and_in_the_past?,
+    :grant_payment_certificate_received?
+  ]
+
   def grant_payment_certificate_received?
     user = assigned_to
     tasks = Conversion::Task::ReceiveGrantPaymentCertificateTaskForm.new(tasks_data, user)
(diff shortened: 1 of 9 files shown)


app/models/conversion/project.rb
```diff
@@ -22,6 +22,12 @@ def self.policy_class
   scope :with_academy_urn, -> { where.not(academy_urn: nil) }
   scope :by_conversion_date, -> { ordered_by_significant_date }
 
+  MANDATORY_CONDITIONS = [
```
mec, line 25: What do you think about this being `MANDATORY_TASKS` to make it closer to what it is doing?

mec (changes requested):
Could remove the `WIP` from the commit message on the second one.

I prefer a slightly different name for the constant of tasks.

I suspect we can take this one step further if we want to and have a method that can check the complete status of any tasks, but that can be a further optimisation/refactor.

app/models/conversion/project.rb
lozette: I was going to make it MANDATORY_TASKS but the "date in the past" isn't a task, I think? Or maybe it's pretty much a task even if it's a project attribute and I'm just splitting hairs

mec: Yeah, okay! the rename wasn't a blocker it was the WIP in the commit! 👍

mec (approved):
🚢

PR #1540, opened 2024-04-25
(163941) Refactor out the previous `Project.completed?` method

lozette:
We previously determined if a project was completed by seeing if the `completed_at` date was present. Since we added an enum value for Project.state, which includes `completed`, we get `Project.completed?` for free. The enum method looks to see if a project has a state of `completed`.

We forgot to remove the old `Project.completed?` method which looks at the value of `completed_at`. This was an oversight.

Remove this old method and let the enum value of `state` take over. When we introduced the enum, we did a data migration to update all projects with a `completed_at` date to have the `completed` state, so this refactor should not affect the live data.

## Checklist

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -107,10 +107,6 @@ class Project < ApplicationRecord
     @member_of_parliament ||= fetch_member_of_parliament
   end
 
-  def completed?
-    completed_at.present?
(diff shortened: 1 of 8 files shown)


mec (approved):
🚢

PR #1545, opened 2024-04-30
(164669) Remove the "By month" view from the Team section

lozette:
Under the "Team" section we had a By month view, which was divided into Revised and Confirmed project (by date). This section hadn't actually worked for a while - it was broken when we implemented the new By month section under All projects. We did not have adequate test coverage to show it was broken.

Because no users had complained about it being broken, we assume the section is not being used by users, and have decided to remove it.

## Checklist

- [x] Attach this pull request to the appropriate card in DevOps.
- [x] Update the `CHANGELOG.md` if needed.

PR #1546, opened 2024-04-30
(164686) Remove remaining "Confirmed" and "Revised" project listing code

lozette:
## Changes

To be merged after #1545

A while ago, we changed the All projects -> By month views to be split by
Conversions and Transfers, not Confirmed date & Revised date. The old Confirmed
& Revised code has been hanging around, so we should remove it.

Remove the old views, which were not linked from anywhere. Then remove the services which fetched projects based on their Confirmed or Revised dates.

## Checklist


mec (approved):
🚢

PR #1547, opened 2024-05-01
(164292) Update assignment user to project autocomplete

mec:
This work sets out with the seemingly simple task of taking our existing select boxes that get progressively enhanced into autocompletes and fixing an issue with them:

The issue we had is that as the number of users grows we are loading them ALL onto the page as the select options and any no-JS users would be faced with a cumbersome select box with hundreds of options.

The alternative is offered here. We swap the select for a text field that accepts a users email address. We validate the address is formatted correctly and the resulting user is appropriate for assigning to the project.

With that done, we use the full accessible autocomplete to call a new user search endpoint on the application providing real time suggestion based on what the user is typing.

There is a lot here becuase this is our first 'modern' JavaScript so we had to change a bunch of our tooling to support it.

The interaction is quite complex as well (at least it felt it to me!), this is essentially an edit form so we have to load the existing user from the email address to populate the autocomplete.

The actual submitted value is in a hidden field that is updated by the autocomplete, but this does mean the value is NOT updated with a 'no results found' or an empty field. This means the prior value is submitted, which can feel a little odd, but is probably acceptable for now - mainly because we do not allow the 'clearing' of the assigned user i.e. a project cannot go from assigend to a user to not assigened.


lozette (approved):
Nice work

PR #1552, opened 2024-05-07
Release 68

lozette:
### Changed

- Remove the "By month" view under Teams (viewable to users in a regional team or RCS only)


mec (approved):
🚀

PR #1559, opened 2024-05-08
(Spike) Use the Grape gem to build an API

lozette:
## Changes

Add the [Grape](https://github.com/ruby-grape/grape/) gem to enable us to add an API to the application. 

To demonstrate the API works, add a very simple "healthcheck" endpoint (cribbed from the [Grape demo app](https://github.com/ruby-grape/grape-on-rails)).

Next, show we can use Api key based authentication. Create a basic ApiKey model with a token & expiry date. Use this token in an "auth" endpoint to show we can authenticate users.

## Checklist

### DIFF
--- a/app/api/api.rb
+++ b/app/api/api.rb
@@ -1,0 +1,5 @@
+class Api < Grape::API
+  prefix "api"
+  format :json
+  mount V1::Healthcheck
+end
--- a/app/api/api.rb
+++ b/app/api/api.rb
@@ -1,6 +1,5 @@
 class Api < Grape::API
   prefix "api"
   format :json
-  mount V1::Healthcheck
-  mount V1::Auth
+  mount V1::Api
 end
--- a/app/api/api.rb
+++ b/app/api/api.rb
@@ -2,4 +2,5 @@ class Api < Grape::API
   prefix "api"
   format :json
   mount V1::Healthcheck
+  mount V1::Auth
 end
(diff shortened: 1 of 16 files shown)


db/migrate/20240509104059_create_api_keys.rb
```diff
@@ -0,0 +1,9 @@
+class CreateApiKeys < ActiveRecord::Migration[7.0]
+  def change
+    create_table :api_keys, id: :uuid do |t|
+      t.timestamps
```
mec, line 4: If we are going to the trouble of storing this should we add a 'description' of the keys so we can understand what is using the API?

app/api/api.rb
```diff
@@ -0,0 +1,5 @@
+class Api < Grape::API
+  prefix "api"
+  format :json
+  mount V1::Healthcheck
```
mec, line 4: I think a path based version would be better, for consistency with the other API if nothing else?

https://github.com/ruby-grape/grape?tab=readme-ov-file#path

mec (changes requested):
Couple of comments that would improve this for me.

app/api/api.rb
lozette: I've added the versioning as a separate commit as it was too complex to squash into the existing commits!

mec (approved):
Thank you! 🚢

PR #1574, opened 2024-05-13
(165402) Basic Api to create a Conversion project

lozette:
## Changes

Add a service object to create the minimum "bare bones" Conversion project in the application, without validations.

Then use the Api to create a conversion project from parameters, using the service object.

This is very rudimentary and results in a Conversion project which our platform considers invalid - but as the Prepare application (which will be posting to the endpoint) does not have all the required information, this is what we need to do.

Later on, we can edit these projects in the Complete application to add the required attribute values.

The required parameters for the POST are:

```
    conversion_project: {
                  urn: 123456,
                  incoming_trust_ukprn: 10066123,
                  advisory_board_date: "2024-1-1",
                  advisory_board_conditions: "Some conditions",
                  provisional_conversion_date: "2025-1-1",
                  directive_academy_order: true,
                  created_by_email: "email@education.gov.uk",
                  created_by_first_name: "Firstname",
                  created_by_last_name: "Lastname"
    }
```

## Checklist


app/api/v1/api.rb
```diff
@@ -4,5 +4,6 @@ class Api < Grape::API
 
     mount Auth
     mount Healthcheck
+    mount Conversions::ProjectsController
```
mec, line 7: I would like this to be named a just a resource, so `conversions` and then switched so projects is first, giving us a path of `api/v1/projects/conversions` and later `api/v1/projects/transfers` - it might just be me, but I am more comforatble with that, I think of them all as types of 'projects', we also model them this way.

mec (approved):
As a proof of concept this is solid, but we have a way to go. I am happy to put this in to move us forward, but I would like to see some changes before we consider this public.

No one can use the API without a API key so we have control of when it gets used.

It is raising lots of interesting problems for us, which is exactly what we need it to be doing!

PR #1591, opened 2024-05-16
Use validation on URN and UKPRN in the API project creation service

lozette:
## Changes

We use these validators to validate URN and UKPRN in the Create project form and Project model. Use them in the service object that creates projects via the API as well.

We also had to correct the URN in one of the API tests, as the original URN in that test (12345) was invalid. That test was not demonstrating param validation failure, so correct it to get the right test case!

## Checklist

### DIFF
--- a/app/services/api/conversions/create_project_service.rb
+++ b/app/services/api/conversions/create_project_service.rb
@@ -1,36 +1,58 @@
 class Api::Conversions::CreateProjectService
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveRecord::AttributeAssignment
+
   class ProjectCreationError < StandardError; end
 
+  attribute :urn
+  attribute :incoming_trust_ukprn
+  attribute :advisory_board_date
+  attribute :advisory_board_conditions
+  attribute :provisional_conversion_date
+  attribute :directive_academy_order
+  attribute :created_by_email
+  attribute :created_by_first_name
+  attribute :created_by_last_name
+
+  validates :urn, presence: true, urn: true
+  validates :incoming_trust_ukprn, presence: true, ukprn: true
+
   def initialize(project_params)
-    @project_params = project_params
+    super(project_params)
   end
 
   def call
     user = find_or_create_user
-    project = Conversion::Project.new(
-      urn: @project_params[:urn],
-      incoming_trust_ukprn: @project_params[:incoming_trust_ukprn],
-      conversion_date: @project_params[:provisional_conversion_date],
-      advisory_board_date: @project_params[:advisory_board_date],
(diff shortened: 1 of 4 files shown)


app/services/api/conversions/create_project_service.rb
```diff
@@ -1,26 +1,51 @@
...
+  attribute :advisory_board_date
+  attribute :advisory_board_conditions
+  attribute :provisional_conversion_date
+  attribute :directive_academy_order
+  attribute :created_by_email
+  attribute :created_by_first_name
+  attribute :created_by_last_name
+
+  validates :urn, presence: true, urn: true
+  validate :establishment_exists
```
mec, line 19: There is a decision here for me about how far we trust the data being posted from Prepare, in theory the establishment and trust MUST exist so do we need this? What is our stance, should we document it?

app/services/api/conversions/create_project_service.rb
```diff
@@ -1,26 +1,51 @@
...
+  validates :incoming_trust_ukprn, presence: true, ukprn: true
+  validate :trust_exists
+
   def initialize(project_params)
     @project_params = project_params
```
mec, line 24: I dont' think we need this? calling super assigns all the attributes and then we can just get them via their getters later?

app/services/api/conversions/create_project_service.rb
```diff
@@ -33,4 +58,24 @@ def call
   rescue ActiveRecord::RecordInvalid
     raise ProjectCreationError.new("Failed to save user during API project creation, email: #{@project_params[:created_by_email]}")
```
mec, line 59: We should be careful here, the email is PII and this may leak that into the logs?

app/services/api/conversions/create_project_service.rb
```diff
@@ -1,26 +1,51 @@
...
-      directive_academy_order: @project_params[:directive_academy_order],
-      regional_delivery_officer_id: user.id
-    )
-
-    if project.save(validate: false)
-      project
+
+    if valid?
+      project = Conversion::Project.new(
+        urn: @project_params[:urn],
```
mec, line 33: If we don't use `@project_params` we can just do `urn: urn` here and everywhere else we try and access `@project_params`.

mec (changes requested):
We're still in spike territory here I think (the branch is spike/), I think we know this can work and we should go away now and spend some time 'designing' what we want this to be so we can come at with a clear set of parameters?

app/services/api/conversions/create_project_service.rb
lozette: This is fair, happy to get rid of it tbh!!

mec (approved):
🚢

PR #1594, opened 2024-05-16
When creating the project via the API, also create an empty TasksData

lozette:
## Changes

When we create projects, we also create their TasksData. We need to do this when creating projects via the API, as there is no other way to create a TasksData object (not even if we manually edit the project).

## Checklist


mec (approved):
🚢

PR #1601, opened 2024-05-20
(167284) Update service support policies

mec:
This work allow Service Support users to update project and task data even after the project is completed.

It also fixes an issue where the 'complete project' button was shown on complted projects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - When creating a new project, two digit dates are no longer accepted which
   means incorrect dates cannot be submitted.
+- The 'complete project' button is no longer shown on completed projects.
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


lozette:
When a project is completed it has a banner saying "This project was completed on 20 May 2024 and cannot be changed.". I think we should check with @SteveMilnes if we need to change this text for Service Support users, as they *can* change it, and the banner might be confusing.

mec:
> When a project is completed it has a banner saying "This project was completed on date and cannot be changed.". I think we should check with @SteveMilnes if we need to change this text for Service Support users, as they _can_ change it, and the banner might be confusing.

Yeah, it says cannot be changed:

image

mec:
Could this be a simple as adding 'except by Service Support users' to this message?

SteveMilnes:
I think we should keep the banner so it's really obvious that this project is done with. But it's probably useful to call out that the service support user can make changes to the project, so it's clear to them while all this is new. 

We might find later on we don't need that anymore but I think a bit of reassurance and guidance here can help with changes and new functionality.

We can change the content so it just says:

> Important
> 
> Project completed
> 
> This project was completed on {DATE}.
> 
> As a Service Support team member, you can make changes to this project.

mec:
image

lozette (approved):

PR #1602, opened 2024-05-20
Put the newly-created user (from API) into a regional team

lozette:
When we create a new project via the API with a new (unknown to Complete) user, we were putting this new user into the Regional casework services team. This is most likely incorrect, as RCS do not use the Prepare service. Instead, make an assumption that the Prepare user is from the same region as the establishment, and put them in the same regional team.

If this is not correct, the new user can be amended by Service Support later on


mec (approved):
🚢

PR #1608, opened 2024-05-23
Update export header and value

mec:
We have heard from users that their preference for the
'sponsored_grant_type' column header must be 'Project route and
sponsored grant type' and when not applicable the value must be
'voluntary converter - not eligible'.

Our content designer came up with this solution after much discussion
with users.

## Changes

## Checklist


mec:
Closing becuase the ticket number changed so the branch name changed.

PR #1609, opened 2024-05-23
(152117) Change the "All conditions met" task for transfers

lozette:
## Changes

The "All conditions met" task should be called "Authority to proceed" for Transfers.

We have decided to keep the name of the attribute the same (`all_conditions_met`) for simplicity, but change the accompanying text. We have added an alias `authority_to_proceed` on the attribute.

We have also added two more checkboxes to this task. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - A simple POST Api endpoint to create a Form a MAT Conversion project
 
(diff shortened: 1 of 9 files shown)


mec (approved):
I do think we should stop doing data migrations in schema migratiions and use a one off service that we trigger manually, but we've done it so often now I suppose the precedent is set.

Good move on the alias! 👍

lozette:
Agree with you on the migrations, I did prevaricate on it. They make recreating the db a pain.

PR #1612, opened 2024-05-23
Disable autocomplete on uid fields

mec:
***Speculative work***

We have a number of fields where the user has to enter a unique
identifier, having autocomplete enabled on fields like these increases
the chance of errors at the cost of easily entering repeated uids, which
feels like a good trade off.

If a users enters an incorrect uids, the chances of doing so again also
increase as the browser will store the incorrect uid for
autocompletion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - The skip to content link is now underlined.
 - Service support users can now edit all contacts.
+- URN and UKPRN fields no longer autocomplete.
 
 ## [Release-70][Release-70]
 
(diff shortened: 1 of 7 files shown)


lozette:
I can totally see why you've done this but I know it will make my local testing more of a pain 😆 

Also I'm not 100% sure it's necessary, as entering an existing URN will generate a validation error when submitting the form. Duplicate UKPRNs won't be a validation error, though, so it could be useful there.

lozette (approved):

mec:
> Also I'm not 100% sure it's necessary, as entering an existing URN will generate a validation error when submitting the form. Duplicate UKPRNs won't be a validation error, though, so it could be useful there.

@lozette yep, it's not so much about the URN but the UKPRN for the trust - but for consistency I did both.

PR #1623, opened 2024-05-30
handle when UKPRNs are changed upstream

lozette:
## Changes

Background: We have an issue on Production where projects keep becoming invalid.
The UKPRN entered for the incoming trust, which was valid when the project was
created, is being changed upstream in the Academies DB. That change is then
being ingested into the Academies API, and when we try to load an existing
project, the UKPRN is not found and it significantly breaks the application
for users.

Because the Project model is "at rest", we do not think it needs validations.
Any Project creation or editing is done via FormObjects, which have their own
validations. We do not need to be validating the Projects on load into listing
pages, view pages or exports.

This PR enables the application to still work (load listing pages, show pages and exports) even if an invalid UKPRN is present in one of the projects.

This work is not ideal - ideally, noone should be updating UKPRNs in the Academies DB! - but we need to handle this reality and keep the app usable. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - URN and UKPRN fields no longer autocomplete.
 - When creating a new project and submitting an empty form, the errors are now
   shown correctly.
+- If a UKPRN is changed in the Academies DB, handle the exception so that the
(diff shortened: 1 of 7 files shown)


mec (approved):
This is going to be interesting! 

I know you know this, but we need to resolve this long term as this is sticking plaster, but fixing the app right now is the priority! 🚢

PR #1625, opened 2024-05-30
(168424) Update RPA, SUG and FA letters export

mec:
Users want to be able to select a date range for the conversion projects that
get included in the RPA, SUG and FA letters export.

This work enables that and attempts to make the code more conventional and
closer to the real world name and application of the export.

We namespace everything we can with `rpa_sug_and_fa_letters` which aligns with
the UI.

The idea here is that the export is a model of sorts (we use a from object as a
stand in) and we are `creating` a new export with date parameters, this goes as
far as using resourceful routing.

All the other exports will get the same treatment resulting in a set of clear
exports that encapsulate:

- the criteria for which projects are included
- the column exported
- any sorting order

Screenshot 2024-05-30 at 14-15-50 RPA SUG and FA letter


app/controllers/all/export/by_significant_date/transfers/academies_due_to_transfer/projects_controller.rb
```diff
@@ -5,7 +5,7 @@ def index
     @data = export_months.map do |month|
       {
         month: month,
-        count: service.conversion_projects_by_significant_date(month: month.month, year: month.year).count
+        count: service.transfer_projects_by_significant_date(month: month.month, year: month.year).count
```
lozette, line 8: 😱

lozette (approved):
😮‍💨 Nice one!

PR #1633, opened 2024-06-03
(168819) Ingest two extra fields into GIAS::Establishments

lozette:
## Changes

There has been a request to ingest two more fields into the Gias::Establishments table from the GIAS export - open date and establishment status name. These fields will not be added to any exports, they are just for the PowerBI reports.

## Checklist


mec (commented):
I may be missing some context here, but this makes me really uneasy!

We always said that ingesting GIAS data into complete was only ever going to be used to populate the new academy details and nothing else.

If DfE needs these two attributes they should go in via the Academies DB ingestion of GIAS data, otherwise we are creating to dependancies for essentially the same information and adding to the confusion in the database about where data gets sourced?

mec (approved):
Whilst I worry about the wider implications of this change, which we've talked about, the code changes are solid. 👍

PR #1635, opened 2024-06-03
Do not raise an error when the Members API does not return a constituency

lozette:
We found a bug with the exports where a project is in an unknown constituency. We thought we had all exceptions handled but evidently not this one - and this error is likely to become more frequent following the upcoming constituency boundary changes.

Handle an unknown constituency. Also flesh out the error message to return the response status and searched-for term, which will make future debugging easier.

## Checklist


mec (approved):
🚢

PR #1638, opened 2024-06-04
Do not return a Member of Parliament for projects

lozette:
Background: The Parliamentary Members API is not efficient enough to use when generating exports with large numbers of projects. Additionally, during the pre-election period, the API does not actually return any MP details so it is pointless for us to use a lot of processing power for no results.

Return `nil` for all MPs on all projects. Our presenter code can handle a nil MP record.

We have had to add the `nocov` instruction to the relevant implementation code as our CI rules expect 100% code coverage, and skipped tests do not count towards our code coverage.


mec (approved):
🚢

PR #1645, opened 2024-06-05
Front-end App Insights Telemetry

DrizzlyOwl:
## Changes

User telemetry can be made available in App Insights if the user opts into Analytical Cookies.

I have set the AuthenticatedUserContext which means we can track individual user sessions throughout their journey. In addition, I have also included the click tracking plugin which enables us to have almost feature parity with Google Analytics and Google Tag Manager.

I am not sure if my Ruby code is correct for printing the currently authenticated user's name into the JS. Similarly, I'm not sure if there is anything more I need to do to print the connection string for App Insights.


DrizzlyOwl:
Superseded by https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/1650

PR #1652, opened 2024-06-06
Add SignificantDateHistoryReasons

lozette:
## Changes

We will need to start recording a reason for a chaange to Significant Dates on
projects. These reasons will be taken from a list of predefined reasons in a
form. To begin, we add a SignificantDateHistoryReason model, which sits between
the SignificantDateHistory record and its Note.

To ensure we do not break the existing SignificantDateHistory functionality, we
need ot preserve the connection between SignificantDateHistory and Note. Once
the new SignificantDateHistoryReasons are in place, we can do a data migration
to break these connections and re-join the existing SignificantDateHistory
records to their Notes via a SignificantDateHistoryReason record.

## Checklist

### DIFF
--- a/db/migrate/20240606152236_create_significant_date_history_reason.rb
+++ b/db/migrate/20240606152236_create_significant_date_history_reason.rb
@@ -1,0 +1,11 @@
+class CreateSignificantDateHistoryReason < ActiveRecord::Migration[7.0]
+  def change
+    create_table :significant_date_history_reasons, id: :uuid do |t|
+      t.timestamps
+      t.string :reason_type, index: true
+      t.uuid :significant_date_history_id
+    end
+
+    add_column :notes, :significant_date_history_reason_id, :uuid
+  end
+end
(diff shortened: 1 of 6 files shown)


db/migrate/20240606152236_create_significant_date_history_reason.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSignificantDateHistoryReason < ActiveRecord::Migration[7.0]
+  def change
+    create_table :significant_date_history_reasons, id: :uuid do |t|
+      t.timestamps
+      t.string :reason_type
```
mec, line 5: Thinking that, like you said, this might be an enum, but using a string over an integer would be clearer for database poking friends anyway so this is still good.

Might be worth an index on `reason_type` as it could be common to query by type?

mec (changes requested):
Nice! I managed to get this working with the existing setup with a one line change! 🎉 

Couple of thoughts on here, let me know what you think?

mec (approved):
🚢

PR #1658, opened 2024-06-10
(169272) Significant date reasons

mec:
With the new `SignificantDateHistoryReason` model in place, we want any changes to significant dates to add a new reason.

We update the `SignificantDateCreatorService` and the two places it is currently used:

- the stakeholder kick off task for conversions and transfers
- any change to significant date

We also add a user to the date history as there is no longer a one to one association between a date history and the note which made delegating the user possible.

## This work includes a data migration

We also want to migrate the existing data to use a reason.

Once deployed, we need to run the `SignificantDateReasonsDataMigration`, see the file for details.

### DIFF
--- a/app/controllers/date_histories_controller.rb
+++ b/app/controllers/date_histories_controller.rb
@@ -2,7 +2,7 @@ class DateHistoriesController < ApplicationController
   include Projectable
 
   def index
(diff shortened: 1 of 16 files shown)


lozette (approved):
Let's do it :shipit:

PR #1663, opened 2024-06-12
(169569) Update and expose date history index

mec:
**_Based ontop of #1658 and should not be released until the data migration has been run in production!_**

With the new `SignificantDateHistoryReason` model in place and the data migrated over to use it, we can start showing users the current date history for a project ahead of adding in more options for them to provide reasons with notes.

This work focuses on the date histories index view, updating to use 'cards' and adding the header section that shows the current state and significant date.

image


lozette (approved):
🚀

PR #1666, opened 2024-06-13
Release 74

lozette:
Added

- Microsoft Application Insights browser tracking has been added to the application and will only be enabled for users who have opted in to optional cookies.
- The by month conversions export now includes the Academy URN and DfE/LAEST number.

Changed

- The academies due to convert export now includes the projects for a supplied date range.
- The pre transfer grants export now include the projects for a supplied date range.
- The loading time of the by local authority view has been improved, it has to load all projects so is still a slow view.
- Content updates to export pages following date picker additions
- The application footer now uses the programme wide layout.


mec (approved):
🚀

PR #1667, opened 2024-06-13
WIP: Two step date changing

mec:
WIP: see commits


mec:
This has moved onto a different branch now.

PR #1673, opened 2024-06-17
Don't include `linting` gems in the Dockerised application

lozette:
(Speculative!)

We have had an issue twice now where the latest version of StandardRB v1.37.0 has caused the deploying container to crash on Azure. Aside from attempting to figure out why this happens on deployments but not locally, we do not actually need to include Standard (and ERBLint) in our Dockerised version.

Exclude the gems in the `linting` group when building the application in Docker.

See previous PR https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/1665


DrizzlyOwl:
Does this tell Ruby not to include the linting specific gems?

lozette:
@DrizzlyOwl It does yep

DrizzlyOwl (approved):
LGTM

PR #1676, opened 2024-06-18
Create DAO Revocation Model

lozette:
## Changes

We are going to want to record a DAO Revocation for sponsored projects.
This will be a decision attached to a project, recording when the DAO was
revoked and who authorised the decision.

The DaoRevocation will be a 1:1 association on projects.

## Checklist

### DIFF
--- a/app/models/dao_revocation.rb
+++ b/app/models/dao_revocation.rb
@@ -1,0 +1,15 @@
+class DaoRevocation < ApplicationRecord
+  belongs_to :project
+
+  validates :date_of_decision, :decision_makers_name, presence: true
+  validate :conversion_project_with_dao
+  validate :at_least_one_reason
+
+  private def at_least_one_reason
+    errors.add(:base, :reason_required) unless reason_school_closed || reason_school_rating_improved || reason_safeguarding_addressed
+  end
+
+  private def conversion_project_with_dao
+    errors.add(:base, :incorrect_project_type) unless project.is_a?(Conversion::Project) && project.directive_academy_order
+  end
+end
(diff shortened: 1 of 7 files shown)


db/migrate/20240618140233_create_dao_revoked_decisions.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDaoRevokedDecisions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :dao_revoked_decisions, id: :uuid do |t|
+      t.timestamps
+      t.date :date_of_decision
+      t.string :decision_makers_name
+      t.boolean :reason_school_closed
+      t.boolean :reason_school_rating_improved
+      t.boolean :reason_safeguarding_addressed
+      t.references :project, index: true, foreign_key: true, type: :uuid
```
lozette, line 10: Maybe we don't need this as it'll be a 1:1 relation!

mec:
@lozette how would you feel about changing this model to `DaoRevocation`?

app/models/dao_revocation.rb
```diff
@@ -0,0 +1,15 @@
+class DaoRevocation < ApplicationRecord
+  belongs_to :project, dependent: :destroy
```
mec, line 2: This feels iffy to me, destroying the revocation destroys the parent Project? I would expect a Conversion::Project to declare a `dao_revocation` association with the destroy on it but not the other way around - am I missing some of the context or design?

Do you think we need to think about making it fast to scope Projects that have a DAO Revocation, perhaps by reversing this association, it feels wrong that a 'project belongs to a DAO revocation' but then we can do something like:

`Conversion::Project.where.not(dao_revocation_id: nil)`

to get all DAO project?

mec (changes requested):
Not 100% about the assocation on this.

app/models/dao_revocation.rb
lozette: No putting the destroy on the wrong bit was my fault!

mec (approved):
🚢 Nice!

PR #1678, opened 2024-06-19
(170377) Reasons for change of significant date

mec:
With the SignificantDateHistoryReason in place, this work adjusts the user journey to collect one of a fixed number of 'reasons' along with the details.

We had to break things and then put them back together working so this PR is a little large - apologies, we couldn't see a way to deliver anything smaller, mostly because we resuse across earlier dates, later dates, conversions and transfers.

As there are only two steps and the first collects one attribute, we kept this simple and render the second step from the first with the already initialised objects, this has the trade off that the path doesn't change as you might expect.

Becuase the form objects don't actually exist we couldn't use the GOV.UK form builder to render the big reasons forms, this meant doing it by hand, which made us very grateful for the form builder, it was a lot of work to get the errors working! Our solution is a little rough around the edges, particualrly setting the correct id for the error, alternatives welcome!

Like out task list, this is built in a way that any changes to the reasons is very easy to deliver, developers can add the changes to the arrays of reasons and things will mostly update.


config/locales/significant_dates.en.yml
```diff
@@ -19,3 +16,62 @@ en:
...
+      stakeholder_kick_off: Confirmed as part of the stakeholder kick off task
+      progressing_faster_than_expected: Project is progressing faster than expected
+      correcting_an_error: Correcting an error
+      incoming_trust: Incoming trust
+      outgoing_trust: Outgoing trust
+      academy: Academy
+      school: School
+      local_authority: LA(local authority)
+      diocese: Diocese
+      tupe: TuPE (Transfer of Undertakings Protection Employement rights)
```
lozette, line 30: Typo in Employment

config/locales/significant_dates.en.yml
```diff
@@ -19,3 +16,62 @@ en:
...
+      enter_details: Enter details about the reason for the change.
+      new_date_history_reasons_new_later_form:
+        correcting_an_error: Correcting an error
+        incoming_trust: Incoming trust
+        outgoing_trust: Outgoing trust
+        academy: Academy
+        school: School
+        local_authority: LA(local authority)
+        diocese: Diocese
+        tupe: TuPE (Transfer of Undertakings Protection Employement rights)
```
lozette, line 53: Typo in Employment

app/controllers/date_history/reasons/earlier_controller.rb
```diff
@@ -0,0 +1,29 @@
+class DateHistory::Reasons::EarlierController < ApplicationController
+  include Projectable
+
+  def new
+  end
+
+  def create
```
lozette, line 7: As mentioned - yes I think these two need authorization, same as the DateHistoryController

lozette (changes requested):

lozette (approved):

PR #1684, opened 2024-06-20
(Fix) Pick the Headteacher contact for "school main contact" where possible

lozette:
## Changes

We ingest Headteacher contact details where possible from GIAS records as `Contact::Establishment` records.

However, due to an oversight we were not marking those contacts as `school_or_academy` contacts (via the `category` attribute). When picking the school main contact for exports, we first try to get the "main contact" (as chosen via a select box on the application), and if none is present then we get the first `school_or_academy` contact. 

Because the Headteacher records were not being marked as `school_or_academy` they were never being chosen.

Fix this via:

- Ensuring all new Contact::Establishment records are always category `school_or_academy`
- Ensuring that when picking the "first" school_or_academy contact, Contact::Establishment records are picked first
- Updating all existing Contact::Establishment records to have the `school_or_academy` category

## Checklist

### DIFF
--- a/app/services/contact_establishment_data_migration.rb
+++ b/app/services/contact_establishment_data_migration.rb
@@ -1,0 +1,19 @@
+# Data Migration service
+# This service is designed to be used once to update the category for all Contact::Establishment records
+#
+# Usage:
+#
+# `bin/rails runner "ContactEstablishmentDataMigration.new.migrate!"`
+#
+class ContactEstablishmentDataMigration
+  def migrate!
+    contacts = Contact::Establishment.all
+    puts "#{contacts.count} Contact::Establishment records to update"
+
+    contacts.each do |contact|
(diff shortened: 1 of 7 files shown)


app/services/contact_establishment_data_migration.rb
```diff
@@ -0,0 +1,19 @@
...
+# Usage:
+#
+# `bin/rails runner "ContactEstablishmentDataMigration.new.migrate!"`
+#
+class ContactEstablishmentDataMigration
+  def migrate!
+    contacts = Contact::Establishment.all
+    puts "#{contacts.count} Contact::Establishment records to update"
+
+    contacts.each do |contact|
```
mec, line 13: You could just `update_all`: https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-update_all

mec (approved):
🚢

PR #1692, opened 2024-06-25
(168445) DAO Revocation flow

mec:
A DAO revocation is applied to a Conversion project that has a DAO applied to it.

The design is a 'check your answers' stepped form. This work delivers the basic flow but does not expose it in the application, that will follow.

There are four 'steps' here plus teh check your answers view:

- start page: infomation only
- confirm page: collects responses but does not save them
- reasons: collects up to three reasons for the revocation
- minister name: the name of the minister who approved the revocation
- date: the date on which the revocation was approved

We created a form object to handle everyting except the start page, which is just a action and view.

We do not want to create the actual DaoRevocation in the database until the check answers view is confirmed, so we store the temporay values in the session, we felt this was the right place as the confirmation page is framed as the user confirming they have taken the steps, so it is quite personal. Whilst orphaned data in the session is possible, the session should be deleted by the browser on close, so this is acceptable.

Coming back to the start, the session will be cleared and so the flow must be restarted. The values are stored in the session keyed by the project ID so multiple projects can be stored, there is a theoretical limit to the number of values that can be stored in the session, but as we only store five simple data types, we think this is fine, it's worh noting that DAO revocation at this stage in the overall project life cycle are unlikely.

We have the concept of 'checkable' which is the form being in an acceptable state for submission of changing, if the check answer view is requested before this state is reached, the user is redirected back to the start.

As a user could (however unlikely) make request to the steps manually, the check answer view handles an empty state and links to each steps so that a user can complete the task, although restarting is the preferred route.

We check at the start of the journey that no DaoRevocation is present on the current project, we do not check the project type or that a DAO is present - we may want to add this with the work that exposes this steps to users, we want to try and keep the complexity here down for review!

### DIFF
--- a/app/controllers/dao_revocations_controller.rb
+++ b/app/controllers/dao_revocations_controller.rb
@@ -1,0 +1,132 @@
+class DaoRevocationsController < ApplicationController
+  include Projectable
+
+  before_action :redirect_if_project_revoked
+  before_action :redirect_if_invalid_step, only: %i[step update_step change_step update_change_step]
+
+  def start
+    authorize @project, :dao_revocation?
+  end
+
+  def step
+    authorize @project, :dao_revocation?
+    if requested_step.eql?(DaoRevocationSteppedForm.first_step)
+      delete_session
+    end
+
+    @step_form = DaoRevocationSteppedForm.new
+
+    render requested_step.to_s
+  end
+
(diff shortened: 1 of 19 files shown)


lozette (approved):
V happy with this 🚀

PR #1723, opened 2024-07-02
(Fix) Update Project state after DaoRevocation destruction

lozette:
When fiddling with test data, we realised that if a DaoRevocation is destroyed, it causes the parent project to become unavailable in the UI, because the DaoRevocation does not exist but the project's state is still :dao_revoked.

This is a rare occurrence as DaoRevocations probably won't be deleted, but it may happen if we're asked to remove one manually. This ensures the parent project does not become inaccessible.


mec (approved):
🚢

PR #1728, opened 2024-07-03
(Chore) Remove funding agreement contact ID

lozette:
This association hasn't been used for a long time. Remove it to keep the codebase tidy.


mec:
@lozette there is no migration file?

lozette:
@mec There is now that I remembered to add it 🤦

mec (approved):
🚢

PR #1744, opened 2024-07-09
Configuration for HSTS Header

DrizzlyOwl:
The HSTS Header requires that we set the max-age property when using preload. Additionally, the 2024 ITHC recommends we also add the includeSubDomains property too.


mec (approved):
🚢

PR #1748, opened 2024-07-11
Get a project's MP by postcode

lozette:
## Changes

On May 31st 2024 constituency boundaries changed in the UK.

We previously retrieved an MP for a project by the establishment's constituency. We get the constituency from GIAS. There are currently no plans to update the constituencies in GIAS for establishments. This breaks our ability to get an MP by constituency.

Instead, we can get an MP for an establishment using the establishment's postcode. 

Refactor the Members API work to retrieve an MP for a project by establishment postcode.

Note that the MP details are not currently being exported due to performance concerns with the Parliamentary Members API, so none of this is user-facing work. 

To view an MP for a project, you can manually change the URL for a project to be `<project_id>/mp` and see the correct MP for an establishment.

### DIFF
--- a/app/controllers/member_of_parliament_controller.rb
+++ b/app/controllers/member_of_parliament_controller.rb
@@ -5,47 +5,27 @@ class MemberOfParliamentController < ApplicationController
   rescue_from Api::MembersApi::Client::Error, with: :members_api_client_error
 
   def show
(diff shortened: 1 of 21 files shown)


lozette:
We may not need this after all as apparently GIAS is being updated

PR #1779, opened 2024-07-25
Articles of association content changes

lozette:
## Changes

Add a new "sent" field to the Articles of Association tasks for both Conversions and Transfers. Amend the content to reflect the process more accurately.

## Checklist

### DIFF
--- a/config/locales/conversion/tasks/articles_of_association.en.yml
+++ b/config/locales/conversion/tasks/articles_of_association.en.yml
@@ -30,6 +30,8 @@ en:
           title: Received
         cleared:
           title: Cleared
+        sent:
+          title: Sent cleared version to trust's solicitors
         signed:
           title: Checked with solicitors that incoming trust have signed the articles
           hint:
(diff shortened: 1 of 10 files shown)


config/locales/conversion/tasks/articles_of_association.en.yml
```diff
@@ -19,9 +30,13 @@ en:
           title: Received
         cleared:
           title: Cleared
+        sent:
+          title: Sent cleared version to trust's solicitors
         signed:
-          title: Signed by school or trust
+          title: Checked with solicitors that incoming trust have signed the articles
+          hint:
+            html: <p>It can be helpful to check the Companies House website to make sure the signed articles have been uploaded</p>
```
mec, line 38: Would it be good to make `Companies House website` a link to the site?

mec (changes requested):
I would like to see the locales for the new task actions added with them rather than before.

I also left a comment about a tiny improvement that might be worth adding.

mec (approved):
🚢

PR #1780, opened 2024-07-26
Add "project main contact" details to RPA/SUG report

lozette:
A user reported that they were adding a contact as the "Project main contact" but that contact was not being output to the RPA/SUG report. Despite the "select the main contact for a project" task
indicating that "this person will receive funding letters", the main contact for a project was not actually being included in the report.

As a temporary fix, add this contact into the RPA/SUG report. Also update the column name to make it (hopefully) more obvious this is the *Project* main contact. 

When we rework the contacts pages, this commit will become redundant, as users will be able to choose the contacts output to the RPA/SUG report.


mec (approved):
🚢

PR #1784, opened 2024-07-31
(174872) include form a mat in by trust view

mec:
Users want to see both 'regular' trusts and 'form a multi academy' trusts in the all projects by trust view.

This is a little messy because we do not model the form a MAT trust, it is simply an attribute of a `Project`.

We did not want to change this by adding a model at this point, so we work around it. Full details are in the commit, but in summary we treat each type differently and combine them before sorting, this requires adding more code to handle this in actions, routes and views.

### DIFF
--- a/app/services/by_trust_project_fetcher_service.rb
+++ b/app/services/by_trust_project_fetcher_service.rb
@@ -26,10 +26,6 @@ class ByTrustProjectFetcherService
     @trusts + @form_a_mat_trusts
   end
 
-  private def all_projects_grouped_by_identifier
-    @all_projects_grouped_by_ukprn.merge(@all_projects_grouped_by_trust_reference_number)
-  end
-
   private def projects_grouped_by_ukprn
     Project.active.not_form_a_mat.group_by(&:incoming_trust_ukprn)
   end
--- a/app/services/by_trust_project_fetcher_service.rb
+++ b/app/services/by_trust_project_fetcher_service.rb
@@ -1,21 +1,41 @@
 class ByTrustProjectFetcherService
-  def call
-    projects = projects_by_trust
+  def initialize
+    @all_projects_grouped_by_ukprn = []
+    @all_projects_grouped_by_trust_reference_number = []
 
-    if projects
-      trusts = get_trusts(projects.keys)
-
-      sort_trust_objects_by_name(build_trust_objects(trusts, projects))
-    else
-      []
-    end
+    @trusts = []
+    @form_a_mat_trusts = []
   end
 
-  private def projects_by_trust
-    projects = Project.active.not_form_a_mat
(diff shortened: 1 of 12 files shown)


config/routes.rb
```diff
@@ -176,7 +176,8 @@
           end
           namespace :trusts do
             get "/", to: "projects#index"
-            get ":trust_ukprn", to: "projects#show", as: :by_trust
+            get "/ukprn/:trust_ukprn", to: "projects#show", as: :by_trust
```
lozette, line 179: Small thing: I think this would be better described as `:by_trust_ukprn`

app/services/by_trust_project_fetcher_service.rb
```diff
@@ -1,38 +1,70 @@
...
 
-  private def projects_by_trust
-    projects = Project.active.not_form_a_mat
-    return false unless projects.any?
+  private def all_projects_grouped_by_identifier
+    @all_projects_grouped_by_ukprn.merge(@all_projects_grouped_by_trust_reference_number)
+  end
+
+  private def projects_grouped_by_ukprn
+    Project.active.where(new_trust_reference_number: nil).group_by(&:incoming_trust_ukprn)
```
lozette, line 30: Would it be worth adding a scope for form a mat projects, and not form a mat projects? If you add `Project.form_a_mat` I believe you get `Project.not_form_a_mat` for free. Although might not be worth doing if this is the only place we'll ever use the scopes.

mec: `Project.not_form_a_mat` is a scope, but yeah, good shout.

mec: > Project.form_a_mat I believe you get Project.not_form_a_mat for free

TIL! ❤️

mec: Oh wait! `not_form_a_mat` is a declared scope on Project, the partner `form_a_mat` is not available for free!? :(

lozette: Oh I'm not sure! I didn't realise there was already a scope! 🤔

mec: I've added the scope - scopes are better - thank you!

spec/models/project_spec.rb
```diff
@@ -770,6 +770,18 @@
         expect(Project.not_form_a_mat).to include(single_converter_project)
       end
     end
+
+    describe "form_a_mat scope" do
+      it "returns only NON form a MAT projects" do
```
lozette, line 775: This should be "returns only form a MAT projects" I think?

lozette (approved):
Just one annotation to change!

PR #1786, opened 2024-08-01
Release 81

lozette:
Added

- The "project main contact" (selected via a task) is output to the RPA/SUG export

Changed

- content changes to articles of association task in conversions and transfers to describe the process more accurately
- addition of 1 checkbox to articles tasks in conversions and transfers
- the all project by trust view now includes 'form a multi academy trusts'


mec (approved):
🚀

PR #1787, opened 2024-08-02
(175117) Collect group reference number

mec:
Projects can be grouped together for advisory board to receive additional funding.

This work adds the group reference number to both create conversion and transfer
projects. The group reference number originates in the Prepare application and
so, for now is a matter of copy and pasting the value from one application to
the other (one day the applications should be connected).

We validate the group reference number format and that the project incoming
trust UKPRN matches that of the group (this grouping is by trust).

As the behaviour is shared, we add it to the base `create_project_form`.

Right now, we do not think form a MAT projects needs this behaviour.

### DIFF
--- a/config/locales/transfer_project.en.yml
+++ b/config/locales/transfer_project.en.yml
@@ -46,6 +46,7 @@ en:
       transfer_create_project_form:
         new_trust_reference_number: Trust reference number (TRN)
         new_trust_name: Trust name
+        group_id: Group reference number
     legend:
       transfer_project:
         advisory_board_date: Date of advisory board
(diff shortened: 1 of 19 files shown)


.erblint-rubocop20240802-77069-ye9hf4
```diff
@@ -0,0 +1,7 @@
+---
+require: standard
```
lozette, line 2: Should this be in this commit?

config/locales/transfer_project.en.yml
```diff
@@ -93,6 +94,10 @@ en:
         establishment_sharepoint_link: Provide a link to the SharePoint folder for this academy. This is where you save all the relevant academy documents.
         incoming_trust_sharepoint_link: Provide a link to the SharePoint folder for the incoming trust. This is where you save all the relevant trust documents.
         outgoing_trust_sharepoint_link: Provide a link to the SharePoint folder for the outgoing trust. This is where you save all the relevant trust documents.
+        group_id_html:
+          <p>If this academy is transfering  as part of a group, enter the group reference number.</p>
```
lozette, line 98: Typo: `transferring` has two rs

lozette (changes requested):

.erblint-rubocop20240802-77069-ye9hf4
mec: Nope! I was thinking today how can we get these to go into `tmp`...

lozette (approved):

PR #1789, opened 2024-08-02
(175119) Group reference number editing

mec:
Whilst users have to supply the group reference number, there is a chance it
could be incorrect, so we want to allow it to be changed.

We already have a pattern for this with the concept of 'edit project form' for
both conversion and transfers.

This work extracts the common behaviour across those forms and then adds in the
group reference number for editing.

Editing the group reference number needs to allow it to be set to nothing, i.e.
the association to the group cancelled, this is achieved by submitting an empty
value.


lozette (approved):
👨‍🍳 💋

PR #1792, opened 2024-08-06
(175120) Views for project groups

mec:
Projects are grouped together to go to the advisory board to receive additional funding. We have added a model for this `ProjectGroup` and allowed new projects to provide the reference number from the Prepare application and edit the number. With a number in place, the project is grouped with all other projects with the same group reference number.

This work adds a index view of all these project groups and a show view of a individual group.

We also add an indicator on a project to show it is part of a group and link from the project to the group.


lozette (approved):
🚢

PR #1793, opened 2024-08-06
(175578) No longer import Headteacher data from GIAS

lozette:
We are trying to simplify the Contacts data and workflow.

In order to try to make sure the Headteacher contacts are as accurate as possible, we are going to stop importing them from GIAS and expect the users to input this data themselves.

A first step is to remove the headteacher importing service.


mec (approved):
🚢

PR #1803, opened 2024-08-13
(175554) New contact creation flow

lozette:
## Changes

We want to ensure users create contacts with a very specific job title, for
example "Headteacher" for whoever is the "headteacher-like" person at a school,
or "CEO" for the person in charge of a trust. Because we previously allowed
users to type anything into the Job title/role field for contacts, there was
confusion for end users (usually export users) over which users have which
roles, and who to send certain types of paperwork to.

This commit steers users to create contacts of a certain role and predetermines
their job title. For example, the Headteacher's job title will always be
"Headteacher".

Users can still create "other" contacts with different job titles in all
categories, but we hope this creation flow will encourage them to always
create at least *some* contacts with the expected job titles.

Screenshot 2024-08-16 at 15 04 57

## Checklist


app/forms/contact/create_chair_of_governors_contact_form.rb
```diff
@@ -0,0 +1,24 @@
+class Contact::CreateChairOfGovernorsContactForm < Contact::CreateProjectContactForm
+  attribute :category, default: "school_or_academy"
```
mec, line 2: I don't think we need this? It's declated on the base form and we hard wire it's value in the `save` method.

app/forms/contact/create_chair_of_governors_contact_form.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def save
```
mec, line 4: Depending on your taste, we could loose duplicating the `save` method and overide the getters, the subclasses would then be:

```ruby
  def category
    "school_or_academy"
  end

  def title
    "Headteacher"
  end

  def organisation_name
    nil
  end
```

When `organisation_name` is nil we default to the establishment name, it could be:

```ruby
  def organisation_name
    @project.establishment&.name
  end
```

if we want to be more obvious.

app/controllers/external_contacts_controller.rb
```diff
@@ -7,17 +7,37 @@ def index
...
+    end
   end
 
   def create
     authorize @project, :new_contact?
-    @contact = Contact::CreateProjectContactForm.new(contact: Contact::Project.new, project: @project, args: contact_params)
 
-    if @contact.save
-      redirect_to project_contacts_path(@project), notice: I18n.t("contact.create.success")
+    params = @project.is_a?(Conversion::Project) ? conversion_contact_params["contact_type"] : transfer_contact_params["contact_type"]
```
mec, line 21: Could we extract this to a method to better describe what it is doing, something like:

`def contact_type_from_params`

app/forms/contact/create_project_contact_form.rb
```diff
@@ -66,6 +66,25 @@ def save
...
+    case category
+    when "school_or_academy"
+      @project.establishment.name
+    when "incoming_trust"
+      @project.incoming_trust.name
+    when "outgoing_trust"
+      @project.outgoing_trust&.name
+    when "local_authority"
+      @project.local_authority&.name
+    end
```
mec, line 85: Should we default to 'other' if only to keep SonarCloud happy?

app/views/external_contacts/edit.html.erb
```diff
@@ -6,27 +6,32 @@
...
+  <div class="govuk-grid-column-full">
+    <span class="govuk-caption-l"><%= t("contact.edit.title") %></span>
+    <h1 class="govuk-heading-l"><%= t("helpers.label.new_contact_form.enter_contact_details") %></h1>
+  </div>
+</div>
 
 <div class="govuk-grid-row govuk-body">
   <div class="govuk-grid-column-two-thirds">
-    <%= form_for @contact, url: project_contact_path(@project), method: :put do |form| %>
+    <%= form_for [@project, @contact], url: project_external_contacts_other_contact_path(@project, @contact) do |form| %>
```
mec, line 18: I am not sure why we need `@project` here?

app/views/external_contacts/other_contacts/new.html.erb
```diff
@@ -0,0 +1,41 @@
...
+<div class="govuk-grid-row govuk-body">
+  <div class="govuk-grid-column-full">
+    <span class="govuk-caption-l"><%= t("helpers.label.new_contact_form.subtitle") %></span>
+    <h1 class="govuk-heading-l"><%= t("helpers.label.new_contact_form.enter_contact_details") %></h1>
+  </div>
+</div>
+
+<div class="govuk-grid-row govuk-body">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for [@project, @contact], url: project_external_contacts_other_contact_path(@project, @contact) do |form| %>
```
mec, line 18: Same here. What does passing `@project` to `form_for` here do?

mec (changes requested):
I have offered an alternate implementation that reduces the code, see what you think?

I am seeing a bug where editing a contact duplicates it, I don't think it exists on main, I had a quick nose but couldn't see why.

One overall thing I noticed is that a user can edit a contact and change the role to something non-standard which would negate all the effort (as I understand it) here - is there a plan to handle that next?

app/forms/contact/create_chair_of_governors_contact_form.rb
lozette: This is great, I had over-egged it a bit

app/forms/contact/create_project_contact_form.rb
lozette: Unfortunately if we add an else clause here with "nil" for the organisation name (organisation name can be blank), standard.rb removes it as it's redundant! I think we should maybe add an exception is sonarcloud, as there's no reason why we can't have an empty organisation name.

lozette:
Identified the edit duplication bug - I had the wrong method on the edit form 🤦

app/forms/contact/create_project_contact_form.rb
mec: SonarCloud is wrong in my view, so happier anyway! 👍

mec (approved):
🚢

PR #1813, opened 2024-08-19
Update Dockerfile

mec:
Github started to tell us about these updates with it's new 'Unchanged
files with check annotations' feature, so this commit fixes them.

We also had a GitHub security(?!) 'Arguments in long RUN instructions
should be sorted' so we sort that out here as well.


lozette (approved):
🤷 🚢

PR #1816, opened 2024-08-21
(177562) Use the Persons API for MP contacts

mec:
This works sets out to restore the MP contacts to the application as the Members API was not working for us.

DfE has introduce a centrally managed source of MP contacts that will work for us, using GIAS constituencies and have begun the work to expose it on the 'Persons API'

Here we create a client to use the new API and replace the two places we want to show MP contacts:

- a test page on each project at `project/id/mp`
- in the RPA, SUG and FA letters export

Right now there is no production API so we cannot release this work but once there is we should have everything in place to quickly restore the MP contacts.

None of these feature will work now as we have hard wired `Project` to return `nil` for `member_for_constituency`. Because of this we have to remove some code from the coverage report as it never gets called and skip some specs until the feature can be enabled.

We will be able to test the client itself on the Rails console in development once this work is deployed and the environment variables are set.


lozette (approved):
So much nicer 🥲

PR #1818, opened 2024-08-21
Do not allow an outgoing trust contact to be created for Conversion projects

lozette:
Conversion projects don't have the concept of an outgoing trust, so creating an outgoing trust contact for them doesn't make sense. Show a validation error if the user tries to create one.


mec (approved):
🚢

PR #1824, opened 2024-08-23
(Chore) Split the "manage contacts" feature spec up

lozette:
This spec had become too big and unwieldy; split it up into 3 smaller feature specs.


mec (approved):
🚢 Nice one 👍

PR #1825, opened 2024-08-23
(Fix) Remove non-unique ID from Contact Summary Card

lozette:
The contact summary card's IDs were not unique. We were going to add unique ones but 1) the IDs are not used in any tests; and 2) an ID is not mandatory according to the design system. So instead we have removed them.


mec (approved):
🚢

PR #1828, opened 2024-08-27
Release 84

lozette:
### Changed

- changing 'conversion grant' references to 'conversion support grant' to match
  Apply and Prepare
- Headteacher contacts imported from GIAS are transformed into Project contacts
  and behave like other Project contacts


mec (approved):
🚀

PR #1831, opened 2024-08-27
Docker image improvements

DrizzlyOwl:
Splits the 'Test' image into it's own Dockerfile to ensure we are following SRP. The GitHub Workflow is adapted so that it will cache both a 'deployable' image, and a 'test' image.

The 'test' image builds off of the main app image


mec:
I am going to close this as it is getting stale and I have new work on the Dockerfile incoming.

PR #1832, opened 2024-08-27
Refine the Headteacher data migration

lozette:
When running the Headteacher migration on test, we found two issues:

1) The Academies API timed out. Get around this be getting the contact establishment name (school name) from our local GIAS:Establishment tables.

2) There is a risk that we will try to create a new contact with the same email address as an existing contact. Skip the contact if this is the case.


mec (approved):
🚢

PR #1834, opened 2024-08-28
CVE-2021-31799: A command injection vulnerability in RDoc

DrizzlyOwl:
There is an open CVE for the RDoc Gem. Whilst the app doesn't use it directly, `irb` has it as a dependency. This PR simply bumps the version constrained in the `Gemfile.lock` so that the patched version is installed.

[1] https://www.ruby-lang.org/en/news/2021/05/02/os-command-injection-in-rdoc/


mec (approved):
Wow! That is a big jump?! 🚢

PR #1838, opened 2024-08-28
(178320) Add a new "Data consumers" team

lozette:
We are going to consolidate the EFSA & APOU users into one "can export" team.

First of all, create a new team these users will eventually belong to and give the team the same permissions as existing ESFA & AOPU teams.


mec (approved):
🚢

PR #1840, opened 2024-08-29
Migrate users in the AOPU and ESFA teams into the "data consumers" team

lozette:
We are consolidating all ESFA and AOPU users into a new "data consumers" team. This migration moves all ESFA and AOPU users innto the new team.


mec (approved):
🚢

PR #1842, opened 2024-08-29
Update to Rails 7.1

mec:
Take the steps needed to update to Rails 7.1 to prepare a path to get to 7.2.

We didn't hit too many issues, a couple of new defaults need work for us and the
new transaction behaviour will take some effort so for now we have opted out, to
get to 7.2 we will have to fix this, we can do so by opting in and fixing with:

`config.active_record.commit_transaction_on_non_local_return = false`

We don't foresee any issue with 7.1 so have committed to the new caching with:

`config.active_support.cache_format_version = 7.1`

and we don't use the cache for anything that would cause us issue, worst case a
users has to start the stepped DAO revocation process again if midway through
during the deployment.


config/environments/development.rb
```diff
@@ -6,7 +6,7 @@
   # In the development environment your application's code is reloaded any time
   # it changes. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = false
+  config.enaable_reloading = true
```
lozette, line 9: typo here in `enable`

lozette (changes requested):

PR #1846, opened 2024-08-30
Release 85

lozette:
Changed

- the actions for the 'Commercial transfer agreement' task for conversion projects, any completed actions will be migrated over, users may be required to confirm additional actions after this change.
- the actions for the 'Commercial transfer agreement' task for transfer projects.
- notes added during a change of significant date, including during the stakeholder kick off task now only appear under the date history tab of a project and not the project notes tab.

Added

- Add a new "Data consumers" team which will eventually replace the ESFA and AOPU teams


mec (approved):
🚀

PR #1860, opened 2024-09-04
Update env.example

mec:
With the Persons API going into production, we should document the two
variables.


lozette (approved):

lozette:
You can remove MEMBERS_API_HOST as well

PR #1862, opened 2024-09-04
Release 87

lozette:
### Changed

- the Member of parliament for a school is now included in the RPA, SUG and FA letters export when one can be found via the new Persons API.


mec (changes requested):
Should we put a note into the change log to say why we have this empty looking release?

mec (approved):
My bad! There is a note about the change motivating the release! 🚀

PR #1863, opened 2024-09-05
Remove the MP details from the Conversion "by month" exports

lozette:
Adding the MP details to exports affects the performance. Users don't need the MP details in this export anyway, so remove them.


mec (approved):
🚢

PR #1871, opened 2024-09-10
Add OAuth to Persons API client

mec:
***This work should not be merged until the variables are added to each environment***

The Persons API will soon move to OAuth only so we need to add that to
the client.

We use the 'client_credentials' grant type, which involves using a
secret and id to get an auth token. In the case of the Persons API this
is from an Azure identity platform instance at DfE where the Complete
app has rights.

We make an API call to the identity platform passing along the required
parameters and get back a token.

The token is then stored in the Rails cache as this is a secure Redis
instance. The token has a 1 hour expiry set, so we set the same expiry
on the cache minus 1 minute, this lets the application use the same
token throughout it's lifetime renewing just before expiry if needed.

Once the token expires another one is fetched and cached.

There are quite a few environment variables to use here, we've added
them to both the example.env and test.env.

We also have to provide another fake connection for the identity
platform api calls for testing.

Right now if the identity platform request fails for any reason, we
raise an error, we may want to log something and return empty data and
let the application carry on, but that can come later.


lozette (approved):
Nice 💯

mec:
We should not merge this until the Persons API is deployed to production and the required environement variables are setup.

PR #1872, opened 2024-09-10
Identify ourselves when sending requests to Academies API

DrizzlyOwl:
## Changes

Adds the `User-Agent` HTTP Header to the faraday http client when interacting with the Academies API.

This will help with segmenting traffic reports for Academies API

## Checklist


mec (approved):
We don't strictly log non-user facing changes - but all good! 🚢

PR #1875, opened 2024-09-12
Allow operators to override the default minimum number of app Containers

DrizzlyOwl:
The default number of containers is `1` but in Production we occasionally run into issues with Container App host migrations. Having `2` minimum in Production should cover us for any interruptions


mec (approved):
🚢

PR #1876, opened 2024-09-12
Not applicable tasks clear their values

mec:
When marked as not applicable, some tasks clear their other values. We want this
to be the default behaviour for all optional tasks.

This work does that using callbacks on the `BaseTaskForm` which all tasks
inherit from.


lozette (approved):
Nice, I hadn't seen define_model_callbacks before!

PR #1889, opened 2024-09-18
Add an "inactive" state to projects

lozette:
Projects will soon be able to be added to Complete via an API call. These projects will be automatically created by the Prepare application.

Because the Prepare application does not hold all the mandatory information we need to create projects in Complete, these projects will be invalid. We need a way to indicate they are not ready to be worked on yet - they need to be triaged by DOs and put into a valid state.

To this end, add an "inactive" state to indicate these projects are not valid yet, and therefore cannot be worked on ("active" state).

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -92,7 +92,8 @@ class Project < ApplicationRecord
     active: 0,
     completed: 1,
     deleted: 2,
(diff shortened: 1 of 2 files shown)


mec (approved):
🚢

PR #1891, opened 2024-09-19
Set project state to "inactive" when created via the API

lozette:
When a project is created via the API, set its state to "inactive" to indicate it is not ready to be worked on yet.


mec (approved):
🚢

PR #1892, opened 2024-09-19
API can create transfer projects

mec:
The Complete API can already create a conversion project and a form a MAT
conversion project. This work adds transfer and form a MAT transfer projects
along with some improvements and fixes.

The API is not used in production yet so we have time to make improvements
beyond this work, but we wanted to keep this part as small as we could.


lozette (approved):
Nice work

PR #1899, opened 2024-09-25
Chore update Academies API spec helpers

mec:
We are going to need these helpers in some work on the application API so we
wanted to give them a once over.


lozette (approved):
Thanks for doing these 🌈

PR #1906, opened 2024-09-25
(Fix) Add missing outgoing_trust_ukprn to Transfers API params

lozette:
We require the outgoing trust UKPRN to create Transfers via the API, but had accidentally omitted it from the API params.


mec (approved):
🚢

PR #1907, opened 2024-09-25
(181801) Show the MP's display name with title

lozette:
The consumers of the exports want to address their letters to MPs using the MP's full name & title (e.g. the Right Honourable). We can get this from the Persons API as `displayNameWithTitle`

Keep the first & last name for now in case we still need to use these elsewhere.


mec (approved):
🚢

PR #1909, opened 2024-09-26
Add validaton to new projects on API

mec:
This work sets out to update the validation applied when a new project is created via the API. Essentially the validation is the same as the manual process the form objects use, with slight variations.

There is a lot of duplication here which is to be expected as the objects work in similar ways with variations, the end goal is to get rid of the manual forms, so at that point a lot of the duplication goes away.

Along the way we tried to make the specs as clear and as similar as possible, but this is possibly subjective!


lozette (approved):
Approved but you got a small typo!

PR #1910, opened 2024-09-26
(181519) Add "Change to government policy" DAO revocation reason

lozette:
Due to a new policy, we need to add a "Change to government policy" DAO
revocation reason. This reason acts like al lthe others - it is stored along
with a note.


mec (approved):
🚢

PR #1917, opened 2024-09-27
Release 90

lozette:
Changed

- The MP's name is now their full name with title (e.g. The Right Honourable)

Added

- Add new "Change to government policy" DAO revocation reason


mec (approved):
🚀

PR #1921, opened 2024-09-27
Better API validation errors

mec:
This work makes the validation errors returned by the API structured instead of
one big string.

Based on #1920

### DIFF
--- a/spec/api/v1/transfers_spec.rb
+++ b/spec/api/v1/transfers_spec.rb
@@ -108,7 +108,9 @@ RSpec.describe V1::Transfers do
           as: :json,
           headers: {Apikey: "testkey"}
 
-        expect(response.body).to eq({error: "Created by email is invalid"}.to_json)
+        expect(response.body).to include("validation_errors")
+        expect(response.body).to include("created_by_email")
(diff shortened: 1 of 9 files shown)


spec/api/v1/transfers_spec.rb
```diff
@@ -62,222 +26,219 @@ def valid_form_a_mat_transfer_parameters
...
       end
 
-      it "returns Not Allowed" do
+      it "returns Not Allowed for transfers" do
         get "/api/v1/projects/transfers", headers: {Apikey: "testkey"}
         expect(response.body).to eq({error: "405 Not Allowed"}.to_json)
         expect(response.status).to eq(405)
       end
+
+      it "returns Not Allowed for fotm a MAT transfers" do
```
lozette, line 47: typo "form"

lozette (approved):
This is really good, thanks! One small typo but otherwise great

spec/api/v1/transfers_spec.rb
mec: This is actually in a different bit of work that this was based on! I'll fix it in another PR! 👍

PR #1925, opened 2024-09-30
(182257) Do not show inactive projects in search results

lozette:
Inactive (state 4) projects are not ready to be worked on and therefore should not show up in search results (for now)


mec (approved):
🚢

PR #1928, opened 2024-10-01
Ensure Inactive projects don't appear in listing pages

lozette:
Inactive (state: 4) projects are not included in most listing pages but add some tests to make sure they dofinitely don't. Also exclude them from the "added_by" page and "not completed" by team pages.


mec (approved):
🚢

PR #1929, opened 2024-10-01
Add handover form object

mec:
To 'handover' a project created on the API we need to add some
attributes that do not exist in the Prepare application and so cannot be
provided, projects added via the API are invalid until this is done.

The attributes differ for conversions and transfers and the actual
assignment to a user is different depending on the team the project is
assigned to.

As the variation is so small, we kept to this single fancy form object
with two validation contexts. Project use single table inheritance, so
we can assign the project type attributes regardless.

The default context is used when the project is assigned to the Regional
casework services team.

The 'assign' context is used when the project is assigned to one of the
regional teams and so requires extra attributes to be set.

This essentially makes this a very small 'stepped' form with only two
steps, which we have done before, we'll send all the attributes to the
second step when we need to.

We ensure the unused attributes are set to nil so we can simply save
whichever Project is passed in.

This will likely need refinement once in use, but it's a lot to review already
so I would like to get this in first!

### DIFF
--- a/app/forms/new_handover_stepped_form.rb
+++ b/app/forms/new_handover_stepped_form.rb
@@ -1,0 +1,69 @@
+class NewHandoverSteppedForm
+  include Teamable
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attribute :assigned_to_regional_caseworker_team, :boolean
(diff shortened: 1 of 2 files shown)


lozette (approved):
Really like this pattern :+1:

PR #1931, opened 2024-10-02
User can handover inactive projects

mec:
There is a lot here, apologies! This is one journey with four paths and
we are handling it all with the fancy `NewHandoverSteppedForm` object so
it all comes at once!

This delivers a basic, working path for:

- handing over a conversion to Regional Casework Services
- handing over a conversion to a user in the Regional teams
- handing over a transfer to Regional Casework Services
- handing over a transfer to a user in the Regional teams

Designs are here:

 https://lucid.app/lucidspark/92056362-3523-413b-b954-7095607f95e9/edit?viewport_loc=13690%2C40250%2C5056%2C5276%2C0_0&invitationId=inv_1f6d15aa-a815-4f0e-91ac-c52940ae8ab8

The approach is that we don't want users to be able to return midway
through the steps (if they are on a path with steps) so we favour
rendering views over new paths - that means passing all the form
parameters to the second step and then posting them off - this feels
okay with so few attributes to manage and we do a similar thing for Date
Histories.

The bulk of the work is in the `HandoversController` which has to know
what type of project and render the right form views.

Most of the changes here are new views and the content to go with them.

We extended the `Projectable` controller concern to take care of invalid
UUIDs - this feels neater than the constraints we have in the routes (we
could switch everything else, but not here!).

We add validation that the email address matches up with a `User`.

We still need to:

- handle form a MAT projects (not sure these can come into Complete from Prepare yet?)
- autocomplete on the user email on the assign form
- add the journey to the navigation

### DIFF
--- a/app/controllers/all/handover/handovers_controller.rb
+++ b/app/controllers/all/handover/handovers_controller.rb
@@ -1,11 +1,12 @@
 class All::Handover::HandoversController < ApplicationController
-  after_action :verify_authorized
+  include Projectable
   rescue_from ActiveRecord::RecordNotFound, with: :not_found_error
+
+  before_action :set_form, except: [:check, :new]
+  after_action :verify_authorized
 
   def check
     authorize Project, :handover?
-
-    @project = Project.find(params[:id])
 
     case @project.type
     when "Conversion::Project"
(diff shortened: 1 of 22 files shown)


lozette (approved):
:rocket:

PR #1934, opened 2024-10-02
(Fix) Do not show deleted projects in the "Added by" page

lozette:
A user reported that they were seeing a deleted project in their "Added by" page. We were not explicitly excluding deleted projects from the "Added by" page or scope.


mec (approved):
This is good, but it might be worth switching these to:

`Project.added_by.active`

as that would exclude `deleted`, `completed`, `dao_revoked` and `inactive` more succinctly?

PR #1946, opened 2024-10-10
Remove regional handover step

mec:
We discovered this step was unnecessary and we can instead assume that if
the project is not being handed over to the regional casework team it
will be assigned to the current regional delivery officer (the current
user).


lozette (approved):
Nice 🚀

PR #1961, opened 2024-10-17
Enable users to run this repository using Codespaces

lozette:
## Changes

We had a requirement to run this application on Windows instead of MacOS (which most developers use). After a lot of discussion it was decided that enabling Codespaces on this repo was the "simplest" solution of many complex ones! This PR adds the Codespaces config and instructions on how to use it.

Co-authored by @mec with help from @peteryates

### DIFF
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -13,6 +13,12 @@
   <%= form_tag("/auth/azure_activedirectory_v2", method: "post") do %>
     <%= button_tag(t("sign_in.button", type: :submit), class: "govuk-button") %>
   <% end %>
+  <% if Rails.env.development? && ENV["CODESPACES"] %>
+    <%= form_tag("/auth/developer", method: "post") do %>
+      <%= button_tag("Developer login", class: "govuk-button") %>
+    <% end %>
+  <% end %>
+
   <h2 class="govuk-heading-m">Register for access</h2>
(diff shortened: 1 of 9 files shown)


app/views/sessions/new.html.erb
```diff
@@ -13,6 +13,12 @@
   <%= form_tag("/auth/azure_activedirectory_v2", method: "post") do %>
     <%= button_tag(t("sign_in.button", type: :submit), class: "govuk-button") %>
   <% end %>
+  <% if Rails.env.development? && ENV["CODESPACES"] %>
```
mec, line 16: I think we should mention why we had to do this in the commit message.

doc/running-with-codespaces.md
```diff
@@ -0,0 +1,47 @@
...
+After creating the Codespace, you will need to create a user and seed the database to 
+enable yourself to log in and see data. Open a terminal in your IDE and follow the 
+instructions in [Seeding the database](/doc/seeding-the-database.md)
+
+You will need to re-do the seeds every time you destroy and rebuild the Codespace, but
+*not* if you merely stop and restart the Codespace.
+
+### Logging into the application
+
+After you have created your user, you will need to log in via the "Developer login"
```
mec, line 44: It might be worth mentioning that the email address has to match that of the `User` you created?

doc/running-with-codespaces.md
```diff
@@ -0,0 +1,47 @@
...
+
+If you are not a developer and wish to run the application 
+(e.g. a Content Designer, UX) or are using a Windows machine, we recommend
+running this application using [Github Codespaces](https://github.com/features/codespaces).
+
+## Prerequisites
+
+You will need a Github account with read/write access to the repository. 
+
+You will
```
mec, line 11: Odd spacing here?

mec (changes requested):
Could of tiny documents things I would love to see.

mec (approved):
🚢

PR #1962, opened 2024-10-17
Release 92

lozette:
Changed

- New DfE font and header have been applied to the application.
- The links in the footer have been updated to the latest versions.

Fixed

- The primary contact for school in the export no longer uses the first contact if one is not set.
- The primary contact for the school is now labelled in the export.
- The primary contact for the incoming trust in the export no longer uses the first contact if one is not set.
- The primary contact for the incoming trust is now labelled in the export.
- The primary contact for the outgoing trust in the export no longer uses the first contact if one is not set.
- The primary contact for the outgoing trust is now labelled in the export.


mec (approved):
🚀

PR #1966, opened 2024-10-21
Bump deploy-azure-container-apps-action v3.0.0

Stretch96:
* Uses OIDC with Azure/login instead of Credential based authentication
* https://github.com/DFE-Digital/deploy-azure-container-apps-action/compare/v2.6.0...v3.0.0


lozette (approved):

DrizzlyOwl (commented):
@Stretch96 we need to consider that the 'worker' container has it's own credentials :)

DrizzlyOwl (approved):

PR #1972, opened 2024-10-22
Release 93

lozette:
Changed

- removed trust from the RPA radio button options text for conversions

Fixed

- The note added during DAO revocation only appears on the associated project notes, instead of all projects.


mec (approved):
🚀

PR #1979, opened 2024-10-29
Remove the Single Worksheet task from Conversions

lozette:
This task is no longer required. Remove it and its associated db columns, internationalisations etc.


mec (approved):
🚢

PR #1983, opened 2024-11-04
Revert "Remove the Single Worksheet task from Conversions"

lozette:
Reverts DFE-Digital/dfe-complete-conversions-transfers-and-changes#1979

This shouldn't be deployed until comms is completed


mec (approved):
🤦

PR #1984, opened 2024-11-04
Remove the Single Worksheet task from Conversions

lozette:
This task is no longer rquired. Remove it and its associated db columns, internationalisations etc.


edavey:
Is it worth expanding on the narrative a bit here to explain why the task is no longer needed? I see there's more info in https://dfe-gov-uk.visualstudio.com/Academies-and-Free-Schools-SIP/_workitems/edit/186134

edavey (approved):
👍

PR #1986, opened 2024-11-08
Give Service Support users access to see projects from Prepare

lozette:
Grant PASS users access to the Handover tab, so they can see projects handed over from the Prepare application. This is so they can respond to user queries.


mec (approved):
🚢

PR #1994, opened 2024-11-20
Remove diffy gem

lozette:
This isn't used anywhere, so remove it so we don't have to keep updating it.


edavey (approved):
Thanks for removing cruft!

edavey (approved):

PR #1996, opened 2024-11-21
Amend "Bank details changing" task to show its status correctly

lozette:
A user reported a bug where they were filling in the Transfer task "Confirm if the bank details for the general annual grant payment need to change" but the task was never showing as "Completed" in the task list.

This task is unusual in that it is the only boolean task which is displayed with radio buttons in the UI.

The task was indeed always showing as "Not started" despite being filled in.

After some investigation, we have determined that:

1. the boolean was set to be false by default, which is fine for checkboxes but for radio buttons it gives the impression the task has been filled in, when it is only the default value.

Therefore this column needs to be nil by default.

2. the mechanism for checking if the task was completed did not work for tasks with radio buttons.
`attributes.values.all?(&:present?)` will return false if the "No" radio button is checked (because No is false).

Instead we need to check if the value is nil with
`!attributes["yes_no"].nil?`

To fix this, change the db column to be nul by default, not false; and override the `completed?` method to check this specific attribute for nil.

Also add a test to make sure this is all working!


edavey (approved):
Very clear and thorough, thanks!

PR #2006, opened 2024-11-28
Cache the trivydb for a day to mitigate TOOMANYREQUESTS error

DrizzlyOwl:
## Changes

Occasionally we run into a problem where the Docker image scan task encounters a TOOMANYREQUESTS error and times out. By caching the vulnerability database we reduce the frequency that we need to check for new versions.


edavey (approved):
Thanks, this sounds like it might solve the issue!

PR #2012, opened 2024-12-02
(Fix) Account for the change in year in export tests

lozette:
These tests started failing in December 2024 as we weren't factoring in the change of year 2024 -> 2025. We were expecting *this* year when looking for the next month's data,
but in December the "next month" exports will start returning *next* year's results.


edavey (approved):
Neat fix, thanks!

PR #2014, opened 2024-12-03
Ensure source is checked out for Sonar scan to work

DrizzlyOwl:
* without the source, the sonar-project.properties file is missing which prevents Sonarcloud from understanding how we want it to scan the project

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -151,6 +151,9 @@ jobs:
     needs: specs
     steps:
       -
+        name: Checkout
+        uses: actions/checkout@v4
+      -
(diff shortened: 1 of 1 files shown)


edavey (approved):
Thanks for fixing this. How was the CI action working before?

DrizzlyOwl:
> Thanks for fixing this. How was the CI action working before?

The checkout line was [inadvertently removed by me](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/commit/34600b06f70fc3ea7f22422e8c31b8bdac5a7c8d#diff-e6971a6e2989960eb8c20fb53622b949608b9f0129a80c9e645b208710044d42L167)

PR #2015, opened 2024-12-03
(Fix) Force Header and DfE Nav links to always be white

lozette:
These were being overridden by the DfE styles, ending up blue when they should be white.


dangood84 (approved):

RichikaD (approved):
LGTM!

PR #2016, opened 2024-12-03
Revert "Update ruby Docker tag to v3.3.6"

lozette:
Reverts DFE-Digital/dfe-complete-conversions-transfers-and-changes#1993

Reverting this change while we investigate some failures on the dev environment.


edavey (approved):
thanks Laura

PR #2017, opened 2024-12-03
Upgrade to Ruby 3.3.6

lozette:
Renovate attempted this upgrade by only updating the Dockerfile and Gemfile, and not the .ruby-version. Retry this upgrade with all parts in place.


edavey (approved):
thanks for resolving this, along with the OpenStruct loading issue.

PR #2030, opened 2024-12-12
Return 'Healthy' or 'Unhealthy' string after db test

edavey:
The other products in the programme are all returning a string rather than a JSON representation.

We now:

- conform to that convention
- include a db connection test

## Checklist

- [ ] Attach this pull request to the appropriate card in DevOps.
- [x] Update the `CHANGELOG.md` if needed.

PR #2033, opened 2024-12-13
Change healthcheck endpoint to render plain text

lozette:
In PR #2030 we altered the API healthcheck (`/api/v1/healthcheck`) endpoint to return plain text instead of JSON. We forgot that we actually have two healthcheck endpoints, one simply at `/healthcheck`. Change this endpoint to also return plain text.


edavey (approved):
Thanks.

PR #2039, opened 2024-12-16
Remove Service support "Upload GIAS data" functionality

lozette:
Remove the area in Service support where GIAS data was uploaded.

We have recently set up a database sync whereby all GIAS establishment data is now synced into the `gias_establishments` table on the Complete database.

As a result we no longer need anyone from Service support to perriodically upload GIAS data into this table.

Remove the pages in the Service support view where this upload happened. Also remove all the services, jobs, mailers and tests associated.

## Checklist


edavey (approved):
thanks for clearing this out now that it's no longer needed.

PR #635, opened 2023-04-06
Deploy Release 20 to the test environment

lozette:

PR #2052, opened 2025-01-08
194142 - Fix: Enforce uniqueness of URN for active/inactive projects with trigger

edavey:
## Enforce uniqueness of URN for active/inactive projects with db trigger

See [Ticket 194142: Duplicate Projects on Complete (107926 and 135287)](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/194142)

We add a `EnforceUniquenessOfProjectUrn` db trigger to raise an
error and rollback "after insert" if that insert is attempting
to add a new project record where an existing active or inactive
project with the same URN exists.

We've been seeing some duplicate transfers and conversions being
created by our automated "handover" process between "prepare" and
"complete". Our "complete" API receives POSTs to:

- `/api/v1/projects/conversions` and
- `/api/v1/projects/conversions`

these are handled by:

which both use the `UrnUniqueForApiValidator` to perform this
validation at the application level, but for a reliable
constraint we need to use a control at the database layer.

## Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -41,6 +41,8 @@
 **/*.db*
 **/*.dump*
 **/*.sql*
+# Allow trigger definitions
+!/db/triggers/*.sql*
 **/*.sqlite3*
 ### Environment variables
 **/.env
--- a/config/application.rb
+++ b/config/application.rb
@@ -24,6 +24,9 @@ require "govuk/components"
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
+adapter = ActiveRecord::ConnectionAdapters::SQLServerAdapter
+adapter.exclude_output_inserted_table_names["projects"] = "uniqueidentifier"
+
 module DfeCompleteConversionsTransfersAndChanges
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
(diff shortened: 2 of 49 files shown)


edavey:
When testing this trigger out manually in the `dev` environment we observed the:

```
the DML statement cannot have any enabled triggers if the
statement contains an OUTPUT clause without INTO clause
```

error in the .NET application:

exception_error_on_ net_dev_env _while_adding_the_conversion_project

So in order to use this trigger approach we'll need, on the .NET side to also (as we've done here on the Ruby side):

- configure the application to play nicely with a table which has a UUID as primary key (`products`)
- find a way to ensure that the trigger is added as part of the ER migration process

Alternatively, can we find a different solution which will be less work to implement and manage? Ideas offered up in a grouop review session include:

- using UPSERT rather than INSERT
- implementing the "outbox" pattern whereby the calling service (a "hosted service" in the Prepare app) provides an ID for the transaction and this is used to avoid POSTing the same request 
- increasing the time interval in the "hosted service" jobs generated in the Prepare app
- using a job queue e.g. in Redis to avoid parallel processing

edavey:
@paullocknimble  has put together a 

> POC on how to lock the rows to stop duplicates being sent, it is half way house to implementing the outbox pattern but would require less effort, it's not finished and would require lots of testing but as a concept it should work  
[POC on locking rows so they are not sent twice by paullocknimble · Pull Request #635 · DFE-Digital/academies-academisation-api](https://github.com/DFE-Digital/academies-academisation-api/pull/635)

edavey:
I'm closing this as our current plan is to fix this at the "Prepare" side, eg by using a job queue which guarantees that each job is processed exactly once

PR #2055, opened 2025-01-13
193590: Set PageVisitTime metric in AppInsights

edavey:
See ticket [193590: Add tracking of "Time on page" with AppInsights](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/193590)

This is to track "time on page".

See https://learn.microsoft.com/en-us/azure/azure-monitor/app/javascript-sdk-configuration

### DIFF
--- a/app/javascript/application-insights.js
+++ b/app/javascript/application-insights.js
@@ -10,7 +10,8 @@ const clickPluginConfig = {
 
 const appInsights = new ApplicationInsights({
   config: {
-    connectionString: applicationInsightsConnectionString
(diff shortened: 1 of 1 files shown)


edavey:
This has now been tested in the dev environment and I've confirmed that the new custom `PageVisitTime` metric is now coming through into AppInsights. 

Note that this was temporarily enabled in `dev`. AppInsights is normally only loaded in `production`.

DrizzlyOwl (approved):

PR #2060, opened 2025-01-16
Handle some common errors, previously seen as unhandled exceptions

edavey:
See Slack channel [`#temp-ruby-complete-alerts`](https://ukgovernmentdfe.slack.com/archives/C089F7F3CD6) and [ticket 197151](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/197151)

In this PR we address 4 types of error which have been surfaced by our new Exception Notification process. These are currently manifesting as unhandled exceptions but in fact they are routine errors which can and should be handled by the application:

1. where `ActionController::InvalidAuthenticityToken` is raised as a result of a request having an an invalid auth token

2. where a very long request path was triggering an error in the `HighVoltage` gem

3. where a path with invalid encoding was causing Rack to raise `InvalidParameterError`  which is rescued by Rails as a internal `ActionController::BadRequest`

4. where a query string was causing Rack to raise `ParameterTypeError`  which is rescued by Rails as a internal `ActionController::BadRequest`

See the individual commits for details.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,5 +1,13 @@
 class PagesController < ApplicationController
-  include HighVoltage::StaticPage
-
   skip_before_action :redirect_unauthenticated_user
+
+  def show
+    return not_found_error unless StaticPages.names.include?(static_page_name)
+
+    render static_page_name
+  end
+
+  private def static_page_name
+    params[:id]
+  end
 end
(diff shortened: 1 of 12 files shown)


app/controllers/pages_controller.rb
shuldt, line 1: Nice!

lib/middleware/handle_bad_encoding.rb
shuldt, line 1: Very neat

shuldt (approved):
Very clearly communicated changes, thank you for this. 

These are good solutions, and I learned some new tips and tricks for error handling from this!

There is a minor typo in the last commit message: "and clears it if it can't parsed" should be "and clears it if it can't BE parsed"

PR #2064, opened 2025-01-22
196052 Improve performance of local auth project page

edavey:
See ticket [196052](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/196052)

We've had reports of frequent timeouts when viewing the list of projects for a particular local authority. At URLs like:

`https://www.complete.education.gov.uk/projects/all/local-authorities/353`

On investigation we can see that this is because our persisted `Project` entity does not record the local authority information. Therefore to get the projects for a given LA we are currently call the Academies API for every single project and then filtering that list to select the projects for our given Local Authority!

See the `ByLocalAuthorityProjectFetcherService#projects_for_local_authority(la_code)` https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/797540d03ba3d0e55ffe1a9b85dc5eb893ab72a6/app/services/by_local_authority_project_fetcher_service.rb#L12C3-L14C6

```rb
def projects_for_local_authority(local_authority _code)
  all_projects
	.includes(:assigned_to)
	.select { |p| p.establishment.local_authority_code == local_authority_code }
end
```

### Solution

We need to record the Project's local authority in our persisted entities. As we have a list of Local Authorities (`LocalAuthority`) we can do this with an association which specifies that "a Project belongs to a Local Authority", i.e. by adding a `local_authority_id` foreign key to `projects` so that we can then set the LA:

```rb
Project.local_authority = LocalAuthority.find_by!(code: "801") # for Bristol City Council
```

If this were a high volume service then we would deploy each of these step separately to avoid the chance of
race conditions. But with such a small number of users that is not necessary.

#### 1. Add new field to Project

Create `projects.local_authority_id` in UUID format, allowing null values, for now. This will be the foreign key to the `LocalAuthority` entity backed by the `local_authorities` table

#### 2. Start writing local_authority_id for new projects

We'll begin storing `Project#local_authority_id` when creating projects using:

- `Transfer::CreateProjectForm`
- `Conversion::CreateProjectForm`

At this stage we'll set the property explicitly, rather than through an association e.g.:
```rb
Transfer::Project.new(
  ...
  local_authority_id: local_authority.id,
  ...
)
```

rather than using the belongs_to association which we'll add in step 4:

```rb
Transfer::Project.new(
  ...
  local_authority: local_authority,
  ...
)
```

#### 4. Switch the "project belongs to local authority" association

At present `Project##local_authority` delegates to the `establishment` which hits the Academies API. We'll change over to getting local authority from the our own `local_authorities` table using our new `Project#local_authority_id` fk.

#### 5. Make the new foreign key mandatory

In a future deployment we can make `project.local_authority_id` non null-able.


shuldt (approved):
Happy with the solution and its implementation - do let me know what decisions you make regarding deployment (in stages or bundled as all commits in this PR)

edavey:
When checking in the **dev** environment we found several projects which had blank `local_authority_id` keys. All except one has invalid URNs (having been added by the new .NET service under development which has not yet added the require validation against the Academies API). There was one record which should have been associated with the "City of London" local authority which is missing from the `complete.local_authorities` table. This is being added to the seeding in a further commit. It's present in **test** (and therefore in **production)**.

PR #2073, opened 2025-02-04
(196052) Step 1 of 2: Refactor project to record local authority

edavey:
See ticket [196052](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/196052)

After some more testing on  the dev and test envs we've observed that sometimes the data migration to backfill `projects#local_authority_id` fails to populate a few rows due to transient errors thrown by the Academies API (which is queried for each row). 

We're changing tack on the "deploy in one go" approach (see [PR 2064](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/2064)) and will  now do the refactoring in 2 deployments:

#### 1. step 1 of 2 (THIS PR)

- i) create the new field, 
- ii) start populating it, 
- iii) backfill the new field

#### In between

Fix up any rows which didn't get backfilled. If not we may see "whiny nil" errors where projects are missing their local authority association. 

#### 2. step 2 of 2

- i) start using the new association 
- ii) fix up the slow pages (e.g. `projects/all/local-authorities`) to get the benefit of the refactoring

### DIFF
--- a/app/controllers/project_groups_controller.rb
+++ b/app/controllers/project_groups_controller.rb
@@ -14,7 +14,7 @@ class ProjectGroupsController < ApplicationController
 
     @project_group = ProjectGroup.find(params[:id])
 
-    projects = @project_group.projects.includes([:local_authority])
+    projects = @project_group.projects
     AcademiesApiPreFetcherService.new.call!(projects)
 
     @grouped_projects = projects.sort_by { |project| project.establishment.name }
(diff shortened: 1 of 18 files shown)


shuldt (approved):
Took me a little while to get my head around the verts and reverts, but yes, I'm happy with this and very happy to see a change in approach as required by testing.

PR #2078, opened 2025-02-06
(199738) Avoid attempting to return an excessive number of search results

edavey:
See https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/199738

We've seen errors coming from the project search module when the search results are very numerous, which then causes an excessive number of requests to the Academies API. 

These errors have only been seen in the development environment but could just as easily occur in production if a user searches for "the" or "1".

We add two guard clauses to avoid attempting to fetch large quantities of records from the Academies API:

- if the search term given is less than 4 characters we raise with "Search term too short"

- if the number of results is more than 100 we raise with "Too many results"

These "Search Errors" are handled by showing the user a new "search error" page which gives them helpful information on the problem:

helpful_search_error_page

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -7,6 +7,8 @@ class SearchController < ApplicationController
     @query = query
     @count = @results.count
     @pager, @results = pagy_array(@results)
+  rescue ProjectSearchService::SearchError => error
+    render "pages/search_error", locals: {error_message: error.message}
   end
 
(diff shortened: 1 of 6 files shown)


DrizzlyOwl (approved):
There is room to improve the UX here as it's not strictly an "error", the user just needs to _do more_ but otherwise the feature is an improvement

PR #2080, opened 2025-02-06
194142 Fix: prevent duplicate inserts by enforcing uniqueness of URN for active projects

edavey:
### Enforce uniqueness of URN for active projects

See ticket [194142](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/194142)

We add a composite uniqueness index on the `projects` table
to ensure that only one project in an active state with a given
URN can exist.

We've been seeing some duplicate transfers and conversions being
created by our automated "handover" process between "prepare" and
"complete". Our "complete" API receives POSTs to:

- `/api/v1/projects/conversions` and
- `/api/v1/projects/conversions`

these are handled by:

which both use the `UrnUniqueForApiValidator` to perform this
validation at the application level, but for a reliable
constraint we need to use a control at the database layer.

As new projects are added with the "active" state it's sufficient to
use the rule that no two projects may have the same state and URN.

When deploying this change we'll need to identify any existing projects 
which violate this rule and handle them, probably by moving
all but the most recent duplicate into the deleted state. 

See the list of [project states][] .

[project states]:
https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/blob/fa42c6bca30696c4ca34b3fe2a3db568244f3294/app/models/project.rb#L92C1-L98C4

## Checklist

- [ ] Understand if production contains records which violate this new constraint. If so 
   add a data migration to fix up before adding the unique index

### DIFF
--- a/db/migrate/20250205114547_add_unique_project_index_on_urn_and_state.rb
+++ b/db/migrate/20250205114547_add_unique_project_index_on_urn_and_state.rb
@@ -1,0 +1,5 @@
+class AddUniqueProjectIndexOnUrnAndState < ActiveRecord::Migration[7.1]
+  def change
+    add_index :projects, [:urn, :state], unique: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2025_01_27_172701) do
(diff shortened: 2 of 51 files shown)


shuldt (approved):

edavey:
I'm closing this as our current plan to to fix this on the "Prepare" side, e.g.  by using a job queue which guarantees that each job is processed exactly once.

PR #2085, opened 2025-02-10
200459 Fix: set local authority on imported projects

edavey:
See [200459: Imported projects missing LA](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/200459)

We failed to recognise that  projects (transfers and conversions) are not only created within the service via the GUI using the `CreateProjectForm` classes, but also programmatically in a pipeline job from the "Prepare" service. The classes responsible are:

In this PR we extend these two services to set the local authority on the project being created from the local authority of the establishment returned by the Academies API.


shuldt (approved):
Good catch, lgtm

PR #2092, opened 2025-02-18
Make Projects#local_authority_id non-nullable

edavey:
This will ensure that projects always have their local authority set.

This will reduce the number of errors we see in the development environment caused by projects created by the .NET app which is under-construction and which has incomplete validation.


shuldt (approved):
Seems sensible!

PR #2101, opened 2025-02-25
Fix: Pin forked AppInsights gem to tag rather than branch

edavey:
We got caught out after rewording the fix commit on `fix/correct-pattern-match-for-exception-handling` not appreciating that pinning the gem to the branch was actually tieing it to SHA `51cb59a34e3e36ff406f9ec3c8332fb2a2ff89c5` which was of course replaced when rewording the commit.

We now use a tag to pin the forked gem: this is more obviously a pointer to a particular SHA.


shuldt (approved):
Delighted to approve this just as I ran into the error being fixed on my machine

PR #2104, opened 2025-03-03
200841: Find inactive project to handover by urn

edavey:
## Show 'find inactive project' when there are many projects to handover

See ticket [200841](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/200841)

At certain times (e.g. now, start of March) there are a lot of projects to hand over, more than can be listed on a single index page.
    
If there is more than one page of projects to handover we now offer a  "Search for inactive project by URN" feature.
    
If a matching inactive project is found we provide a link to that project's "check handover details" page. If no matching inactive project exists we return the user to list of projects to be handed over.
    
### User story
   
- So that I can review a particular project which is to be handed over
- As a Regional Case Officer team leader at a time when there are over 20 projects in the "to be handed over" (inactive) state
- I want to locate this "to handover" (inactive) project without paging through multiple records

### Search facility displayed where quantity of projects requires pagination

multiple_pages_search_offered

---

### Project is found by URN

found_inactive_project

---

### Project is NOT found by URN

not_found-redirect_to_list

---

### Search facility NOT shown where quantity of projects does NOT require pagination

single_page_no_search_needed

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased][unreleased]
 
+### Changed
+
+- Find a project to be handed over by URN, if there are too many to show on a
+  single page.
+
 ## [Release-108][release-108]
 
 ### Changed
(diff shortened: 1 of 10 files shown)


shuldt (approved):
Looks really useful!

PR #2108, opened 2025-03-05
204455: FIX - Link 'By month' to current month's conversions to show active state

edavey:
See [ticket 204455](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/204455)

Previously the "By month" link in the main navigation was linking to:

`projects/all/by-month/conversions/from/to` 
(`date_range_this_month_all_by_month_conversions_projects_path`)

which was showing a list of conversions for the current month. But because the url was not matching the more literal "Conversions" link in the sub-nav:

`/projects/all/by-month/conversions/from/3/2025/to/3/2025` 
(`date_range_all_by_month_conversions_projects_path`)

the current "Conversions" tab was not being styled as active. In this commit we change the main nav's "By month" link to use the more explicit link style for the current month's conversions. In this way we now see that the default tab is "Conversions".

## Before

before_ by_month

## After

after_by_month

## Checklist

### DIFF
--- a/app/views/shared/navigation/_all_projects_primary_navigation.html.erb
+++ b/app/views/shared/navigation/_all_projects_primary_navigation.html.erb
@@ -1,45 +1,27 @@
 <div class="moj-primary-navigation">
-
   <div class="moj-primary-navigation__container">
-
     <div class="moj-primary-navigation__nav">
-
       <nav class="moj-primary-navigation" aria-label="Primary navigation">
-
         <ul class="moj-primary-navigation__list">
-
           <% if policy(:project).handover? %>
             <%= render partial: "shared/navigation/primary_navigation_item", locals: {name: t("navigation.primary.all_projects.handover"), path: all_handover_projects_path, namespace: "/projects/all/handover"} %>
           <% end %>
-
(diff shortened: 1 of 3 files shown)


app/views/shared/navigation/_all_projects_primary_navigation.html.erb
```diff
@@ -1,44 +1,27 @@
 <div class="moj-primary-navigation">
-
```
shuldt, line 2: Is there a rationale for these re-formatting changes?

shuldt (approved):
Looks good

app/views/shared/navigation/_all_projects_primary_navigation.html.erb
edavey: > Is there a rationale for these re-formatting changes?

thanks for the question: I've now broken out the reformatting into a separate commit with an explanation of why it's done: https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/2108/commits/83798cd5f830bbe81d994e13d42ac81257ec5e5b

shuldt: Thanks for that, makes sense!

PR #2109, opened 2025-03-06
Show a error msg if project to hand over is missing UKPRN

edavey:
See Ticket [202355](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/202355) 

We regularly see the scenario in support where: 

- a project has been imported from Prepare into Complete without an  "incoming trust UKPRN" to identify the trust which will be managing the new academy. 
- the user attempts to "hand over" the project (move it from the "inactive" or "to be handed over" state)
- the user sees a generic error message

Behind the scenes a validation error has occurred when the `NewHandoverForm` attempts to assign the project and set its state to "active":

```
ActiveRecord::RecordInvalid:
  Validation failed: 
  Incoming trust ukprn UKPRN must be 8 digits long and start with a 1. 
  For example, 12345678.
``` 

In this PR we anticipate this scenario and show the user a message explaining that the project must:

- be deleted in Complete
- corrected in Prepare
- resubmitted to Complete

cant_handover_missing_ukprn

## Checklist

### DIFF
--- a/app/controllers/all/handover/handovers_controller.rb
+++ b/app/controllers/all/handover/handovers_controller.rb
@@ -27,6 +27,8 @@ class All::Handover::HandoversController < ApplicationController
   def create
     authorize Project, :handover?
 
+    return redirect_to_project if @project.incoming_trust_ukprn.blank?
+
     if @form.valid?
       case @form.assigned_to_regional_caseworker_team
       when true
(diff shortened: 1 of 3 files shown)


app/controllers/all/handover/handovers_controller.rb
```diff
@@ -41,6 +43,14 @@ def create
     end
   end
 
+  private def bail_missing_incoming_trust_ukprn
```
shuldt, line 46: Is 'bail' conventional naming in this project for this type of error - I can't find it used elsewhere

shuldt (approved):
Nicely handled, should really improve user experience

app/controllers/all/handover/handovers_controller.rb
shuldt: Fab - I don't mean to nitpick, just trying to learn about the project :)

PR #2112, opened 2025-03-11
Switches to composite actions instead of workflows for build and push image

shuldt:
Alters the build and push image yml config to use composite actions instead of workflows. 

This is a requirement of upgrading to deploy-azure-container-apps-action to v5, which otherwise constitutes a breaking change. 

This upgrade was introduced by the renovate [PR 2097 'DFE-Digital/deploy-azure-container-apps-action action to v5'](https://github.com/DFE-Digital/dfe-complete-conversions-transfers-and-changes/pull/2097).

## Changes

This commit includes the v5 upgrade, and reconfigures the build, import and deploy steps to use github actions instead of workflows.

**Please note:** the annotate-release value in the deploy job is commented out. We are currently experiencing problems with this annotation step - the expected behaviour is that it run only in the 'web' container app of the deploy matrix, however it is also running in 'worker', and that is failing. 

As this does neither interfere with the deployment, nor with any downstream consumers of that annotation (there are none) we are going to separately debug this once the composite actions changes are merged.

## Checklist

### DIFF
--- a/.github/workflows/build-and-push-image.yml
+++ b/.github/workflows/build-and-push-image.yml
@@ -20,7 +20,7 @@ env:
 jobs:
   set-env:
     name: Determine environment
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     outputs:
       environment: ${{ steps.var.outputs.environment }}
       branch: ${{ steps.var.outputs.branch }}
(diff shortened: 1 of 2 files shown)


edavey (approved):
does the fact that we are running in both web and worker builds mean that the overall build time is longer? Or do the builds run in parallel?

DrizzlyOwl:
> does the fact that we are running in both web and worker builds mean that the overall build time is longer? Or do the builds run in parallel?

The only part of the workflow that references two separate container apps is the final deploy step which needs to restart both services. The build and import stages only process a single image so should remain roughly the same timing as before

PR #2120, opened 2025-03-19
Release 111

edavey:
### Fixed

- Allow handover of (Form a MAT) projects with a TRN instead of incoming trust UKPRN.
- Ensure that we don't send the "New project to assign" email on update if the project is already assigned to a caseworker.
- Send the "New project to assign" email when a project created by Prepare is handed over.


shuldt (approved):
Let's go!

PR #2139, opened 2025-04-17
(210076) Fix: toggle User#manage_team via UI

edavey:
Ticket [210076](https://dev.azure.com/dfe-gov-uk/Academies-and-Free-Schools-SIP/_workitems/edit/210076)

In e5fb0398172082431f56104fa43d951b4c1c8667 we renamed `User#team_leader` to `User#manage_team` -- but failed to change the form used by service support team members to edit the attribute via the UI.

Now it's again possible to toggle this "manage_team" property using the form at:

`/service-support/users/{userId}/edit`

We also add some guidance to advise the user that setting the `User#manage_team` attribute:
    
> will only be applied for RDOs and RCS team members. If the user is a member of a different team they will need to have a user-specific "Manage Team Capability" set. Raise a support ticket to request this.

## Screenshot

edit_user_manage_team

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Handle pagination errors gracefully.
+- Allow 'manage team' user attribute to be set by service support
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


shuldt (approved):
looks good

